PR #459, opened 2019-09-12
Rework processes plugin

0815ak:
The backend for the processes-plugin was totally reworked to have one dedicated class to use IPC and internally communicate via events. It was also adapted to use the "single"-mode for this plugin.

For the multiple processes feature, there now is a convenient feature that automatically adds a new process-line, whenever a task takes some time to perform.

Also the shell-env-library is now used for shell-related commands and henceforth the process.env-class was removed.


marcmo:
functionality already included

PR #477, opened 2019-09-25
indexer support for cancellation and progress monitoring

marcmo:


marcmo:
will be merged without pull

PR #509, opened 2019-10-23
Fixed neon task issue

ir210:
Use neon binary from node_modules directory instead of the system one.

### DIFF
--- a/rakefile.rb
+++ b/rakefile.rb
@@ -690,7 +690,7 @@ task :build_embedded_indexer do
   cd "#{APPS_DIR}/indexer-neon" do
     npm_install
     sh "#{NPM_RUN} build-ts-neon"
-    sh 'node_modules/.bin/electron-build-env neon build --release'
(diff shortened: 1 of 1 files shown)


marcmo:
this causes weird build problems on windows.

rakefile.rb
```diff
@@ -690,7 +690,7 @@ def package_and_copy_neon_indexer(dest)
   cd "#{APPS_DIR}/indexer-neon" do
     npm_install
     sh "#{NPM_RUN} build-ts-neon"
-    sh 'node_modules/.bin/electron-build-env neon build --release'
+    sh 'node_modules/.bin/electron-build-env node_modules/.bin/neon build --release'
```
marcmo: # better
    if OS.windows?
      sh 'node_modules/.bin/electron-build-env neon build --release'
    else
      sh 'node_modules/.bin/electron-build-env node_modules/.bin/neon build --release'
    end

PR #535, opened 2019-11-08
Pcap integration

akermu:
This is a development branch, please to not merge.

### DIFF
--- a/application/apps/indexer-neon/native/Cargo.lock
+++ b/application/apps/indexer-neon/native/Cargo.lock
@@ -250,6 +250,7 @@ dependencies = [
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "nom 5.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "pcap 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "proptest 0.9.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "proptest-derive 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc-hash 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


application/apps/indexer-neon/native/Cargo.lock
```diff
@@ -1143,7 +1152,11 @@ dependencies = [
 "checksum nom 5.0.1 (registry+https://github.com/rust-lang/crates.io-index)" = "c618b63422da4401283884e6668d39f819a106ef51f5f59b81add00075da35ca"
 "checksum num-integer 0.1.41 (registry+https://github.com/rust-lang/crates.io-index)" = "b85e541ef8255f6cf42bbfe4ef361305c6c135d10919ecc26126c4e5ae94bc09"
 "checksum num-traits 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)" = "6ba9a427cfca2be13aa6f6403b0b7e7368fe982bfa16fccc450ce74c46cd9b32"
+<<<<<<< HEAD
```
marcmo: you have to correctly update the cargo.lock file...here the conflicts of a merge were committed

application/apps/indexer/dlt/src/dlt_pcap.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use crate::dlt::{StorageHeader, DltTimeStamp};
+use pcap::Capture;
+use std::path::Path;
+use std::fs::File;
+use std::io::Write;
+use std::error;
+
+pub fn convert_pcap_to_dlt(pcap_path: &Path, dlt_path: &Path, port: u16) -> Result<(), Box<dyn error::Error>> {
+    let mut cap = Capture::from_file(pcap_path).unwrap();
```
marcmo: avoid `unwrap` since it might panic.
you already return a `Result` so propagate errors accordingly

application/apps/indexer/dlt/src/dlt_pcap.rs
```diff
@@ -0,0 +1,45 @@
...
+// from E.S.R.Labs.
+
+use crate::dlt::{StorageHeader, DltTimeStamp};
+use pcap::Capture;
+use std::path::Path;
+use std::fs::File;
+use std::io::Write;
+use std::error;
+
+pub fn convert_pcap_to_dlt(pcap_path: &Path, dlt_path: &Path, port: u16) -> Result<(), Box<dyn error::Error>> {
```
marcmo: filtering for a port is probably required. but then we would also need a function to ask for ports that occur in the trace. s.th. like `fn get_ports(pcap_path: &Path) -> Result<Vector<u16>, Error>`

akermu: We would then need to parse the pcap trace first, if we need this information, but it definitely makes sense.

marcmo:
feature has been implemented in other commit

PR #544, opened 2019-11-21
[](feat) change exec path detection for darwin

DmitryAstafyev:


application/electron/src/controllers/files.parsers/file.parser.dlt.ts
```diff
@@ -17,6 +17,16 @@ import { Subscription } from "../../tools/index";
...
+interface IDLTFilters {
+    [key: string]: string[];
+}
+
+interface IDLTOptions {
+    logLevel: number;
+    filters: IDLTFilters;
+    fibexFilePath?: string;
+}
+
```
marcmo: why is this defined in 2 places? should only exist once

application/electron/src/services/files/service.file.picker.ts
```diff
@@ -0,0 +1,128 @@
...
+ * @class ServiceFilePicker
+ * @description Opens files and returns fullpath
+ */
+
+class ServiceFilePicker implements IService {
+
+    private _logger: Logger = new Logger('ServiceFilePicker');
+    // Should detect by executable file
+    private _subscription: { [key: string]: Subscription } = {};
+    private _options: any;
```
marcmo: why do you need this field? seems not to be used

application/electron/src/controllers/files.parsers/file.parser.dlt.ts
marcmo: why is this defined in 2 places? should only exist once

application/ipc/electron.ipc.messages/file.filepicker.response.ts
```diff
@@ -0,0 +1,35 @@
...
+    constructor(params: IFilePickerResponse) {
+        if (typeof params !== 'object' || params === null) {
+            throw new Error(`Incorrect parameters for FilePickerResponse message`);
+        }
+        if (!(params.files instanceof Array)) {
+            throw new Error(`files should be defined as IFilePickerFileInfo[].`);
+        }
+        if (params.error !== undefined && typeof params.error !== 'string') {
+            throw new Error(`error can be defined only as string`);
+        }
```
marcmo: those type assertions should not be needed. we know what the type of params is. so why add those checks?

application/electron/src/services/files/service.file.picker.ts
DmitryAstafyev: you are right. It isn't used. Will be removed

application/electron/src/controllers/files.parsers/file.parser.dlt.ts
DmitryAstafyev: I would say it's a more complex issue. And it's good we can resolve it white it's "small". We have "indexer-neon", we have some common things between render and process (front-end and back-end). Many common things (like interfaces) are shared between render and process via IPC message-classes. But not all, and DLT is a good case.

Now I've found in solution (and in render, and in process) direct references to indexer. Like `../../../apps/indexer-neon`. I believe it's wrong, because it breaks namespaces of parts of solution. 

In another hand, we have to share some interfaces between render and indexer-neon (for the process it isn't a problem, because indexer-neon there is a node module). The best way I think - it's creating "outside" folder `common` and subfolder `interfaces`. Also into `common`, folder `ipc` should be moved. 

it will look like
```
application
└── common
    ├── interfaces
    └── ipc
```

In this case, we can avoid from direct references to `indexer-neon` from render.

application/ipc/electron.ipc.messages/file.filepicker.response.ts
DmitryAstafyev: Because we have to check also it in run-time. Simple use-case, some information comes from `indexer-neon`, we are expecting field `params.files` will be an array. And during compilation - no any issues. But in run-time, something goes wrong and instead of an array we've gotten `undefined`. 

Without run-time checks, such error will go forward and fires some bug, for example, on render level. So, developer will start to debug somewhere in front-end, while problem comes from another place.

BTW, a time ago, we had exactly same use-case: in run-time indexer-neon sends wrong data and because we have such checks in scope of IPC message-classes, issue was located very fast

PR #615, opened 2019-12-19
Sparkline integration

fur-gan:
Create Sparklines for serial port plugin to show activity on popup window for available ports

### DIFF
--- a/application/client.plugins/projects/serial/src/lib/views/dialog/components.ts
+++ b/application/client.plugins/projects/serial/src/lib/views/dialog/components.ts
@@ -52,12 +52,10 @@ export class SidebarVerticalPortDialogComponent implements OnInit, OnDestroy {
 
     ngOnInit() {
         this._subscriptions.Subscription = Service.getObservable().event.subscribe((message: any) => {
-            if (typeof message !== 'object' && message === null) {
+            if (typeof message !== 'object' || message === null) {
                 return;
             }
-            if (message) {
-                this._onSpyState(message);
-            }
+            this._onSpyState(message);
             this._forceUpdate();
         });
         this._ng_ports = this._requestPortList();
@@ -72,16 +70,14 @@ export class SidebarVerticalPortDialogComponent implements OnInit, OnDestroy {
     }
 
     ngOnDestroy() {
-        if (this._interval) {
-            clearTimeout(this._interval);
-        }
+        clearTimeout(this._interval);
         Object.keys(this._subscriptions).forEach((key: string) => {
             this._subscriptions[key].unsubscribe();
         });
         this._destroyed = true;
     }
 
-    public observe(): { event: Observable<any> } {
+    public onInterval(): { event: Observable<any> } {
         return { event: this._subjects.event.asObservable() };
     }
 
--- a/application/client.plugins/projects/serial/src/lib/views/dialog/components.ts
+++ b/application/client.plugins/projects/serial/src/lib/views/dialog/components.ts
@@ -5,17 +5,8 @@ import { IPortInfo, IPortState } from '../../common/interface.portinfo';
 import { IOptions, CDefaultOptions } from '../../common/interface.options';
 import { SidebarVerticalPortOptionsWriteComponent } from '../sidebar.vertical/port.options.write/component';
 import { EHostEvents } from '../../common/host.events';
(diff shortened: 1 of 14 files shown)


application/client.plugins/projects/serial/src/lib/views/dialog/components.ts
```diff
@@ -52,21 +55,46 @@ export class SidebarVerticalPortDialogComponent implements OnInit, OnDestroy {
...
         });
+        this._interval = setInterval(() => {
+            this._subjects.event.next(true);
+        }, this._timeout);
     }
 
     ngOnDestroy() {
+        if (this._interval) {
+            clearTimeout(this._interval);
+        }
```
DmitryAstafyev: doesn't necessary to check. it will not be an error, clearTimeout "eats" everything without errors. So you can leave just line 76

application/client.plugins/projects/serial/src/lib/views/dialog/components.ts
```diff
@@ -52,21 +55,46 @@ export class SidebarVerticalPortDialogComponent implements OnInit, OnDestroy {
...
+        Object.keys(this._subscriptions).forEach((key: string) => {
+            this._subscriptions[key].unsubscribe();
+        });
         this._destroyed = true;
     }
 
+    public observe(): { event: Observable<any> } {
+        return { event: this._subjects.event.asObservable() };
+    }
```
DmitryAstafyev: could you please name somehow events (onData, onPortData, onCloseDialog and so on). It allows better understand a code

application/client.plugins/projects/serial/src/lib/views/dialog/port.available/components.ts
```diff
@@ -0,0 +1,197 @@
...
+    port: IPortInfo;
+    options: IOptions;
+    state: IPortState;
+}
+
+interface Irgb {
+    red: number;
+    green: number;
+    blue: number;
+    opacity: 1;
```
DmitryAstafyev: try to use in an interface only types, but not values (opacity: 1)

application/client.plugins/projects/serial/src/lib/views/dialog/port.available/components.ts
```diff
@@ -0,0 +1,197 @@
...
+    private _destroyed: boolean = false;
+
+    constructor(private _cdRef: ChangeDetectorRef) {
+    }
+
+    ngAfterViewInit() {
+        if (!this._ng_isConnected(this.port)) {
+            this._read = 0;
+            this._canvas = document.getElementById(`canvas_${this.port.path}`) as HTMLCanvasElement;
+            this._createChart().then(() => {
```
DmitryAstafyev: Please use @ViewChild and ElementRef from Angular to get access to node (canvas in your case). Try to avoid from direct calling DOM while we are within Angular

application/client.plugins/projects/serial/src/lib/views/dialog/port.available/components.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    private _createChart(): Promise<void> {
+        return new Promise((resolve, reject) => {
+            this._ctx = this._canvas.getContext('2d');
+            if (this._chart) {
+                reject(`Chart for ${this.port.path} already exists!`);
+            } else {
+                this._chart = new Chart(this._ctx, {
+                    type: 'line',
+                    data: {
```
DmitryAstafyev: This method doesn't have any asynchronous operations, it might me not a promise, but return Error (in case of error) or undefined (in case of success).

application/client.plugins/projects/serial/src/lib/views/dialog/components.ts
```diff
@@ -52,21 +55,46 @@ export class SidebarVerticalPortDialogComponent implements OnInit, OnDestroy {
             if (typeof message !== 'object' && message === null) {
                 return;
             }
+            if (message) {
+                this._onSpyState(message);
+            }
```
DmitryAstafyev: this is not your 
```
if (typeof message !== 'object' && message === null) {
                return;
}
```
but it's a mistake. 
should be
```
if (typeof message !== 'object' || message === null) {
                return;
}
this._onSpyState(message);

```
after this correction you can remove checking of message

PR #632, opened 2020-01-07
Manage recently connected ports

fur-gan:
Popup window show recently connected ports (via config file saved on user PC)
- directly connect to port
- remove individual port
- remove all ports

Notification on error/warning (serial plugin)
- show error as notfication and log it

### DIFF
--- a/application/client.plugins/projects/serial/src/lib/services/service.ts
+++ b/application/client.plugins/projects/serial/src/lib/services/service.ts
@@ -32,7 +32,7 @@ export class Service extends Toolkit.APluginService {
     private _onAPIReady() {
         this.api = this.getAPI();
         if (this.api === undefined) {
-            this._logger.error('API not found!');
+            this._notify('error', 'API not found', ENotificationType.error);
             return;
         }
         this._subscriptions.onSessionOpen = this.api.getSessionsEventsHub().subscribe().onSessionOpen(this._onSessionOpen.bind(this));
@@ -58,6 +58,17 @@ export class Service extends Toolkit.APluginService {
 
     private _onSessionChange(guid: string) {
         this.session = guid;
+    }
+
+    private _notify(caption: string, message: string, type: ENotificationType) {
+        this.notify(caption, message, type);
+        if (type === ENotificationType.error) {
+            this._logger.error(message);
+        } else if (type === ENotificationType.warning) {
+            this._logger.warn(message);
+        } else {
+            this._logger.info(message);
+        }
     }
 
     public getObservable(): {
@@ -88,7 +99,7 @@ export class Service extends Toolkit.APluginService {
                     this.state = response;
                     this._subjects.event.next(message);
                 }).catch((error: Error) => {
-                    this._logger.error(error);
+                    this._notify('error', error.message, ENotificationType.error);
                 });
                 return;
             }
@@ -111,7 +122,7 @@ export class Service extends Toolkit.APluginService {
                 resolve();
             }
         }).catch((error: Error) => {
-            this._logger.error(error);
+            this._notify('error', `Failed to save read load of ports: ${error.message}`, ENotificationType.error);
         });
     }
 
@@ -139,7 +150,7 @@ export class Service extends Toolkit.APluginService {
             this._openQueue[options.path] = true;
             this.emptyQueue(options.path);
         }).catch((error: Error) => {
-            this._logger.error(error);
+            this._notify('error', `Failed to connect to ${options.path}: ${error.message}`, ENotificationType.error);
         });
     }
 
@@ -152,7 +163,7 @@ export class Service extends Toolkit.APluginService {
             this._openQueue[port] = false;
             this.sessionConnected[this.session][port] = undefined;
         }).catch((error: Error) => {
-            this._logger.error(error);
+            this._notify('error', `Failed to disconnect from ${port}: ${error.message}`, ENotificationType.error);
         });
     }
 
@@ -161,7 +172,7 @@ export class Service extends Toolkit.APluginService {
             stream: this.session,
             command: EHostCommands.list,
         }, this.session).catch((error: Error) => {
-            this._logger.error(error);
+            this._notify('error', `Failed to request port list: ${error.message}`, ENotificationType.error);
         });
     }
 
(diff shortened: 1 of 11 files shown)


application/client.plugins/projects/serial/src/lib/services/service.ts
```diff
@@ -87,7 +88,7 @@ export class Service extends Toolkit.APluginService {
                     this.state = response;
                     this._subjects.event.next(message);
                 }).catch((error: Error) => {
-                    this._logger.error(error);
+                    throw(error);
```
DmitryAstafyev: using `throw` will drop an exception and break the whole JS thread. In addition, because you are calling `throw` from a promise it will be uncatched promise exception.

Why we need throw here?

application/client.plugins/projects/serial/src/lib/services/service.ts
```diff
@@ -100,7 +101,7 @@ export class Service extends Toolkit.APluginService {
             if (Object.keys(this.sessionConnected).length > 0) {
                 Object.keys(this.sessionConnected).forEach(session => {
                     Object.keys(this.sessionConnected[session]).forEach(port => {
-                        if (ports[port]) {
+                        if (ports[port] && this.sessionConnected[session][port]) {
```
DmitryAstafyev: it doesn't make sense, because you already take only existing keys with ```Object.keys(this.sessionConnected[session])```

application/client.plugins/projects/serial/src/lib/services/service.ts
```diff
@@ -170,28 +171,62 @@ export class Service extends Toolkit.APluginService {
...
+                stream: this.session,
+                command: EHostCommands.spyStop,
+                options: options,
+            }, this.session).catch((error: Error) => {
+                throw(error);
+            });
+        }
+        return new Promise((resolve) => {
+            resolve();
+        });
```
DmitryAstafyev: if you need just resolve promise with returning, better to use:
```
return Promise.resolve();
```
because it's more clean

application/client.plugins/projects/serial/src/lib/services/service.ts
```diff
@@ -204,8 +239,29 @@ export class Service extends Toolkit.APluginService {
...
+                type: type
+            }
+        });
+        if (type === ENotificationType.error) {
+            this._logger.error(message);
+        } else if (type === ENotificationType.warning) {
+            this._logger.warn(message);
+        } else {
+            this._logger.info(message);
+        }
```
DmitryAstafyev: You are using method `notify` outside (like `Service.notify`), but API could be available. So to prevent any errors (not available API), better add here checks for it, like `if (this.api === undefined) { ... }`

application/client.plugins/projects/serial/src/lib/services/service.ts
DmitryAstafyev: Also in many other places of code you have `throw`. We should get rid of it or if it's really needed wrap it into `try {} catch {}` to prevent breaking of the whole workflow

DmitryAstafyev: I would do more. Because it's front-end (render), the plugin process is included in the main core JS thread. It means `throw` will break also the core process. So, I would add a strict rule for plugin's developing - using of unwrapped throws are prohibited at all on render side. 

So it could be reflected in documentations.

application/sandbox/serial/process/src/services/service.sessions.ts
```diff
@@ -17,6 +44,7 @@ class ServiceSessions {
         PluginIPCService.subscribe(IPCMessages.PluginInternalMessage, this._onIncomeRenderIPCMessage);
         PluginIPCService.on(PluginIPCService.Events.openStream, this._onOpenStream);
         PluginIPCService.on(PluginIPCService.Events.closeStream, this._onCloseStream);
+        ServiceConfig.setDefault<{[ports:string]: {[port: string]: IPortOptions}}>( { ports: {} } );
```
DmitryAstafyev: I would recommend to create an interface for settings and add one more level, like
```
interface IRecentPortData {
    [ports: string]: {[port: string]: IPortOptions };
}

interface IPluginSettings {
    recent: IRecentPortData;
}
```

Also, not clear, what does mean `ports:string`. Is it some kind of hash? or?

application/sandbox/serial/process/src/services/service.sessions.ts
```diff
@@ -98,8 +126,8 @@ class ServiceSessions {
                         }
                     }));
                 });
-            case ECommands.write:
-                return this._income_onWrite(message).then(() => {
+            case ECommands.send:
+                return this._income_onSend(message).then(() => {
```
DmitryAstafyev: section `catch` in missed on promise

application/sandbox/serial/process/src/services/service.sessions.ts
```diff
@@ -318,6 +403,73 @@ class ServiceSessions {
...
+                });
+            }).catch((error: Error) => {
+                this._logger.error(`Failed to remove settings due to error: ${error.message}`);
+            });
+        });
+    }
+
+    private _updateConfig(options: IOptions): Promise<{}> {
+        return new Promise((resolve, reject)=> {
+            ServiceConfig.read<{[key: string]: any}>().then((settings: {[key: string]: any}) => {
```
DmitryAstafyev: See the previous comment. After you will create an interface for settings, you can use it also here `<...>`

application/sandbox/serial/process/src/services/service.sessions.ts
fur-gan: `{ "ports":                  // <-- [ports: string]
    { "COM4": { ... } }
    { "COM1": { ... } }
    { "COM5": { ... } }
}`

fur-gan: Above is the content of the config file, so the `ports:string` is the key for it to read the ports

fur-gan: `interface IPluginSettings {`
`    recent: IRecentPortData;`
`}`
I'm not sure where this interface would help

application/sandbox/serial/process/src/services/service.sessions.ts
fur-gan: There is a `catch `section is in line 138, or what do you exactly mean?

application/sandbox/serial/process/src/services/service.sessions.ts
fur-gan: Would it maybe not be better to do it like this:
`interface IRecentPortData {`
`    [port: string]: IOptions;`
`}`
`interface IPluginSettings {`
`    recent: IRecentPortData;`
`}`

PR #668, opened 2020-01-21
Fixed signing for OSX

meatlink:
Fixing issues with code signing and notarising on OSX


marcmo:
the lint warning will be fixed in a separate commit and has nothing to do with this fix

Issue #479, opened 2019-10-01
support drag & drop with folder


marcmo:
for file concatenation or merging this is convenient

PR #805, opened 2020-05-12
[#479](feat) Support drag&drop with folder

fur-gan:


application/client.core/src/app/environment/services/service.file.opener.ts
```diff
@@ -69,47 +69,55 @@ export class FileOpenerService implements IService, IFileOpenerService {
...
-                    id: 'opening-file-dialog',
-                    caption: `Opening files`,
-                    component: {
-                        factory: DialogsMultipleFilesActionComponent,
-                        inputs: {
-                            files: files
+            ServiceElectronIpc.request(new IPCMessages.FileCheckRequest({
+                files: files.map((file: IFile) => file.path),
+                guid: session.getGuid(),
+            }), IPCMessages.FileCheckResponse).then((checkResponse: IPCMessages.FileCheckResponse) => {
```
DmitryAstafyev, line 80: If there are no empty session `_setSessionForFile`  will create a new session, but it `IPCMessages.FileCheckRequest` will return error; session will stay empty. 

From this point of view, better call `_setSessionForFile` after a list of files was gotten. I mean first `IPCMessages.FileCheckRequest` and after `_setSessionForFile`  if no error with previous.

application/common/ipc/electron.ipc.messages/file.check.request.ts
```diff
@@ -0,0 +1,30 @@
+export interface IFileCheckRequest {
+    guid: string;
```
DmitryAstafyev, line 2: I'm not sure, we need session `guid` for such a procedure.

application/common/ipc/electron.ipc.messages/file.check.request.ts
```diff
@@ -0,0 +1,30 @@
...
+    files: string[];
+}
+
+export class FileCheckRequest {
+    public static signature: string = 'FileCheckRequest';
+    public signature: string = FileCheckRequest.signature;
```
DmitryAstafyev, line 8: Probably `FileCheckRequest` isn't a self-explaining name. Actually we do not check files, but we get a "flat" list of files from files/folders. Please use some other a self-explaining name of IPC message (and IPC file)

application/common/ipc/electron.ipc.messages/file.check.response.ts
```diff
@@ -0,0 +1,44 @@
+export interface IFileCheckResponse {
+    guid: string;
```
DmitryAstafyev, line 2: No need session

application/common/ipc/electron.ipc.messages/file.check.response.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+interface IFile {
+    lastModified: number;
+    lastModifiedDate: Date;
+    name: string;
+    path: string;
+    size: number;
+    type: string;
+}
```
DmitryAstafyev, line 14: export type also, and add type into `index.ts`

application/common/ipc/electron.ipc.messages/index.ts
```diff
@@ -202,6 +202,12 @@ export { IFileInfoRequest, FileInfoRequest };
 import { IFileInfoResponse, FileInfoResponse } from './file.fileinfo.response';
 export { IFileInfoResponse, FileInfoResponse };
 
+import { IFileCheckRequest, FileCheckRequest } from './file.check.request';
+export { IFileCheckRequest, FileCheckRequest };
+
+import { IFileCheckResponse, FileCheckResponse } from './file.check.response';
+export { IFileCheckResponse, FileCheckResponse };
```
DmitryAstafyev, line 209: add `IFile`

application/electron/src/services/files/service.file.opener.ts
```diff
@@ -22,6 +22,15 @@ interface IOpenFileResult {
...
 
+interface IFile {
+    lastModified: number;
+    lastModifiedDate: Date;
+    name: string;
+    path: string;
+    size: number;
+    type: string;
+}
+
```
DmitryAstafyev, line 33: Reuse `IFile` definition from IPCMessages store

application/electron/src/services/files/service.file.opener.ts
```diff
@@ -228,6 +260,44 @@ class ServiceFileOpener implements IService {
...
+                type: 'file',
+            });
+        } else if (!(fs.lstatSync(file).isDirectory())) {
+            // Other
+            return;
+        } else {
+            // Directory
+            fs.readdirSync(file).forEach((subFile: string) => this._listFiles(file + '/' + subFile));
+        }
+    }
```
DmitryAstafyev, line 291: With IPC handlers do not use any IO sync operation. It will block main JS thread. From user point of view we could get frozen UI. Use only async operations. For example instead `fs.statSync`, use `fs.stat`.

application/electron/src/services/files/service.file.opener.ts
```diff
@@ -198,6 +211,25 @@ class ServiceFileOpener implements IService {
...
+    private _ipc_FileCheckRequest(request: IPCMessages.TMessage, response: (instance: IPCMessages.TMessage) => any) {
+        const req: IPCMessages.FileCheckRequest = request as IPCMessages.FileCheckRequest;
+        this._getFiles(req.files).then(() => {
+            response(new IPCMessages.FileCheckResponse({
+                files: this._files.slice(),
+                guid: req.guid,
+            })).catch((error: Error) => {
+                this._logger.error(`Fail to respond to files ${req.files} due error: ${error.message}`);
+            });
+            this._files = [];
```
DmitryAstafyev, line 223: it doesn't make sense to use variable `_files` to deliver results of the method `_getFiles`. But in this context, it's also dangerous, because it's IPC, IPC works async always. It means could be more than 2 requests at the same time at work. It means the variable `_files` will used multiple times at the same time. 

Better just return results from `_getFiles`

application/electron/src/services/files/service.file.opener.ts
```diff
@@ -228,6 +256,94 @@ class ServiceFileOpener implements IService {
...
+            });
+        });
+    }
+
+    private _getName(file: string): string {
+        if (file.trim() === '') {
+            return '';
+        }
+        return path.basename(file);
+    }
```
DmitryAstafyev, line 345: Do we really need this method? Probably we can just call `path.basename` where it would be needed?

application/electron/src/services/files/service.file.opener.ts
```diff
@@ -228,6 +256,94 @@ class ServiceFileOpener implements IService {
...
+                        } else {
+                            // Directory
+                            return fs.readdir(file, (err, files) => {
+                                if (err) {
+                                    const errorMessage = `Fail to list files of directory ${file} due to error: ${err.message}`;
+                                    ServiceNotifications.notify({
+                                        caption: 'Error with directory',
+                                        message: errorMessage,
+                                        type: ENotificationType.warning,
+                                    });
```
DmitryAstafyev, line 316: We cannot use notifications here because:
- In the case of a request from front-end (render) notifications could be used. Render could show it by itself
- Task of service (in our case IPC request): build and return list of file. Only this. We should not do any other actions, which weren't required by IPC request.

application/client.core/src/app/environment/services/service.file.opener.ts
```diff
@@ -68,50 +68,61 @@ export class FileOpenerService implements IService, IFileOpenerService {
...
+                files: files.map((file: IFile) => file.path),
+            }), IPCMessages.FileListResponse).then((checkResponse: IPCMessages.FileListResponse) => {
+                this._setSessionForFile().then((session: ControllerSessionTab) => {
+                    TabsSessionsService.setActive(session.getGuid());
+                    if (checkResponse.error !== undefined) {
+                        const errorMessage = `Fail check paths due error: ${checkResponse.error}`;
+                        this._logger.error(errorMessage);
+                        reject(new Error(errorMessage));
+                    }
+                    if (checkResponse.files.length === 1) {
```
DmitryAstafyev, line 86: function will be rejected, but not stopped. JS will go to line 86 after `reject`

application/client.core/src/app/environment/services/service.file.opener.ts
```diff
@@ -68,50 +68,61 @@ export class FileOpenerService implements IService, IFileOpenerService {
...
+            }
+            ServiceElectronIpc.request(new IPCMessages.FileListRequest({
+                files: files.map((file: IFile) => file.path),
+            }), IPCMessages.FileListResponse).then((checkResponse: IPCMessages.FileListResponse) => {
+                this._setSessionForFile().then((session: ControllerSessionTab) => {
+                    TabsSessionsService.setActive(session.getGuid());
+                    if (checkResponse.error !== undefined) {
+                        const errorMessage = `Fail check paths due error: ${checkResponse.error}`;
+                        this._logger.error(errorMessage);
+                        reject(new Error(errorMessage));
```
DmitryAstafyev, line 84: Can be done like it
```
reject(new Error(this._logger.error(`Fail check paths due error: ${checkResponse.error}`)));
```

application/client.core/src/app/environment/services/service.file.opener.ts
```diff
@@ -68,50 +68,61 @@ export class FileOpenerService implements IService, IFileOpenerService {
...
+                    if (checkResponse.files.length === 1) {
+                        // Single file
+                        ServiceElectronIpc.request(new IPCMessages.FileOpenRequest({
+                            file: checkResponse.files[0].path,
+                            session: session.getGuid(),
+                        }), IPCMessages.FileOpenResponse).then((openResponse: IPCMessages.FileReadResponse) => {
+                            if (openResponse.error !== undefined) {
+                                const errorMessage = `Fail open file/folder "${files[0].path}" due error: ${openResponse.error}`;
+                                this._logger.error(errorMessage);
+                                reject(new Error(errorMessage));
```
DmitryAstafyev, line 95: no return

application/client.core/src/app/environment/services/service.file.opener.ts
```diff
@@ -68,50 +68,61 @@ export class FileOpenerService implements IService, IFileOpenerService {
...
+                            }
+                        }).catch((error: Error) => {
+                            const errorMessage = `Fail open file/folder "${files[0].path}" due error: ${error.message}`;
+                            this._logger.error(errorMessage);
+                            reject(new Error(errorMessage));
```
DmitryAstafyev, line 100: no return

application/client.core/src/app/environment/services/service.file.opener.ts
```diff
@@ -68,50 +68,61 @@ export class FileOpenerService implements IService, IFileOpenerService {
...
-                        { caption: 'Merge', handler: this.merge.bind(this, files), },
-                        { caption: 'Concat', handler: this.concat.bind(this, files), },
-                    ],
                 });
-            }
-        }).catch((error: Error) => {
-            this._logger.error(`Cannot continue with opening file, because fail to prepare session due error: ${error.message}`);
+            }).catch((error: Error) => {
+                this._logger.error(`Cannot continue with opening file, because fail to prepare session due error: ${error.message}`);
+            });
```
DmitryAstafyev, line 125: no actions in catch

application/client.core/src/app/environment/services/service.file.opener.ts
```diff
@@ -68,50 +68,61 @@ export class FileOpenerService implements IService, IFileOpenerService {
...
-                ServiceElectronIpc.request(new IPCMessages.FileOpenRequest({
-                    file: files[0].path,
-                    session: session.getGuid(),
-                }), IPCMessages.FileOpenResponse).then((response: IPCMessages.FileReadResponse) => {
-                    if (response.error !== undefined) {
-                        this._logger.error(`Fail open file "${files[0].path}" due error: ${response.error}`);
-                        // TODO: add notification here
-                        return;
+    public open(files: IFile[]): Promise<Error> {
+        return new Promise((resolve, reject) => {
```
DmitryAstafyev, line 72: no `resolve` call

application/electron/src/services/files/service.file.opener.ts
```diff
@@ -228,6 +255,71 @@ class ServiceFileOpener implements IService {
...
+                                if (err) {
+                                    const errorMessage = `Fail to list files of directory ${file} due to error: ${err.message}`;
+                                    self._logger.warn(errorMessage);
+                                    rejected(new Error(errorMessage));
+                                } else {
+                                    Promise.all(files.map((subFile: string) => {
+                                        return listAllFiles(file + '/' + subFile);
+                                    })).then(() => {
+                                        resolve(allFiles);
+                                    }).catch((error: Error) => {
```
DmitryAstafyev, line 306: probably `resolved`?

Issue #797, opened 2020-04-29
Reported version of Platform is incorrect on About tab


marcmo:
This is what I get:
```
Version: 2.2.0
Platform: darwin(x64) / Darwin / 19.4.0
Electron: 8.1.0
...
```
but my OS version is 10.15.4
![image](https://user-images.githubusercontent.com/99541/80567581-c26ae680-89f5-11ea-9779-1c4f5119d1df.png)

Issue #745, opened 2020-03-18
when merging or concatenating with an empty file, error in log


marcmo:
empty files should be ignored

Issue #665, opened 2020-01-20
Search input: support TAB


DmitryAstafyev:
If autocomplete list has some values, by pressing TAB the first match should be selected

marcmo:
related #704

PR #831, opened 2020-06-22
Solution for [#745][#665][#797]

fur-gan:
- Disable the checkbox for empty files when drag&drop multiple files in the dialog window
- support TAB when selecting recently used search terms
- Fix release version detection for Linux and Mac OS

### DIFF
--- a/application/electron/src/controllers/electron/menu.functions/handler.item.about.ts
+++ b/application/electron/src/controllers/electron/menu.functions/handler.item.about.ts
@@ -25,7 +25,7 @@ export default function about() {
     const platform: string = os.platform();
     if (platform === 'darwin' || platform === 'linux') {
         return exec('uname -v', (error: ExecException | null, stdout: string, stderr: string) => {
-            if (error || stderr.trim() === '') {
+            if (error || stderr.trim() !== '') {
                 error ? logger.error(error.message) : logger.error(stderr);
                 return _send('');
             }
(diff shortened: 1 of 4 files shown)


application/electron/src/controllers/electron/menu.functions/handler.item.about.ts
```diff
@@ -1,11 +1,31 @@
...
+const logger: Logger = new Logger('HandlerItemAbout');
+
 export default function about() {
+    const platform: string = os.platform();
+    if (platform === 'darwin' || platform === 'linux') {
+        return exec('uname -v', (error: ExecException | null, stdout: string, stderr: string) => {
+            if (error) {
+                logger.error(error.message);
+                return;
+            }
```
DmitryAstafyev, line 17: What we will send as `platform` in case of error?

application/electron/src/controllers/electron/menu.functions/handler.item.about.ts
```diff
@@ -1,11 +1,31 @@
...
+            if (stderr.trim() === '') {
+                logger.error(stderr);
+            }
```
DmitryAstafyev, line 20: Why we do not "exit" on error? We still can trust output on called command even it has something in `stderr`?

application/electron/src/controllers/electron/menu.functions/handler.item.about.ts
```diff
@@ -1,11 +1,31 @@
...
+            const regex = /\d{1,}\.\d{1,}\.\d{1,}/;
+            ServiceElectron.IPC.send(new IPCMessages.TabCustomAbout({
+                version: ServicePackage.get().version,
+                dependencies: ServicePackage.get().chipmunk.versions,
+                platform: `${os.platform()}(${os.arch()}) / ${os.type()} / ${stdout.match(regex)}`,
```
DmitryAstafyev, line 25: what if `stdout.match(regex)` will be `null`? For example, because `stderr !== ''`, stdout might be just empty.

PR #864, opened 2020-08-07
Filter/Chart/Disabled features on new structure

fur-gan:


application/client.core/src/app/environment/components/sidebar/search.manager/disabled/provider.ts
```diff
@@ -154,13 +163,35 @@ export class ProviderDisabled extends Provider<DisabledRequest> {
...
+                    request: (entity.getEntity().entity as ChartRequest).asDesc().request,
+                    flags: {
+                        casesensitive: false,
+                        wholeword: false,
+                        regexp: true,
+                    }
+                }));
+            } else if (entity.getEntity().entity instanceof FilterRequest) {
+                super.getSession().getSessionSearch().search(entity.getEntity().entity);
+            }
```
DmitryAstafyev, line 189: Please, check, because Filter, Chart, Range and Disabled don't have public property `entity`. Looks like typo.

PR #878, opened 2020-08-19
Add bin to remove entities and fixes

fur-gan:


application/client.core/src/app/environment/components/sidebar/search.manager/charts/list/component.ts
```diff
@@ -57,6 +62,18 @@ export class SidebarAppSearchManagerChartsComponent implements OnDestroy, AfterC
...
+        this._dragging = entity;
+        SearchManagerService.onDragStart(true);
+    }
+
+    private _onRemove() {
+        if (this._dragging) {
+            this._dragging.getEntity().remove(this.provider.getSession());
+        }
+        this._dragging = undefined;
+    }
```
DmitryAstafyev, line 75: This code is repeated for filters, charts, time ranges, and disabled entities. 
Let's try to move out this code into `directive`. It will allow inherit it and reuse for each entity.

DmitryAstafyev (changes requested):

application/client.core/src/app/environment/components/sidebar/search.manager/ranges/list/component.ts
```diff
@@ -95,6 +108,24 @@ export class SidebarAppSearchManagerTimeRangesComponent implements OnDestroy, Af
...
+        this._droppedOut = false;
+    }
+
+    private _onMouseOverGlobal() {
+        if (!this._ignore) {
+            this._droppedOut = true;
+        } else {
+            this._ignore = false;
+        }
+    }
```
DmitryAstafyev, line 127: can be moved into `directive`

application/client.core/src/app/environment/components/sidebar/search.manager/bin/component.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    private _onMouseOverGlobal() {
+        if (this._ng_dragging && !this._ignore) {
+            this._droppedOut = true;
+        } else {
+            this._ignore = false;
+        }
+    }
```
DmitryAstafyev, line 79: can be moved into `directive`

application/client.core/src/app/environment/components/sidebar/search.manager/charts/list/component.ts
```diff
@@ -57,6 +70,24 @@ export class SidebarAppSearchManagerChartsComponent implements OnDestroy, AfterC
...
+        this._droppedOut = false;
+    }
+
+    private _onMouseOverGlobal() {
+        if (!this._ignore) {
+            this._droppedOut = true;
+        } else {
+            this._ignore = false;
+        }
+    }
```
DmitryAstafyev, line 89: can be moved into `directive`

application/client.core/src/app/environment/components/sidebar/search.manager/disabled/list/component.ts
```diff
@@ -60,6 +73,24 @@ export class SidebarAppSearchManagerDisabledsComponent implements OnDestroy, Aft
...
+        this._droppedOut = false;
+    }
+
+    private _onMouseOverGlobal() {
+        if (!this._ignore) {
+            this._droppedOut = true;
+        } else {
+            this._ignore = false;
+        }
+    }
```
DmitryAstafyev, line 92: can be moved into `directive`

application/client.core/src/app/environment/components/sidebar/search.manager/filters/list/component.ts
```diff
@@ -57,6 +70,24 @@ export class SidebarAppSearchManagerFiltersComponent implements OnDestroy, After
...
+        this._droppedOut = false;
+    }
+
+    private _onMouseOverGlobal() {
+        if (!this._ignore) {
+            this._droppedOut = true;
+        } else {
+            this._ignore = false;
+        }
+    }
```
DmitryAstafyev, line 89: can be moved into `directive`

application/client.core/src/app/environment/components/sidebar/search.manager/bin/component.ts
```diff
@@ -0,0 +1,81 @@
...
+            this._droppedOut = false;
+        }
+    }
+
+    private _onMouseOverGlobal() {
+        if (this._ng_dragging && !this._ignore) {
+            this._droppedOut = true;
+        } else {
+            this._ignore = false;
+        }
```
DmitryAstafyev, line 78: can be moved into `directive`

DmitryAstafyev (changes requested):

DmitryAstafyev:
About animation issue (I mean when you drop item above "wrong" place and it moves to it, even would not be held there).
I didn't find Angular API for that. Probably Angular doesn't have it.
But we still can fix it with one trick.
As soon as item dragged, Angular creates placeholder. Placeholder is like a magnet for a dragged item. As soon as you stop dragging an item, it will move to a placeholder. 
Well if we move placeholder - we will change a target for angular animation without breaking internal mechanisms inside a framework.
Here is very-very dirty example. It should be on draggable item

```
// We cannot listen event cdkDragStart, because on this event we don't have yet placeholder.
// As soon as placeholder created by angular all we need inject our marker to remember position of it in DOM tree
public _ng_cdkDragMoved(event: CdkDragMove) {
        // If marker already created - ignore
        if (this._marker !== undefined) {
            return;
        }
        // Here what we need - placeholder
        const ref = event.source._dragRef as any;
        const ph = ref._placeholder as HTMLElement;
        if (ph === undefined) {
            return;
        }
        // create and inject marker
        const marker = document.createElement('span');
        marker.style.display = 'none';
        ph.parentElement.insertBefore(marker, ph);
        // remember marker
        this._marker = marker;
    }

    // Okay. Now when dragging is over, we should move placeholder (never mind where is it now) to initial place
    public _ng_cdkDragReleased(event: CdkDragRelease) {
        if (this._marker === undefined) {
            console.error(`Weird. Marker should be removed before.`);
            return;
        }
        // Get ref to new placeholder
        const ref = event.source._dragRef as any;
        const ph = ref._placeholder as HTMLElement;
        // Move placeholder to inital place in DOM tree
        this._marker.parentNode.insertBefore(ph, this._marker);
        // Remove marker
        this._marker.parentNode.removeChild(this._marker);
        this._marker = undefined;
        ph.style.transform = '';
    }
``` 
This solution fixes animation issue as well.

But you have to change know - when it can be applied, when - not, so, service above all provider should be upgraded.

And sure, this solution should be implemented on `directive` level, but not on level of each component

DmitryAstafyev:
Here is a suggestion about fixing reset-animation-issue:

[drag_drop_anim_suggestion.txt](https://github.com/esrlabs/chipmunk/files/5135690/drag_drop_anim_suggestion.txt)

application/client.core/src/app/environment/components/sidebar/search.manager/directives/item.directive.ts
```diff
@@ -1,30 +1,50 @@
...
+
 
 @Directive({
     selector: '[appSidebarSearchManagerItem]',
 })
 
 export class SidebarAppSearchManagerItemDirective implements OnInit, OnDestroy {
 
+    @ViewChild(MatDragDropResetFeatureDirective) accessor; // First
+
```
DmitryAstafyev, line 18: Please remove the unused block. It's trash from my patch. Sorry, I've forgotten to remove it.

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

PR #895, opened 2020-08-31
New features and fixes for Drag and Drop

fur-gan:


application/client.core/src/app/environment/services/standalone/service.popups.ts
```diff
@@ -72,6 +74,17 @@ export class PopupsService {
...
     }
 
+    private _onMouseUp(event: MouseEvent) {
+        if (event.button !== 0) {
+            return;
+        }
+        if (this._opened.size === 0) {
+            return;
+        }
+        this.remove(this._opened.keys().next().value);
```
DmitryAstafyev, line 84: I think it will close any dialog without action. Dialog should be closed only on click on background, but not on dialogs as itself.

DmitryAstafyev (changes requested):

DmitryAstafyev:
Please, take a look. Just tested (arch linux): dragging between charts & filters doesn't work for me.

application/client.core/src/app/environment/components/sidebar/search.manager/charts/list/component.ts
```diff
@@ -41,21 +44,32 @@ export class SidebarAppSearchManagerChartsComponent implements OnDestroy, AfterC
...
         const prev = event.previousContainer;
         const index = event.previousIndex;
-        if (prev.data !== undefined && (prev.data as any).disabled !== undefined) {
-            const outside: Entity<any> | undefined = (prev.data as any).disabled[event.previousIndex] !== undefined ? (prev.data as any).disabled[index] : undefined;
-            if (outside !== undefined && typeof outside.getEntity().getEntity === 'function' && outside.getEntity().getEntity() instanceof ChartRequest) {
-                this.provider.getSession().getSessionSearch().getDisabledAPI().getStorage().remove(outside.getEntity());
-                this.provider.getSession().getSessionSearch().getChartsAPI().getStorage().add(outside.getEntity().getEntity(), event.currentIndex);
+        if (prev.data !== undefined) {
+            if ((prev.data as any).disabled !== undefined) {
+                const outside: Entity<any> | undefined = (prev.data as any).disabled[event.previousIndex] !== undefined ? (prev.data as any).disabled[index] : undefined;
```
DmitryAstafyev, line 56: Can we provide `interface` for `prev.data` to get rid of `any`?

application/client.core/src/app/environment/components/sidebar/search.manager/charts/list/component.ts
```diff
@@ -41,21 +44,32 @@ export class SidebarAppSearchManagerChartsComponent implements OnDestroy, AfterC
...
+                }
+            } else {
+                const outside: Entity<FilterRequest> | undefined = prev.data[event.previousIndex] !== undefined ? prev.data[index] : undefined;
+                if (outside !== undefined && typeof outside.getEntity === 'function' && outside.getEntity() instanceof FilterRequest) {
+                    this.provider.getSession().getSessionSearch().getFiltersAPI().getStorage().remove(outside.getEntity());
+                    this.provider.getSession().getSessionSearch().getChartsAPI().getStorage().add({
+                        request: outside.getEntity().asDesc().request,
+                        type: EChartType.smooth,
+                    });
+                }
```
DmitryAstafyev, line 69: I am not sure, what is going on here, but just as reminder - filter can be converted into chart only if filter is regular expression and has a group; for example, `cpu=(\d)` can be converted to chart because has group` (\d)`, but `cpu=\d` cannot, because there are no group.

Controller of chart `ChartRequest` (application/client.core/src/app/environment/controller/controller.session.tab.search.charts.request.ts) has a static method `isValid`, which can be (and should be) used to check possobility of converting

DmitryAstafyev (changes requested):

application/client.core/src/app/environment/components/sidebar/search.manager/charts/list/component.ts
DmitryAstafyev: In general, I'm not sure - it's the right place to do converting. I believe it should be on provider level. We have abstract class for `provider`, we can implement necessary methods there. Like `createFrom` (or just `from`) and `isConvertable(entity): bool`.

fur-gan: > I am not sure, what is going on here, but just as reminder - filter can be converted into chart only if filter is regular expression and has a group; for example, `cpu=(\d)` can be converted to chart because has group` (\d)`, but `cpu=\d` cannot, because there are no group.
> 
> Controller of chart `ChartRequest` (application/client.core/src/app/environment/controller/controller.session.tab.search.charts.request.ts) has a static method `isValid`, which can be (and should be) used to check possobility of converting

Yeah, I check the validity when I drag the item so I disable the dropping into the list with `_ng_viablePredicate`

fur-gan: > In general, I'm not sure - it's the right place to do converting. I believe it should be on provider level. We have abstract class for `provider`, we can implement necessary methods there. Like `createFrom` (or just `from`) and `isConvertable(entity): bool`.

Alright, will move that into the provider class of each list

DmitryAstafyev (approved):

PR #900, opened 2020-09-03
Refactoring droplist, Rework DLT column formatting options

fur-gan:


application/client.core/src/app/environment/layout/contextmenu/component.ts
```diff
@@ -37,11 +37,13 @@ export class LayoutContextMenuComponent implements OnDestroy, AfterViewInit {
...
-        this._unsubscribeToWinEvents();
+        ContextMenuService.unsubscribeToWinEvents(EEventType.keydown, this._onWindowKeyDown.bind(this));
+        ContextMenuService.unsubscribeToWinEvents(EEventType.mousedown, this._onWindowMouseDown.bind(this));
     }
 
     public ngAfterViewInit() {
-        this._subscribeToWinEvents();
+        ContextMenuService.subscribeToWinEvents(EEventType.keydown, this._onWindowKeyDown.bind(this));
+        ContextMenuService.subscribeToWinEvents(EEventType.mousedown, this._onWindowMouseDown.bind(this));
     }
```
DmitryAstafyev, line 47: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind

> The bind() method creates a new function that, when called, has its this keyword set to the provided value, with a given sequence of arguments preceding any provided when the new function is called.

It means you will not unsubscribe events, because you push **new function** there, but not existed handler

application/client.core/src/app/environment/components/complex/popups/directive/popups.directive.ts
```diff
@@ -0,0 +1,14 @@
...
+    selector: '[appComplexPopups]',
+})
+
+export class ComplexPopupsDirective implements OnInit {
+
+    constructor(private element: ElementRef) { }
+
+    ngOnInit() {
+        this.element.nativeElement.focus();
+    }
```
DmitryAstafyev, line 13: Event focus can be applied to such kind of HTMLElements like Input, TextArea, Select, CheckBox and other forms controls. But it also possible to "focus" any other node (like `div`, or `span`). To do it right node should have attribute `tabindex` (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex)

Like
```
<div tabindex="0">Tabbable due to tabindex.</div>
```

DmitryAstafyev: One more thing. Please check: as far as I remember `OnInit` called after all data is bound, but it doesn't give a guaranty view is ready. Probably would be safer to use `AfterViewInit` here? Because this hook would be called only after the whole view with children is checked and ready

DmitryAstafyev (changes requested):

DmitryAstafyev:
This pull request would be rejected. Functionality related to refactoring of `search.manager` the component would be merge due to another pull request #903

PR #933, opened 2020-10-13
Support global ENTER/ESC for dialogs

fur-gan:


application/client.core/src/app/environment/services/service.file.opener.ts
```diff
@@ -99,13 +98,18 @@ export class FileOpenerService implements IService, IFileOpenerService {
...
                             caption: `Opening files`,
                             component: {
                                 factory: DialogsMultipleFilesActionComponent,
                                 inputs: {
                                     fileList: fileList,
+                                    onEnter: PopupsService.getObservable().onEnter,
+                                    merge: () => {
+                                        this.merge(fileList),
+                                        PopupsService.remove(guid);
+                                    },
```
DmitryAstafyev, line 112: Please rename input `merge` into `defaultAction` or something like it. The name `merge` confuse a little, looks like dialog would be opened to call merge-function

application/client.core/src/app/environment/components/dialogs/multiplefiles/component.ts
```diff
@@ -15,6 +15,8 @@ import ContextMenuService, { IMenuItem } from '../../../services/standalone/serv
 export class DialogsMultipleFilesActionComponent implements AfterViewInit, OnDestroy {
 
     @Input() fileList: FilesList;
+    @Input() onEnter: Observable<void>;
+    @Input() merge: () => void;
```
DmitryAstafyev, line 19: See comment bellow. Better use name of input like `defailtAction` or simular

DmitryAstafyev (changes requested):

PR #941, opened 2020-10-29
Rework DLT column formatting options

fur-gan:


application/client.core/src/app/environment/components/views/row/columns/headers/component.ts
```diff
@@ -74,6 +74,22 @@ export class ViewOutputRowColumnsHeadersComponent implements AfterViewInit, OnDe
...
+        if (event.target !== undefined) {
+            this._ng_mouseOverHeader = (event.target as HTMLElement).innerText;
+        }
+        this._forceUpdate();
+    }
+
+    public _ng_onMouseOut(event: MouseEvent) {
+        this._ng_mouseOverHeader = '';
+        this._forceUpdate();
+    }
```
DmitryAstafyev, line 91: Why we need it? Would be simpler to bind header in `*ngFor` loop?

application/client.core/src/app/environment/layout/contextmenu/component.ts
```diff
@@ -34,13 +34,15 @@ export class LayoutContextMenuComponent implements OnDestroy, AfterViewInit {
     }
 
     public ngOnDestroy() {
+        ContextMenuService.close = undefined;
```
DmitryAstafyev, line 37: Not a clean solution. I mean it isn't clear how it works and why it's done in this way. See comments bellow

application/client.core/src/app/environment/services/standalone/service.contextmenu.ts
```diff
@@ -17,14 +17,30 @@ export interface IMenu {
...
         onShow: new Subject<IMenu>(),
     };
 
+    public get close(): TClose {
+        return this._close;
+    }
+
+    public set close(func: TClose) {
+        this._close = func;
+    }
```
DmitryAstafyev, line 43: It isn't a mistake or similar. I mean solution works. But... This is not clean solution and here are a few reasons:
- why close-handler defined outside?
- "who" can define close-handler? Anybody?
- why it's public? (related to previous)

As for me, it would be simpler and in logic of existing code to use `subject`. Why not? Just add new subject/event (in addition to "onShow"), like "onClose" and what's all. Any consumers could subscribe on "onClose" event and do, what they want to do without any risk.

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

PR #951, opened 2020-11-06
Show release notes of installed Chipmunk version

fur-gan:


application/client.core/src/app/environment/services/service.release.notes.ts
```diff
@@ -1,6 +1,7 @@
 import { Observable, Subject, Subscription } from 'rxjs';
 import { IService } from '../interfaces/interface.service';
 import { IPCMessages } from './service.electron.ipc';
+import { IVersion } from '../../../../../common/ipc/electron.ipc.messages';
```
DmitryAstafyev, line 4: Please, include export of interface `IVersion` into `application/common/ipc/electron.ipc.messages/index.ts` and get access to it from `IPCMessages.IVersion`... The reason is simple - we should use outside of namespace reference (lile `application/common/ipc/electron.ipc.messages/index.ts`) as less as possible

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

PR #981, opened 2020-12-03
Remember uninstalled default plugins

fur-gan:


application/client.core/src/app/environment/components/tabs/plugins/details/component.ts
```diff
@@ -77,6 +77,9 @@ export class ViewPluginsDetailsComponent implements AfterContentInit, AfterViewI
     }
 
     public _ng_onInstallPlugin() {
+        if (this._ng_plugin.default) {
+            PluginsService.getManager().reinstallDefaultPlugin({ name: this._ng_plugin.name});
+        }
```
DmitryAstafyev, line 82: I am not sure, the client's logic should depend on is plugin default or not. For client it's just a plugin. 
I believe this is the responsibility of backend (electron) to recognize the plugin and do some specific actions.

DmitryAstafyev (changes requested):

application/electron/src/classes/class.statefile.ts
```diff
@@ -97,50 +106,56 @@ export class StateFile<TState> implements IService {
...
-                    this._logger.error(`Wrong format of state file "${this._file}": ${valid.message}.\n File will be reset to defualts values.`);
-                    return this._default(true).then(() => {
-                        resolve(this._defaults);
-                    }).catch(reject);
-                }
-                resolve(state);
+                resolve(this._validate(state));
             }).catch((error: Error) => {
-                this._logger.error(`Fail to read state at "${this._file}" due error: ${error.message}`);
+                reject(this._logger.error(`Fail to read state at "${this._file}" due error: ${error.message}`));
```
DmitryAstafyev, line 120: Ah this is also my usual error ) Weird why TS doesn't check it :/
`reject` always should return `Error` object. Because on catch we are always expecting `Error`, but not a `string`
Well, it would be better
```
reject(new Error(this._logger.error(`Fail to read state at "${this._file}" due error: ${error.message}`)));
```

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

PR #1066, opened 2021-02-15
Fix issue with stuck vertical scroll

fur-gan:


DmitryAstafyev:
It makes sense to add into this pull request:
- change for `application/client.libs/chipmunk.client.components/projects/chipmunk-client-material/package.json` : version up to `1.2.11`
- change dependency in `application/client.core/package.json`

DmitryAstafyev:
Sure also `application/client.core/package-lock.json` should be updated as soon as there is an updated dependency

PR #1077, opened 2021-02-19
Update filter name upon adding a new filter

fur-gan:


application/client.core/src/app/environment/components/sidebar/search.manager/providers/provider.ts
```diff
@@ -67,12 +67,14 @@ export abstract class Provider<T> {
...
         context: Subject<IContextMenuEvent>,
         doubleclick: Subject<IDoubleclickEvent>,
+        update: Subject<void>,
     } = {
         change: new Subject(),
         selection: new Subject(),
         edit: new Subject(),
         context: new Subject(),
         doubleclick: new Subject(),
+        update: new Subject(),
```
DmitryAstafyev, line 77: It isn't clear, why we need the addition event `update` as soon as we already have the event `change`?

DmitryAstafyev (commented):
In general: we already have event `change`. You are suggesting to add one more event `update`. From code isn't really clear why we need two events more or less (based on naming) about same. Is it possible to resolve the issue using the existing event `change`? 
Or maybe I didn't catch an idea?

Issue #703, opened 2020-02-28
Add IPv6 support to DLT live mode


MortronMeymo:
# Expected Behavior
The DLT live mode should be able to connect to a IPv6 only host.

# Current Behavior
Currently only IPv4 is supported.

PR #1079, opened 2021-02-19
[#703](feat) DLT connector: support for IPv6 (UI)

DmitryAstafyev:
implements #703


marcmo:
A couple of small changes needed that I will commit with a second commit

PR #1087, opened 2021-02-26
Create embedded terminal named `shell`

fur-gan:


application/client.core/src/app/environment/components/sidebar/shell/information/component.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    public ngOnDestroy() {
+        this._destroyed = true;
+    }
+
+    private _init() {
+        ElectronIpcService.request(new IPCMessages.ShellEnvRequest({ session: ShellService.session.getGuid() }), IPCMessages.ShellEnvResponse).then((response: IPCMessages.ShellEnvResponse) => {
+            if (response.error !== undefined) {
+                this._logger.error(`Failed to reqeust environment information due to Error: ${response.error}`);
+            } else {
```
DmitryAstafyev, line 47: Can we also add some feedback to user? I mean as soon as something is going wrong would be great to show some error message right in view? Because with the current solution there would be just nothing.

application/client.core/src/app/environment/components/sidebar/shell/information/component.ts
```diff
@@ -0,0 +1,68 @@
...
+    };
+
+    private _logger: Toolkit.Logger = new Toolkit.Logger('SidebarAppShellInformation');
+    private _destroyed = false;
+
+    constructor(private _cdRef: ChangeDetectorRef) {}
+
+    public ngOnInit() {
+        this._init();
+    }
```
DmitryAstafyev, line 37: Because it's async operation, even it would take less than 1 second, it's logical to have here a spinner (while IPC request is processing) as feedback to the user.

application/client.core/src/app/environment/components/sidebar/shell/information/template.html
```diff
@@ -0,0 +1,9 @@
+<mat-expansion-panel [expanded]="false">
+    <mat-expansion-panel-header>
+        <mat-panel-title>Information</mat-panel-title>
+    </mat-expansion-panel-header>
+    <div class="info">
+        <p>pwd: {{_ng_information.pwd}}</p>
+        <p>Shell: {{_ng_information.shell}}</p>
+    </div>
+</mat-expansion-panel>
```
DmitryAstafyev, line 9: See a little above:
- spinner as a suggestion to reflect the progress of the IPC message
- error message in case of some unexpected error

application/client.core/src/app/environment/components/sidebar/shell/input/component.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    private _addCommand(command: string) {
+        ElectronIpcService.request(new IPCMessages.ShellRecentCommandAddRequest({ command: command }), IPCMessages.ShellRecentCommandAddResponse).then((response: IPCMessages.ShellRecentCommandAddResponse) => {
+            if (response.error !== undefined) {
+                this._logger.error(`Failed to add command to recent commands due Error: ${response.error}`);
+            }
+        }).catch((error: Error) => {
+            this._logger.error(`Failed to add command to recent commands due Error: ${error.message}`);
+        });
+    }
```
DmitryAstafyev, line 161: I think it isn't necessary to send a command from FE to add something into recent. Because on BE we know - was the command successful or not, we can save the command into recent without a request from FE. I mean as soon as the command was successful, we just add it into recent on BE and do not until FE will ask to do it.

application/client.core/src/app/environment/components/sidebar/shell/running/component.ts
```diff
@@ -0,0 +1,103 @@
+import { NotificationsService } from '../../../../services.injectable/injectable.service.notifications';
+import { IShellProcess} from '../../../../../../../../common/ipc/electron.ipc.messages';
+import { Component, OnDestroy, ChangeDetectorRef, OnInit } from '@angular/core';
+import { Subscription } from 'rxjs';
+
+import ElectronIpcService, { IPCMessages } from '../../../../services/service.electron.ipc';
+import ContextMenuService, { IMenuItem } from '../../../../services/standalone/service.contextmenu';
+import ShellService from '../services/service';
```
DmitryAstafyev, line 8: It's a great idea/way to organize communication and create an own service for shell. I have two suggestions, just to complete this idea:
- as soon as this service is related only to shell, let's put it into a component folder (see `charts` as an example)
- service can care about all IPC calls and inside components you can call it as simple promises. The benefit is: you will have all IPC calls in one place. It would easy to debug and modify. 
-

DmitryAstafyev (changes requested):

PR #1098, opened 2021-03-02
Add terminated panel

fur-gan:
- Add panel for finished/terminated processes
- Swap places of input panel and running panel
- Show runtime of terminated process on hovering

### DIFF
--- a/application/client.core/src/app/environment/components/sidebar/shell/terminated/template.html
+++ b/application/client.core/src/app/environment/components/sidebar/shell/terminated/template.html
@@ -1,0 +1,27 @@
+<mat-expansion-panel [expanded]="true">
+    <mat-expansion-panel-header>
+        <mat-panel-title>Terminated</mat-panel-title>
+        <mat-panel-description>{{_ng_count()}}</mat-panel-description>
+    </mat-expansion-panel-header>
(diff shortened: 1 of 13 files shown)


application/client.core/src/app/environment/components/sidebar/shell/terminated/template.html
```diff
@@ -0,0 +1,27 @@
+<mat-expansion-panel [expanded]="true">
+    <mat-expansion-panel-header>
+        <mat-panel-title>Terminated</mat-panel-title>
+        <mat-panel-description>{{_ng_count()}}</mat-panel-description>
+    </mat-expansion-panel-header>
+    <div *ngIf="_ng_terminated.length === 0">
+        <p>No processes are terminated</p>
+    </div>
+    <div *ngIf="_ng_terminated.length > 0">
+        <div appTooltip [appTooltipContent]="template" [appTooltipRefreshRate]="500" class="terminated" *ngFor="let terminated of _ng_terminated">
```
DmitryAstafyev, line 10: property `appTooltipRefreshRate` could be removed from here to avoid of refresh of state as soon as duration and received data would not change any more

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

PR #1108, opened 2021-03-08
Remove redundant  shortcuts and fix typos

fur-gan:


DmitryAstafyev:
I've added a dummy item with "stupid" long caption, just to see, how it looks like. Here is result.

Screenshot from 2021-03-09 10-14-25
Screenshot from 2021-03-09 10-14-36

* screenshort 1: space between caption and shortcut label looks too much
* screenshort 2: weird width of menu
* common: does it make sense to render `<span>` or shurtcut in menu layout if it wasn't defined? I mean maybe `*ngIf`

Issue #1080, opened 2021-02-22
macOS UX improvements


marcmo:
- turn on https://www.electronjs.org/docs/tutorial/dark-mode einschalten damit die macOS Title-Bar auch so dunkel ist wie der Rest (auch wenn macOS auf Light gestellt ist)
- use https://www.electronjs.org/docs/tutorial/recent-documents for native recent menu
- use https://www.electronjs.org/docs/tutorial/represented-file Natives Proxy Icon in the titlebar (together with filename), always matching current tab
- setDocumentEdited=true for active/not currently saved live traces (Cmd/Serial/...)
- menues on mac os: e.g. we need a  File > Open... including Shortcut. and preferences under `chipmunk`, not under `file`  (also "Plugins" and "Install cm tool" under "chipmunk")
appname in dock capitalized  -> Chipmunk.app

DmitryAstafyev:
Hello, @marcmo 
Could you please give some more comment/information, about this:
```
setDocumentEdited=true for active/not currently saved live traces (Cmd/Serial/...)
```
It isn't clear, what do you want to see as result

DmitryAstafyev:
I will join to this issue in scope of points, which can be done using macos only

bzld:
Small but very helpful additional fix:

- Configure Chipmunk app metadata to accept text files (or all files?), so that you can open files by dragging them onto the Chipmunk dock icon, or by selecting it in the "Open With" Finder context menu.
 
(Quick googling sounds like this is a .plist thing: https://www.electronjs.org/docs/api/app#event-open-url-macos)

DmitryAstafyev:
> Small but very helpful additional fix:
> 
> * Configure Chipmunk app metadata to accept text files (or all files?), so that you can open files by dragging them onto the Chipmunk dock icon, or by selecting it in the "Open With" Finder context menu.
> 
> (Quick googling sounds like this is a .plist thing: https://www.electronjs.org/docs/api/app#event-open-url-macos)

Would you mind if I will exclude it into the standalone issue (feature request)?

DmitryAstafyev:
I would close this issue as soon as in general it was resolved. As for "Open With", we are not sure, it's a good idea to assign chipmunk with any type of files as soon as chipmunk portable solution. But we always can create a standalone issue for that to have a standalone discussion

PR #1118, opened 2021-03-10
UI improvements

fur-gan:
Various ui improvements such as:
- Chipmunk now in dark-mode theme
- Keyboard shortcuts next to toolbar options
- New toolbar Option "Chipmunk"

### DIFF
--- a/application/electron/src/controllers/electron/controller.electron.menu.ts
+++ b/application/electron/src/controllers/electron/controller.electron.menu.ts
@@ -22,6 +22,10 @@ const MAX_NUMBER_OF_RECENT_FILES = 20;
 
 const MENU_TEMPLATE = [
     {
+        label: 'Chipmunk',
+        submenu: [ ],
+    },
+    {
         label: 'File',
         submenu: [ ],
     },
@@ -105,11 +109,11 @@ export default class ControllerElectronMenu {
     private _create() {
         const template: any[] = this._getTemplate();
         // Add files submenu
-        template[0].submenu.push(...this._getFilesLocalSubmenu());
+        template[1].submenu.push(...this._getFilesLocalSubmenu());
         // Add recent files (if it exists)
         if (ServiceStorage.get().get().recentFiles.length > 0) {
-            template[0].submenu.push({ type: 'separator' });
-            template[0].submenu.push({
+            template[1].submenu.push({ type: 'separator' });
+            template[1].submenu.push({
                 label: 'Open Recent',
                 submenu: [
                     ...this._getRecentFiles(),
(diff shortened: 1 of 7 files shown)


DmitryAstafyev:
Something weird

Screenshot 2021-03-18 at 09 49 14
Screenshot 2021-03-18 at 09 49 20

DmitryAstafyev:
An application menu item (like menu item "Chipmunk") is using in macOS, but as for Linux and windows - doesn't not. It looks weird:

Screenshot from 2021-03-18 10-14-42

For example how VSCode looks like on linux

Screenshot from 2021-03-18 10-15-00

No application menu item - "Code". Menu starts from "File". But it has an analog of macOS app's menu item. It looks like it
 
Screenshot from 2021-03-18 10-15-04

Maybe we can use this.

DmitryAstafyev:
I would close this pull request as soon as issue #1080 requires more specifications/details. We should describe our expectation for each platform as soon as it's different for mac/win/linux

application/electron/src/controllers/electron/controller.electron.menu.ts
```diff
@@ -21,6 +21,10 @@ import * as os from 'os';
 const MAX_NUMBER_OF_RECENT_FILES = 20;
 
 const MENU_TEMPLATE = [
+    {
+        label: 'Chipmunk',
+        submenu: [ ],
+    },
```
bzld, line 28: This seems wrong. The existing code below already adds this if platform==darwin.

application/electron/src/controllers/electron/controller.electron.menu.ts
```diff
@@ -126,15 +130,14 @@ export default class ControllerElectronMenu {
...
+                template[1].submenu.push({
                     label: action.caption,
                     click: HandlerItemExportActionCall.bind(null, action.id),
                 });
             });
         }
-        template[0].submenu.push({ type: 'separator' });
         template[0].submenu.push({
             label: 'Plugins',
             click: HandlerItemPlugins,
```
bzld, line 145: Move this inside the `if (darwin) {} else {}` block right below, and put it to main menu on darwin (right below Settings), and File menu on other platforms.

DmitryAstafyev: Hi Florian, 
We've closed this pull-request without merge as soon as original issue #1080 isn't clean. This issue (#1080) should be described in more detail considering each platform. We should have more or less the same UI and user expectations for all platforms, but issue was creater more to macos direction.

Partly it's already implemented. For example, recent documents are already available via the dock context menu (mac) and taskbar (windows).

PR #1120, opened 2021-03-15
Add support of custom envvars

fur-gan:


DmitryAstafyev:
If we have some table view, which would be scrollable in the vertical direction, we should take care of the stuck header. I mean header always should be on the top.

Also, I think it better to use "Value" instead "Path".

Screenshot 2021-03-17 at 18 48 45

DmitryAstafyev:
The idea of adding a new variable into the list directly - very nice. Here are a couple of suggestions:
- add to the bottom, but not to the top
- if a user isn't at the top of the list - scroll it to the bottom with click on "Add"
- and "Create" better rename to "Add"

DmitryAstafyev:
I think we don't need horizontal scrolling. The left column can have a fixed width and the right column can always fit to the width of the dialog.
Screenshot 2021-03-17 at 18 49 49

DmitryAstafyev:
Without white border it looks better. Isn't?

Screenshot 2021-03-17 at 18 53 52

DmitryAstafyev:
Broken views:

Screenshot 2021-03-17 at 18 50 46

Screenshot 2021-03-17 at 19 03 44

DmitryAstafyev:
About icons

Screenshot 2021-03-17 at 19 05 39

marcmo:
looks all good!

PR #1155, opened 2021-04-14
Shell improvements, adjustments and fixes

fur-gan:


DmitryAstafyev:
Recent commands looks broken
Peek 2021-04-22 11-44

DmitryAstafyev:
Let's make "delete" icon less "aggressive" (smaller, grey and white on hover). Also if there is no space for text, let's use `...`. For example:
Peek 2021-04-22 12-24

application/client.core/src/app/environment/components/sidebar/shell/input/component.ts
```diff
@@ -7,35 +7,13 @@ import { FormControl } from '@angular/forms';
...
-
-export interface INewInformation {
-    shell: string;
-    pwd: string;
-    env: IEnvironment[];
-}
-
 @Component({
     selector: 'app-sidebar-app-shell-input',
     templateUrl: './template.html',
```
DmitryAstafyev, line 19: As soon as you are trying to get access (in styles) to outside node (`mat-autocomplete` would be created outside) you should switch your component to `encapsulation: ViewEncapsulation.None`. Also you need additional class name to recognize your options
```
<mat-option *ngFor="let recent of _ng_recent | async"
                        [value]="recent"
                        [attr.title]="recent"
                        class="recent-requests small t-console t-small app-sidebar-app-shell-input"  <---- here "app-sidebar-app-shell-input"
                        (mouseenter)="_ng_onMouseOver(recent.description)"
                        (mouseleave)="_ng_onMouseOver('')">
                <div class="options">
                    <span class="filename" [innerHTML]="_ng_getSafeHTML(recent.tdescription)"></span>
                    <mat-icon (click)="_ng_onRemove(recent.description)">close</mat-icon>
                </div>
            </mat-option>
```
And in style.less

```
mat-option.app-sidebar-app-shell-input {
    & .options {
        display: flex;
        flex-direction: row;
        align-items: center;
...
```

application/client.core/src/app/environment/components/sidebar/shell/input/styles.less
```diff
@@ -4,42 +4,29 @@
...
+
+& .options {
+    display: flex;
+    flex-direction: row;
+    & span.filename {
+        flex: auto;
+        overflow: hidden;
+    }
+    & mat-icon {
+        margin-right: 0px!important;
```
DmitryAstafyev, line 30: needs addition class or id to recognize your styles. Like
```
mat-option.app-sidebar-app-shell-input {
    & .options {
        display: flex;
        flex-direction: row;
        align-items: center;
...
```

application/client.core/src/app/environment/components/sidebar/shell/input/template.html
```diff
@@ -4,40 +4,37 @@
...
                     <span class="filename" [innerHTML]="_ng_getSafeHTML(recent.tdescription)"></span>
-                </mat-option>
-            </mat-autocomplete>
-        </mat-form-field>
-    </div>
-    <div class="buttons">
-        <button mat-stroked-button class="button" color="primary" (click)="_ng_onClearRecent()">Clear</button>
-        <button mat-stroked-button class="button" color="primary" (click)="_ng_onEnvironment()">Environment</button>
-    </div>
+                    <mat-icon *ngIf="_ng_hoverRecent === recent.description" (click)="_ng_onRemove(recent.description)">close</mat-icon>
```
DmitryAstafyev, line 35: Add/remove node in runtime quite expensive. We can resolve this using just css. I suggest get rid of `*ngIf="_ng_hoverRecent === recent.description"` but on css level do something like
```
.parent {
   & .icon {
       display: none;
   }
   &:hover {
        & .icon {
            display: inline-block;
        }
   }
}

```

DmitryAstafyev (changes requested):

DmitryAstafyev:
Tooltip is broken also
Screenshot from 2021-04-22 12-13-13

application/client.core/src/app/environment/components/sidebar/shell/environment/component.ts
```diff
@@ -1,198 +1,125 @@
...
-    constructor() { }
+    private _sessionID: string;
+    private _subscriptions: { [key: string]: Toolkit.Subscription | Subscription } = {};
+    private _logger: Toolkit.Logger = new Toolkit.Logger('SidebarAppShellEnvironmentComponent');
 
-    public ngOnInit() {
-        this._sortInformation();
-        this._ng_height = Math.round(window.innerHeight * .7);
+    constructor() {
+        this._sessionID = TabsSessionsService.getActive().getGuid();
```
DmitryAstafyev, line 50: Please take into account
```
public getActive(): Session | undefined
```
`getActive()` can return `undefined`. In our lifecircle it isn't possbile situation, but as soon as some changes came, it can be a source of issues.

If we cannot be sure: does session exist or not, better to to move `this._sessionID = TabsSessionsService.getActive().getGuid();` into `ngInit` or `ngViewInit` and do related actions, like show message "session isn't available or something like it

application/client.core/src/app/environment/components/sidebar/shell/environment/component.ts
```diff
@@ -1,198 +1,125 @@
...
+    public _ng_onSetPwd() {
+        this.service.setPwd({ session: this._sessionID }).then((value: string) => {
+            if (value.trim() !== '') {
+                this.service.selectedPreset.information.pwd = value;
+                this.service.setPreset(this._sessionID, { pwd: value }).catch((error: Error) => {
+                    this._logger.error(error);
+                });
+            }
+        }).catch((error: Error) => {
+            this._logger.error(error);
```
DmitryAstafyev, line 104: Shell we say something to user in case of error? Maybe notification?

application/client.core/src/app/environment/components/sidebar/shell/environment/component.ts
```diff
@@ -1,198 +1,125 @@
...
-            this._editing.value = this._prevValue;
-        } else if (this._editing.editing.variable) {
-            this._editing.variable = this._prevVariable;
-        }
+    public _ng_onSetPwd() {
+        this.service.setPwd({ session: this._sessionID }).then((value: string) => {
+            if (value.trim() !== '') {
+                this.service.selectedPreset.information.pwd = value;
+                this.service.setPreset(this._sessionID, { pwd: value }).catch((error: Error) => {
+                    this._logger.error(error);
```
DmitryAstafyev, line 100: Maybe notification to user also?

application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    public ngOnDestroy() {
+        if (this._ng_selected !== undefined) {
+            this._ng_selected.editing = {
+                value: false,
+                variable: false
+            };
+            this._ng_selected = undefined;
+        }
```
DmitryAstafyev, line 46: would be `this._ng_selected = undefined` enough? without change of properties and checks. Like
```
public ngOnDestroy() {
   this._ng_selected = undefined;
   this._ng_adding = false;
   this.setEnvironment(this.information);
}
```

DmitryAstafyev: and do you actually need to drop it as soon as the component would be destroyed in any way? :/

application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts
```diff
@@ -0,0 +1,198 @@
...
+    public _ng_add() {
+        this._ng_selected = undefined;
+        this._ng_adding = true;
+        if (this._ng_variableAdd !== undefined) {
+            this._ng_variableAdd.nativeElement.scrollIntoView();
+        }
+    }
+
+    public _ng_remove() {
+        if (this._ng_selected !== undefined) {
```
DmitryAstafyev, line 123: btw `_ng_selected` cannot be `undefined` according to component's declarations
```
public _ng_selected: IEnvironment;
```
probably checks are stay from previous edition of component

application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts
```diff
@@ -0,0 +1,198 @@
...
+        this._finishCreating(false);
+        this._editing = newEditEnv;
+    }
+
+    private _dropEditChanges() {
+        if (this._editing.editing.value) {
+            this._editing.value = this._prevValue;
+        } else if (this._editing.editing.variable) {
+            this._editing.variable = this._prevVariable;
+        }
```
DmitryAstafyev, line 150: Declaration 
```
private _editing: IEnvironment | undefined;
```
`_editing` can be `undefined`
In this case here we have potential exception `Cannot read value of undefined`
```
       if (this._editing.editing.value) {                      <-- here
            this._editing.value = this._prevValue;
        } else if (this._editing.editing.variable) {       <-- here
            this._editing.variable = this._prevVariable;
        }
```

application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts
```diff
@@ -0,0 +1,198 @@
...
+                selected: false,
+            });
+        }
+        this._ng_adding = false;
+        this._ng_newValue = '';
+        this._ng_newVariable = '';
+    }
+
+    private _setCustomStatus() {
+        if (this._ng_selected !== undefined && this._ng_selected.editing !== undefined) {
```
DmitryAstafyev, line 175: `_ng_selected` cannot be `undefined` or declaration is wrong

DmitryAstafyev (changes requested):

application/client.core/src/app/environment/components/sidebar/shell/environment/component.ts
```diff
@@ -1,198 +1,125 @@
...
-                editing: {
-                    value: false,
-                    variable: false,
-                },
-                value: this._ng_newValue,
-                variable: this._ng_newVariable,
-                selected: false,
+    public _ng_onShellChange(event: MatSelectChange) {
+        this.service.setEnv({ session: this._sessionID, shell: event.value }).then(() => {
+            this.service.selectedPreset.information.shell = event.value;
```
DmitryAstafyev, line 110: Is it okay to have direct access to properties of service? It looks quite dangerous
```
this.service.selectedPreset.information.shell = event.value;
```
Maybe it makes sense to have something like
```
this.service.setShell(event.value);
```

application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts
```diff
@@ -0,0 +1,198 @@
...
+        }
+        this._finishCreating(true);
+        this._setCustomStatus();
+        this._sortInformation();
+    }
+
+    public _ng_onClick(env: IEnvironment) {
+        if (this._ng_selected !== env) {
+            this._changeFocus();
+        }
```
DmitryAstafyev, line 95: Compering of objects isn't a trustable thing in JS. Maybe you want to compare by some `id/uuid`? Like
```
if (this._ng_selected.uuid !== env.uuid) { ... }
```
As soon as name of variable is unique in scope of envvars, property `variable` can be used as unique id:
```
if (this._ng_selected.variable !== env.variable) { ... }
```

application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts
```diff
@@ -0,0 +1,198 @@
...
+            env.editing.value = true;
+            this._prevValue = env.value;
+        } else {
+            env.editing.variable = true;
+            this._prevVariable = env.variable;
+        }
+    }
+
+    public _ng_add() {
+        this._ng_selected = undefined;
```
DmitryAstafyev, line 115: in same time in declaration `_ng_selected` cannot be `undefined`

application/client.core/src/app/environment/components/sidebar/shell/environment/preset/component.ts
```diff
@@ -124,7 +124,7 @@ export class SidebarAppShellPresetComponent implements OnInit, OnDestroy {
     }
 
     public _ng_onReset() {
-        this.service.presets[1].information = copy(this.service.defaultInformation);
+        this.service.selectedPreset.information = copy(this.service.defaultInformation);
```
DmitryAstafyev, line 127: Direct access and modification of service's properties isn't good idea in general

application/client.core/src/app/environment/components/sidebar/shell/running/component.ts
```diff
@@ -72,21 +73,13 @@ export class SidebarAppShellRunningComponent implements OnDestroy, OnInit {
...
-            items: items,
-            x: event.pageX,
-            y: event.pageY,
+    public _ng_onMouseOver(guid: string) {
+        this._ng_hoverGuid = guid;
+    }
+
+    public _ng_onTerminate(process: IPCMessages.IShellProcess) {
+        this.service.terminate({ session: this._sessionID, guid: process.guid }, process.command ).catch((error: string) => {
+            this._logger.error(error);
```
DmitryAstafyev, line 82: It's important operation and user feedback also is important here. Let's add user notification in case of error

application/client.core/src/app/environment/components/sidebar/shell/terminated/component.ts
```diff
@@ -74,6 +76,35 @@ export class SidebarAppShellTerminatedComponent implements OnDestroy, OnInit {
...
+    public _ng_onMouseOver(guid: string) {
+        this._ng_hoverGuid = guid;
+    }
+
+    public _ng_onReplay(process: IPCMessages.IShellProcess) {
+        this.service.runCommand({
+            session: this._sessionID,
+            command: process.command,
+        }).catch((error: Error) => {
+            this._logger.error(error);
```
DmitryAstafyev, line 88: user feedback

DmitryAstafyev (changes requested):

application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts
fur-gan: The marked lines 146-150 are part of the method `_dropEditChanges()`, which is only called inside `_changeFocus(...)`.
The method `_changeFocus(...)` checks in it's first line `if (this._editing !== undefined)` and only calls `_dropEditChanges()` after making sure that it's not `undefined`.

```    private _changeFocus(newEditEnv?: IEnvironment) {
        if (this._editing !== undefined) {            // Making sure it's not undefined
            this._dropEditChanges();                    // Calling the method below
            this._editing.editing.value = false;
            this._editing.editing.variable = false;
        }
        this._finishCreating(false);
        this._editing = newEditEnv;
    }

    private _dropEditChanges() {                              // Since we checked if this._editing is undefined, we don't need to check here anymore
        if (this._editing.editing.value) {
            this._editing.value = this._prevValue;
        } else if (this._editing.editing.variable) {
            this._editing.variable = this._prevVariable;
        }
    }```

File: application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts (hunk 7 above)
Comment (fur-gan) [reply application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts:46]: The idea behind assigning the editing variables to `false` is to prevent any open inputs for editing/adding.
Since `_ng_selected` is directly connected to the environment, we actually modify the environment. The reason to use `_ng_selected` already has the Object which has the  input open, we save us the looping around all environments to find the one which is still being edited.
But yeah, I guess it makes sense to leave out `this._ng_selected = undefined;` since it's being deleted the moment the component is destroyed

File: application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts (hunk 9 above)
Comment (DmitryAstafyev) [reply application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts:150]: If method "this._dropEditChanges()" called only from "_changeFocus()" would it be better move checks and rest stuff there?
```
// body of _dropEditChanges
        if (this._editing !== undefined) {           
            if (this._editing.editing.value) {                      
                this._editing.value = this._prevValue;
            } else if (this._editing.editing.variable) {       
                this._editing.variable = this._prevVariable;
            }                 
            this._editing.editing.value = false;
            this._editing.editing.variable = false;
        }

// In this case in _changeFocus, you just call it
        this._dropEditChanges();
        this._finishCreating(false);
        this._editing = newEditEnv;

```

Comment (fur-gan) [reply application/client.core/src/app/environment/components/sidebar/shell/environment/environment_variables/component.ts:150]: yeah, I've changed it a bit more than that, so I think the problem should also be resolved now
Needed some changes for the other comments as well, that's why

File: application/client.core/src/app/environment/components/sidebar/shell/running/component.ts [hunk 17]
```diff
@@ -72,21 +78,13 @@ export class SidebarAppShellRunningComponent implements OnDestroy, OnInit {
...
-            items: items,
-            x: event.pageX,
-            y: event.pageY,
+    public _ng_onTerminate(process: IPCMessages.IShellProcess) {
+        this.service.terminate({ session: this._sessionID, guid: process.guid }, process.command ).catch((error: Error) => {
+            this._showNotification({
+                caption: 'Failed to terminate process',
+                message: error.message,
+            });
+            this._logger.error(error);
```
DmitryAstafyev, line 87: I would ask you do not pass into logs whole error object. Instead this
```
this._logger.error(error);
```
use please this
```
this._logger.error(error.message);
```
Because `errors` is an object, logger will try to convert it into string and we will get huge output with all properties.

Also, take in account, you can refactor a little your code to this:
```
this._showNotification({
    caption: 'Failed to terminate process',
    message: this._logger.error(error.message),
});
```
Well `logger` always returns same string as you passed into. It allows you make code more compact.

application/client.core/src/app/environment/components/sidebar/shell/terminated/component.ts
```diff
@@ -74,6 +82,35 @@ export class SidebarAppShellTerminatedComponent implements OnDestroy, OnInit {
...
+    public _ng_onReplay(process: IPCMessages.IShellProcess) {
+        this.service.runCommand({
+            session: this._sessionID,
+            command: process.command,
+        }).catch((error: Error) => {
+            this._showNotification({
+                caption: 'Failed to replay command',
+                message: error.message,
+            });
+            this._logger.error(error);
```
DmitryAstafyev, line 94: Do not pass please into logger full error object. Just message is okay.

DmitryAstafyev (changes requested):

application/client.core/src/app/environment/components/sidebar/shell/environment/preset/component.ts
DmitryAstafyev: You still have direct access to `service` properties. Basically, it isn't an issue of code, but an issue of logic. 

Let's say you have a restaurant. And you have a menu. You don't mind changing a menu based on the wishes of your customers. Direct access to service properties, means - each customer can open your menu, remove one page from there and put own one. Another way - create box, which each your customer can use to put his wishes there. And you, based on these wishes, will change a menu.

I am talking about some method, like

```
this.service.setSelectedPresetInfo(info)
```

In your case if we are talking about `default` data. Would not be better to use something like:
```
this.service.resetToDefault();
```
instead
```
this.service.selectedPreset.information = copy(this.service.defaultInformation)
```
?

application/client.core/src/app/environment/components/sidebar/shell/input/component.ts
DmitryAstafyev: please use some *class name* for parent of `& .options`. As soon as you declare it as global we should take care about possible conflicts by classnames

DmitryAstafyev (approved):

marcmo (changes requested):
please squash the commits before we can merge them

PR #1167, opened 2021-05-04
Search support

DmitryAstafyev:


application/apps/indexer/processor/src/search.rs
```diff
@@ -149,29 +157,34 @@ impl SearchHolder {
                 .join("|")
         );
         let matcher = RegexMatcher::new(&regex)?;
+        let mut matchers: Vec<Regex> = vec![];
+        for filter in self.search_filters.iter() {
+            if let Ok(r) = Regex::from_str(&filter.value.clone()) {
```
marcmo, line 162: missing error handling. maybe we can even to this when creating the search holder and making it fail when a regex cannot be built.

application/apps/indexer/processor/src/search.rs
```diff
@@ -134,8 +134,16 @@ impl SearchHolder {
 
     /// execute a search for the given input path and filters
     /// return the file that contains the search results along with the
-    /// number of found matches
-    pub fn execute_search(&self) -> Result<(PathBuf, u64), SearchError> {
+    /// map of found matches. Format of map is an array of matches:
+    /// [
+    ///     [position in stream, ...[number of filter] ],
```
marcmo, line 139: instead of `number of filter` better use `index of matching filter`

PR #1178, opened 2021-05-11
Chart improvement and fixes

fur-gan:


DmitryAstafyev:
Width of y-axis should be considered on a calculation of redirection (on click event)

Way to reproduce:
1. Open some file with charts (better have at least 100K rows to see redirection diff)
2. Click on some point on chart (with label) - as soon as you click on point, point has meta-data about exact row number and you will be redirected to correct row in view
3. Click NOT on point (see (2)), but a little bit up. In this case we don't have any metadata and we should calculate row number and it should be approximately same like in point (2)

Peek 2021-05-17 10-09

application/client.core/src/app/environment/components/views/chart/chart/component.ts
```diff
@@ -135,10 +135,17 @@ export class ViewChartCanvasComponent implements AfterViewInit, AfterContentInit
...
+            if (this._ng_charts === undefined) {
+                return;
+            }
+            const visible: boolean = this._isYAxisVisible();
+            const range: IRange = this._getRange();
+            const rows: number = range.end - range.begin;
             const size: ClientRect = (event.target as HTMLElement).getBoundingClientRect();
-            const rate: number = size.width / rows;
-            position = Math.round(event.offsetX / rate);
+            const width: number = size.width - (visible ? this._ng_charts.chartArea.left : 0) ;
```
DmitryAstafyev, line 145: I see, it's from current implementation, but could we move `getBoundingClientRect` out from here as soon as it's quite expensive operation? We also do it with `_resize` method and it would be more optimal if we just store value of `width` (which was gotten with `_resize`) as private property of class and use it here instead addition call of expensive method `getBoundingClientRect`

application/client.core/src/app/environment/components/views/chart/service.data.ts
```diff
@@ -487,7 +487,8 @@ export class ServiceData {
     }
 
     private _getValidNumberValue(val: string): number | undefined {
-        const value: number = parseInt(val, 10);
+        const regex = /^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$/;
+        const value: number = regex.test(val) ? parseFloat(val) : parseInt(val, 10);
```
DmitryAstafyev, line 491: I am not sure we need regexp here:
1. it's not okay to construct it each time, if it's const you can just create in top of module
```
const FLOAT_TEST = /..../gi;
```
and you will not construct it each time with calling of `_getValidNumberValue`

2. But I think you can get rid of regexp here.
```
const floatOrInt: number = parseFloat(str);
if (floatOrInt % 1 == 0) {
     // I am INTEGER
} else {
     // I am FLOAT
}
```
This solution should be tested for sure

DmitryAstafyev: Using of RegExp would be more expensive in any way. So, better avoid it

application/client.core/src/app/environment/components/views/chart/zoomer/component.ts
```diff
@@ -273,13 +273,27 @@ export class ViewChartZoomerCanvasComponent implements AfterViewInit, OnDestroy
...
+    private _prepMin(min: number): number {
+        if (Number.isInteger(min)) {
+            return Math.floor(min);
+        }
+        return min;
+    }
+
+    private _prepMax(max: number): number {
+        if (Number.isInteger(max)) {
+            return Math.floor(max + max * 0.02);
```
DmitryAstafyev, line 292: Are you sure we need `floor` on `max` value, but not `cail`?

DmitryAstafyev (changes requested):

application/client.core/src/app/environment/components/views/chart/chart/component.ts
```diff
@@ -403,15 +420,32 @@ export class ViewChartCanvasComponent implements AfterViewInit, AfterContentInit
...
+        if (Number.isInteger(min)) {
+            return Math.floor(min - min * 0.02);
+        }
+        return min - min * 0.02;
+    }
+
+    private _prepMax(max: number): number {
+        if (Number.isInteger(max)) {
+            return Math.floor(max + max * 0.02);
+        }
```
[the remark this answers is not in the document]
fur-gan: Yeah, seems like I missed that one. Only changed it on the other max in the zoomer

DmitryAstafyev (approved):

PR #1181, opened 2021-05-18
Replace spinner with message on chart when no chart is being created

fur-gan:


application/client.core/src/app/environment/components/views/chart/chart/template.html
```diff
@@ -1,6 +1,9 @@
 <div [attr.class]="_ng_progress() ? 'loading' : 'ready'">
     <mat-spinner [diameter]="32" mode="indeterminate"></mat-spinner>
 </div>
+<div [attr.class]="_ng_noData() ? 'loading' : 'ready'">
+    <p class="t-medium color-scheme-2">{{_ng_empty}}</p>
+</div>
```
DmitryAstafyev, line 6: Could we just put string label into template. It's constant and would not be changed I guess. Like
```
<p class="t-medium color-scheme-2">No data</p>
```
Also would it be possible to use same container and for spinner and for label?

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

PR #1183, opened 2021-05-24
Fix error message on loading file with chart

fur-gan:


DmitryAstafyev:
@fur-gan Hi!
I **cannot** confirm an issue. Tested on mac/linux/windows - all works.

application/client.core/src/tsconfig.app.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../tsconfig.json",
   "compilerOptions": {
     "outDir": "../out-tsc/app",
-    "types": [],
+    "types": ["node"],
+    "typeRoots": ["../node_modules/@types"],
```
DmitryAstafyev, line 6: I would not do this as soon as our client isn't node application

DmitryAstafyev (changes requested):

DmitryAstafyev:
Maybe this issue is related to your local configuration?

fur-gan:
> Maybe this issue is related to your local configuration?

That could be, I will check

fur-gan:
Yeah, seems like this was also an issue of my internal settings. It's related to the path not being recognized with '/' and works as soon as I replace the '/' with '\\'

PR #1192, opened 2021-06-02
Show scale and details of chart when switching sessions

fur-gan:


DmitryAstafyev:
I think service `SearchManagerService` isn't the right place to store the selected entity. The thing is: service `SearchManagerService` isn't bound to the session. Let say we have
> Session A
>> Filter A
>> Filter B
> Session B
>> Filter C
>> Filter D
Steps to get problems:
1) In the scope of session A select Filter A -> this filter will become selected in `SearchManagerService.selected = Filter A`
2) switch to session B
3) `SearchManagerService.selected = Filter A` still, even Session B doesn't have Filter A

If we would like to remember selected filters/charts/etc we should use a suitable controller from `application/client.core/src/app/environment/controller/session/dependencies/search`

DmitryAstafyev:
Please take a look to this pull request #1200 , it's also covered this issue.

PR #1205, opened 2021-06-15
Rebased grabber

DmitryAstafyev:


application/apps/rustcore/rs-bindings/src/js/session.rs
```diff
@@ -511,18 +540,23 @@ impl RustSession {
                 ));
             }
         };
-        self.op_channel
-            .0
-            .send(Operation::Assign {
+        let op_channel_tx = self.op_channel.0.clone();
+        async move {
+            match op_channel_tx.send(Operation::Assign {
```
marcmo, line 545: this does not look correct...what you are trying to do is to return a future. For that you should not artificially create an async block but rather use a combinator like `ready`
https://docs.rs/futures/0.3.15/futures/future/fn.ready.html

marcmo: Did you check what happens if you block in an async function? Is the js thread then also blocked? if so we need to wrap the blocking call in a `tokio::task::spawn_blocking`

application/apps/indexer/processor/src/search.rs
```diff
@@ -208,6 +250,64 @@ impl SearchHolder {
...
+                .join("|")
+        );
+        let mut values: Vec<ExtractedMatchValue> = vec![];
+        let mut regexs: Vec<Regex> = vec![];
+        for filter in self.filters.iter() {
+            match Regex::from_str(&filter_as_regex(filter)) {
+                Ok(reg) => regexs.push(reg),
+                Err(err) => return Err(SearchError::Regex(format!("{}", err))),
+            }
+        }
```
marcmo, line 292: much nicer: `Regex::from_string(...).map_err(|err| SearchError::Regex(...)?`

application/apps/indexer/processor/src/grabber.rs
```diff
@@ -12,9 +12,12 @@ pub trait GrabTrait {
     fn grab_content(&self, line_range: &LineRange) -> Result<GrabbedContent, GrabError>;
     fn inject_metadata(&mut self, metadata: GrabMetadata) -> Result<(), GrabError>;
     fn get_metadata(&self) -> Option<&GrabMetadata>;
+    fn drop_metadata(&mut self);
     fn associated_file(&self) -> PathBuf;
 }
 
+pub trait AsyncGrabTrait: GrabTrait + Sync + Send + std::fmt::Debug {}
```
marcmo, line 19: do we really need to add those traits with this commit? I thought you only need them for the cancellation stuff

application/apps/indexer/processor/src/search.rs
```diff
@@ -8,8 +8,11 @@ use std::{
     fs::File,
     io::{BufWriter, Write},
     path::{Path, PathBuf},
+    str::FromStr,
 };
+use regex::Regex;
 use thiserror::Error;
+use indexer_base::{progress::ComputationResult, utils};
```
marcmo, line 15: Pay attention to the rust linter warnings here and get rid of unused imports

application/apps/rustcore/rs-bindings/src/js/session.rs
```diff
@@ -622,27 +657,68 @@ impl RustSession {
...
+        };
+        let filters: Vec<SearchFilter> = filters.iter().map(|f| f.as_filter()).collect();
+        info!(
+            "Extract (operation: {}) will be done in {:?} withing next filters: {:?}",
+            operation_id, target_file, filters
+        );
+        let op_channel_tx = self.op_channel.0.clone();
+        async move {
+            match op_channel_tx.send(Operation::Extract {
+                target_file,
```
marcmo, line 706: same issue as above

application/apps/rustcore/rs-bindings/src/js/session.rs
DmitryAstafyev: you are absolutely right, that `async {}` block doesn't necessary here at all as soon as `#[node_bindgen]` macro cares about wrapping into the future.

application/apps/indexer/processor/src/grabber.rs
DmitryAstafyev: This trait related to async functionality, but not to cancelation.

PR #1208, opened 2021-06-17
Offer open multiple (drag&drop) files in tabs

fur-gan:


application/client.core/src/app/environment/services/service.file.opener.ts
```diff
@@ -162,6 +163,24 @@ export class FileOpenerService implements IService, IFileOpenerService {
...
+            return;
+        }
+        const openFile = (index: number) => {
+            if (index >= checked.length) {
+                return;
+            }
+            this.openFileByName(checked[index].path).then(() => {
+                openFile(++index);
+            }).catch((error: Error) => {
+                this._logger.error(`Fail to open file ${checked[index].path} due error: ${error.message}`);
```
DmitryAstafyev, line 178: I think we need to notify user about failing to open the file. And probably we still shoud try to open rest files.

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

PR #1210, opened 2021-06-22
Keep previous input on focus change in search

fur-gan:


DmitryAstafyev:
With this solution, we also should modify the behavior of icon-buttons: save and drop search:
- button save and drop should be visible until search is active

PR #1211, opened 2021-06-22
Add 'Import search filters' in menu

fur-gan:


DmitryAstafyev:
This solution would not work if the user doesn't have an active `Search` tab in the sidebar. As soon as you are subscribing to IPC messages in the scope of `application/client.core/src/app/environment/components/sidebar/search.manager/controls/component.ts`, message `IPCMessages.FiltersOpen` will be gotten only if that component is rendered (means: `Search` Tab is active).

Well I would suggest:
- move listener of IPC from component (also take in account, we are trying always keep away IPC from components, for IPC we have controllers/services)
- check if on IPC message `Search` tab is active -> if not -> open it
- call rest functionality

application/client.core/src/app/environment/components/sidebar/search.manager/controls/component.ts
```diff
@@ -4,13 +4,15 @@ import { ControllerSessionTabSearchStore } from '../../../../controller/session/
...
 import { Session } from '../../../../controller/session/session';
+import { IPCMessages } from '../../../../services/service.electron.ipc';
 
 import * as Toolkit from 'chipmunk.client.toolkit';
 
 import TabsSessionsService from '../../../../services/service.sessions.tabs';
 import HotkeysService from '../../../../services/service.hotkeys';
 import PopupsService from '../../../../services/standalone/service.popups';
 import EventsSessionService from '../../../../services/standalone/service.events.session';
+import ServiceElectronIpc from '../../../../services/service.electron.ipc';
```
DmitryAstafyev, line 15: Using components to listen IPC messages isn't good idea. Please use controllers/service to handle income IPC or sending IPC message to BE

DmitryAstafyev (changes requested):

application/electron/src/services/service.sessions.ts
```diff
@@ -0,0 +1,55 @@
+import Logger from '../tools/env.logger';
+import { IService } from '../interfaces/interface.service';
+import ServiceElectron, { IPCMessages, Subscription } from './service.electron';
+
+/**
+ * @class ServiceSessions
+ * @description Just keep information about sessions
+ */
+
+class ServiceSessions implements IService {
```
DmitryAstafyev, line 10: I think we don't need one more service, just for one additional IPC message. Even more, I don't think we need additional IPC (SessionChange). You can make little modification in:
```application/electron/src/services/service.streams.ts```
Like this:
```
    private _ipc_onStreamSetActive(message: IPCElectronMessages.TMessage) {
        if (!(message instanceof IPCElectronMessages.StreamSetActive)) {
            return;
        }
        if (this._activeStreamGuid === message.guid) {
            return;
        }
        if (this._streams.has(message.guid)) {
            // This is real session
            this._activeStreamGuid = message.guid;
            this._subjects.onSessionChanged.emit(message.guid);
            this._logger.debug(`Active session is set to: ${this._activeStreamGuid}`);
        } else {
            // This is custom tab (like About, Plugins, Settings etc.)
            this._logger.debug(
                `Active session would not set to: ${message.guid} as soon it isn't stream realated session`,
            );
        }
        ServiceElectron.updateMenu();
    }
```

**Review by DmitryAstafyev — CHANGES_REQUESTED** [review]

File: application/common/ipc/electron.ipc.messages/session.change.ts [hunk 3]
```diff
@@ -0,0 +1,20 @@
...
+    constructor(params: ISessionChange) {
+        if (typeof params !== 'object' || params === null) {
+            throw new Error(`Incorrect parameters for SessionChange message`);
+        }
+        if (typeof params.isSession !== 'boolean') {
+            throw new Error(`Expected isSession to be boolean`);
+        }
+        this.isSession = params.isSession
+    }
+}
```
DmitryAstafyev, line 20: See comment below, I think we can use just existed message: ```StreamSetActive```

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

PR #1212, opened 2021-06-22
Adjust popup width to stats width to show the arrows without scrolling

fur-gan:


DmitryAstafyev (changes requested):

DmitryAstafyev:
In general, I would not modify the popup window in general. Setting `width="auto"` doesn't make sense as soon as `width` is configurable property. 
Take a look here https://github.com/esrlabs/chipmunk/pull/1214 . All issue covered with `CSS`

PR #1246, opened 2021-09-21
Refresh process/device list upon opening dropdown list

fur-gan:


DmitryAstafyev (changes requested):
Next test-cases were failed:
Case A (plug device on open adb tab)
1. Do not connect device.
2. Open ADB tab (list of devices is empty - OK)
3. Connect device and again open list of devices; list is still empty - FAILED

Case B (unplug device)
1. Connect device
2. Open ADB tab (list of devices has your device - OK)
3. Disconnect (unplug) device
4. Open list of devices: device is still there - FAILED

Case C (error on device unplug)
1. Repeat all steps from Case B
2. Select in list `No Device is selected` -> Error

DmitryAstafyev (approved):

PR #1302, opened 2021-12-09
Multiple file selection from menu

fur-gan:


application/electron/src/services/files/service.file.opener.ts
```diff
@@ -653,15 +653,23 @@ class ServiceFileOpener implements IService {
             if (parser !== undefined) {
                 dialog
                     .showOpenDialog(win, {
-                        properties: ['openFile', 'showHiddenFiles'],
+                        properties: ['openFile', 'multiSelections', 'showHiddenFiles'],
```
DmitryAstafyev, line 656: Looks like we never use argument `parser` 
```
_openFile(parser?: AFileParser)
```
So we can remove it and also remove whole section ```if (parser !== undefined) { ... } ```

DmitryAstafyev (changes requested):

PR #1321, opened 2022-01-13
Remove y-axis on removed saved search

fur-gan:


application/client.core/src/app/environment/components/views/chart/chart/component.ts
```diff
@@ -531,6 +531,15 @@ export class ViewChartCanvasComponent implements AfterViewInit, AfterContentInit
...
             return;
         }
+        let scales: any | undefined = this._ng_charts?.options.scales;
+        if (scales !== undefined) {
+            Object.keys(scales).forEach((id: string) => {
+                if (id === 'x' || scale.yAxisIDs.includes(id)) {
+                    return;
+                }
+                scales[id].display = false;
+            });
```
DmitryAstafyev, line 541: why we cannot do just like it?
```
if (!scale.yAxisID.includes('y')) {
    scales.y.display = false;
}
```
And shell we back somewhere `scales.y.display = true`?

DmitryAstafyev (changes requested):

application/client.core/src/app/environment/components/views/chart/chart/component.ts
fur-gan: > why we cannot do just like it?
```
if (!scale.yAxisID.includes('y')) {
    scales.y.display = false;
}
```

`includes('y')` would only look for an axis with the name `y`, which is never the case because our y-axis are named `y-<saved regex>`
What we want to do there is to find all y-axes which are no longer included in `scales`. Since `scales` only has y-axes, we need to skip the `x-axis` as well and not modify it.

> And shell we back somewhere `scales.y.display = true`?
No, since that axis is removed we never put it back to `true`.

DmitryAstafyev (approved):

PR #1325, opened 2022-01-17
Warn user when opening unsupported files

fur-gan:


application/electron/src/controllers/files.parsers/file.parser.text.ts
```diff
@@ -57,34 +58,13 @@ export default class FileParser extends AFileParser {
...
-                            type.mime.indexOf('text') !== -1 ||
-                            type.mime.indexOf('application') !== -1
-                        ) {
-                            resolve(true);
-                        }
-                        resolve(false);
-                    },
-                );
-            });
+            readExactlyByFile(file, fs.statSync(file).size)
```
DmitryAstafyev, line 61: Do we need to read the whole file? What if it will be a 10Gb file?

DmitryAstafyev (changes requested):
The general issue is the amount of reading data. 
Here is a simple test:
- open chipmunk
- drag and drop >1Gb text file
Nothing happens because chipmunk reads all 1Gb data from file.
And while chipmunk reads it, you can drop again and again a couple of time same file.
As results chipmunk will start reading same file multiple times in the scope of the same session ... and it breaks things

Well, I think we should not read the whole file. I guess ~50Kb or even less would be okay to understand what we have.
And also, while we are detecting (guessing) a type of file, it should not be possible to drop something else. Probably it's enough to lock on FE only.

application/electron/src/controllers/files.parsers/file.parser.text.ts
fur-gan: Seems like the first 50 Byte suffice to decide if it's a binary file or not. Changed the code accordingly

DmitryAstafyev:
Do not remove your local branch. We will integrate it in the scope of grabber version

PR #1329, opened 2022-01-27
Grabber rs-binding and ts levels

DmitryAstafyev:


application/apps/indexer/session/src/operations.rs
```diff
@@ -98,18 +100,32 @@ impl std::fmt::Display for OperationAlias {
...
 
-#[derive(Debug, Clone)]
-pub enum Operation {
-    Assign {
-        file_path: PathBuf,
-        source_id: String,
+#[derive(Debug)]
+pub enum Source<T, P, S, D>
+where
+    T: LogMessage + 'static,
```
marcmo, line 106: how can the LogMessage have a static lifetime? 
A reference with a static lifetime means that the referenced object must live for the entire life of the program. 
The only way to create it is to create a static variable or a literal with a value known at compile time...and that's not what we have

DmitryAstafyev: No more static time

PR #1331, opened 2022-02-02
add someip-pcap cli command

kruss:


marcmo:
test failing: 
```
failures:
    raw::binary::tests::test_binary_load
```

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+        input: &'a [u8],
+        timestamp: Option<u64>,
+    ) -> Result<(&'a [u8], Option<SomeipLogMessage>), Error> {
+        let time = timestamp.unwrap_or(0);
+        match Message::from_slice(input) {
+            Ok(Message::Sd(header, payload)) => {
+                let len = header.message_len();
+                debug!("at {} : SD Message ({:?} bytes)", time, len);
+                Ok((
+                    &[],
```
marcmo, line 34: `parse` results in an optional `SomeipLogMessage` along with the rest of the data that was handed to the parser. So returning an empty [] is not correct.
`parse` will be called with some rather big buffer that can easily contain more data then just 1 message

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+                        input[..len].to_vec(),
+                    )),
+                ))
+            }
+
+            Ok(Message::Rpc(header, payload)) => {
+                let len = header.message_len();
+                debug!("at {} : RPC Message ({:?} bytes)", time, len);
+                Ok((
+                    &[],
```
marcmo, line 47: same thing, `header.len + payload.len` <= `input.len`

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+                        )),
+                        input[..len].to_vec(),
+                    )),
+                ))
+            }
+
+            Ok(Message::CookieClient) => {
+                debug!("at {} : MCC Message", time);
+                Ok((
+                    &[],
```
marcmo, line 63: same

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+                        None,
+                        input[..Header::LENGTH].to_vec(),
+                    )),
+                ))
+            }
+
+            Ok(Message::CookieServer) => {
+                debug!("at {} : MCS Message", time);
+                Ok((
+                    &[],
```
marcmo, line 75: same

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+            f,
+            "{}",
+            format!(
+                "SOME/IP {}{}",
+                self.name,
+                match &self.description {
+                    Some(description) => format!("\n{}", description),
+                    None => String::from(""),
+                },
+            )
```
marcmo, line 327: `write!` already takes a format string, so another `format!` will only create another copy on the heap...needs to be avoided, s.th. like
```
write!("f, "SOME/IP {}{}", self.name, ...)
```

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+}
+
+fn sd_payload_string(payload: &SdPayload) -> String {
+    let mut string = format!(
+        "- Flags: {}",
+        match (payload.reboot_flag(), payload.unicast_flag()) {
+            (true, false) => String::from("R"),
+            (false, true) => String::from("U"),
+            (true, true) => String::from("R,U"),
+            (false, false) => String::from("-"),
```
marcmo, line 112: no need to create another new String here, just use `&str` references

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+    );
+
+    for (i, entry) in payload.entries.iter().enumerate() {
+        let (entry_description, entry_options) = match entry {
+            SdEntry::FindService(value) => (
+                sd_service_entry_string(
+                    match value.has_ttl() {
+                        true => String::from("FindService"),
+                        false => String::from("StopFindService"),
+                    },
```
marcmo, line 123: more idiomatic:
```
                    String::from(if value.has_ttl() {
                        "FindService"
                    } else {
                        "StopFindService"
                    }),
```

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+        _ => None,
+    };
+
+    let som_type = match fibex_type {
+        Some(value) => match FibexTypes::build(value) {
+            Ok(result) => Some(result),
+            _ => None,
+        },
+        _ => None,
+    };
```
marcmo, line 249: better:
```
    let som_type = fibex_type.and_then(|value| FibexTypes::build(value).ok());
```

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    let signature = format!(
+        "{}::{}",
+        match service_name {
+            Some(value) => value,
+            _ => String::from("'Unknown Service'"),
+        },
+        match method_name {
+            Some(value) => value,
+            _ => String::from("'Unknown Method'"),
```
marcmo, line 265: avoid new String instances, use `&str` whenever possible

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+                Ok(_) => {
+                    format!("{}", value)
+                }
+                Err(error) => format!("{}", error),
+            }
+        }
+        _ => String::from("'Unknown Type'"),
+    };
+
+    format!("- {} {}", signature, content)
```
marcmo, line 286: best to write the whole expression in 1 `format!` expression so that we do not create unnecessary copies

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+            bytes,
+        }
+    }
+}
+
+impl LogMessage for SomeipLogMessage {
+    fn to_writer<W: Write>(&self, writer: &mut W) -> Result<usize, std::io::Error> {
+        let bytes = self.bytes.to_vec();
+        let len = bytes.len();
+        writer.write_all(&bytes)?;
```
marcmo, line 310: why create another copy of bytes? use
```
        writer.write_all(&self.bytes)?;
        Ok(self.bytes.len())
```

marcmo (changes requested):

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+use someip_messages::*;
+use someip_payload::fibex::FibexModel;
+use someip_payload::fibex2som::FibexTypes;
+use someip_payload::som::{SOMParser, SOMType};
+
+use log::debug;
+use serde::Serialize;
+
+pub struct SomeipParser<'m> {
+    model: &'m FibexModel,
```
marcmo, line 13: we need to be able to create a someip-parser even without a FibexModel. In this case we will only parse the header and display the payload as bytes.
So the datatype for `model` should be an `Option`

application/apps/indexer/parsers/src/someip.rs
kruss: Done. Raised fix for: https://github.com/esrlabs/someip/pull/12

application/apps/indexer/parsers/src/someip.rs
kruss: done.

application/apps/indexer/parsers/src/someip.rs
kruss: done.

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,330 @@
...
+                        | MessageType::Notification => method.get_request(),
+                        MessageType::Response => method.get_response(),
+                        _ => None,
+                    }
+                }
+                _ => None,
+            }
+        }
+        _ => None,
+    };
```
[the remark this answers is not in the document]
kruss: Cool!

application/apps/indexer/parsers/src/someip.rs
kruss: done.

application/apps/indexer/parsers/src/someip.rs
kruss: done.

application/apps/indexer/parsers/src/someip.rs
kruss: done.

application/apps/indexer/parsers/src/someip.rs
kruss: Applied where possible..

application/apps/indexer/parsers/src/someip.rs
kruss: done.

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -0,0 +1,562 @@
...
+    format!(
+        "{}: Service {}, Instance {}, Eventgroup {}, Version {}{}",
+        name,
+        entry.service_id,
+        entry.instance_id,
+        entry.eventgroup_id,
+        entry.major_version,
+        match entry.has_ttl() {
+            true => Cow::Owned(format!(", Ttl {}", entry.ttl)),
+            false => Cow::Borrowed(""),
```
marcmo, line 182: same

PR #1344, opened 2022-03-22
Grabber production integration

DmitryAstafyev:


application/apps/indexer/session/src/handlers/search.rs
```diff
@@ -5,46 +5,95 @@ use crate::{
...
+                            operation_api.emit(CallbackEvent::SearchUpdated(found as u64));
+                        }
+                        Err(err) => {
+                            break Err(err);
+                        }
+                    }
+                    if let Err(err) = state.update_search_result(search_res_file.clone()).await {
+                        break Err(err);
+                    }
+                }
```
marcmo, line 85: why update the search result twice?

application/apps/indexer/session/src/handlers/search.rs
```diff
@@ -5,46 +5,95 @@ use crate::{
...
+        });
+        let search_results = loop {
+            match timeout(
+                Duration::from_millis(TRACKING_INTERVAL_MS as u64),
+                rx_result.recv(),
+            )
+            .await
+            {
+                Ok(recv_results) => {
+                    break match recv_results {
```
marcmo, line 55: break match? never seen this before...what does it do?

marcmo (changes requested):

application/apps/rustcore/rs-bindings/Cargo.toml
```diff
@@ -28,3 +28,4 @@ thiserror = "1.0"
 tokio = { version = "1.5", features = ["full"] }
 tokio-util = "0.6"
 uuid = { version = "0.8.2", features = ["serde", "v4"] }
+dlt-core = "0.14"
```
marcmo, line 31: add newline

application/apps/indexer/processor/src/search.rs
```diff
@@ -248,7 +322,12 @@ impl SearchHolder {
                     &self.file_path, e
                 ))
             })?;
-
+        self.bytes_read = in_file_reader.seek(SeekFrom::Current(0)).map_err(|e| {
+            SearchError::IoOperation(format!(
```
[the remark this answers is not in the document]
DmitryAstafyev: We need to understand on which position we've finished reading (searing in) session file. As soon as method `seek` returns position in a file, I'm just "seeking" file to `0` to determine a current position of `BufReader` in file. 
Well `seek(SeekFrom::Current(0))` returns "cursor" position in file.

application/apps/indexer/session/src/handlers/observe.rs
```diff
@@ -311,16 +304,24 @@ async fn listen<T: LogMessage, P: Parser<T>, S: ByteSource>(
                                 kind: NativeErrorKind::Io,
                                 message: Some(e.to_string()),
                             })?;
-                        tx_event.send(Event::Check).map_err(|_| NativeError {
-                            severity: Severity::ERROR,
-                            kind: NativeErrorKind::ChannelError,
-                            message: Some(String::from("Fail to trigger update session state")),
-                        })?;
+                        if !state.is_closing() && last_message.elapsed().as_millis() > NOTIFY_IN_MS
```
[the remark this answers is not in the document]
DmitryAstafyev:
```
                        session_writer
                            .write_fmt(format_args!("{}\n", item))
                            .map_err(|e| NativeError {
                                severity: Severity::ERROR,
                                kind: NativeErrorKind::Io,
                                message: Some(e.to_string()),
                            })?;
```
I guess it's about it.
I think `to_writer` isn't suitable as soon as:
- we should "attach" `/n` to the end
- we should convert `item` to UTF8 text representation, actually, we should trigger `Display` trait on it.
Maybe I'm missing something, but I `write_fmt` looks like most suitable

application/apps/indexer/session/src/handlers/search.rs
DmitryAstafyev: sorry, "copy/paste" mistake

application/apps/indexer/session/src/handlers/search.rs
DmitryAstafyev: sorry, "copy/paste" mistake

application/apps/indexer/session/src/handlers/search.rs
DmitryAstafyev: that's a cool feature of rust, I like it
https://doc.rust-lang.org/rust-by-example/flow_control/loop/return.html
it's a way to return value from a `loop`. It's useful to determine the reason of breaking of loop: does happen because of error or it's expectable "exit"

DmitryAstafyev: yeah and `break match` here because I'm "overwriting" not only Err(...), but also Ok(...) hands.

DmitryAstafyev: probably with mapping it looks more clean
```
                    break recv_results.map_or(
                        Err(NativeError {
                            severity: Severity::ERROR,
                            kind: NativeErrorKind::OperationSearch,
                            message: Some("Fail to receive search results".to_string()),
                        }),
                        |(search_holder, search_results)| {
                            search_results
                                .map(|(file_path, matches, stats)| {
                                    (file_path, matches.len(), matches, stats, search_holder)
                                })
                                .map_err(|err| NativeError {
                                    severity: Severity::ERROR,
                                    kind: NativeErrorKind::OperationSearch,
                                    message: Some(format!(
                                        "Fail to execute search. Error: {}",
                                        err
                                    )),
                                })
                        },
                    );

```

application/apps/indexer/session/src/handlers/observe.rs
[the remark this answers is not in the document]
marcmo: you are right...the `to_writer` interface is not about string representation!

marcmo:
please also locally run `cargo cippy` and fix the 9 warnings @DmitryAstafyev

PR #1370, opened 2022-05-13
Serial feature

fur-gan:


application/apps/indexer/sources/src/serial/serialport.rs
```diff
@@ -0,0 +1,131 @@
...
+    ) -> Result<Option<ReloadInfo>, SourceError> {
+        let mut copied: usize = 0;
+        let mut buffer_error: String = "".to_string();
+        let timeout = tokio::time::timeout(Duration::from_millis(TIMEOUT), async {
+            loop {
+                if let Err(err) = self.port.read_exact(&mut self.read).await {
+                    buffer_error = format!("read from serial stream failed: {}", err);
+                };
+                copied += self.buffer.copy_from_slice(&self.read);
+            }
```
marcmo, line 55: wait a second...is this really working? we basically never break out of the loop until we timeout. But in that case we always return `ReloadInfo::new(0, 0, 0, None)`

PR #1395, opened 2022-06-07
Re-implemented search/match algorithm

fur-gan:


DmitryAstafyev:
is this pull-request still actual as soon as we will go with wasm instead node-binding?

fur-gan:
> is this pull-request still actual as soon as we will go with wasm instead node-binding?

This is the PR for node-bindgen, not wasm. So I'll close this PR now and create a new one for the wasm implementation.

PR #1415, opened 2022-07-15
Add sorting and rework matcher

fur-gan:


application/apps/webassembly/matcher/src/lib.rs
```diff
@@ -25,81 +26,75 @@ impl Matcher {
...
-                Ok(tagged_match) => Ok(tagged_match),
-                Err(err) => Err(err),
-            },
-            None => Ok(item),
+    pub fn set_item(&mut self, item: String) -> Result<usize, String> {
+        match serde_json::from_str(&item) {
+            Ok::<HashMap<String, String>, _>(item) => {
+                self.items_initial.push(item);
+                Ok(self.items_initial.len() - 1)
+            }
```
DmitryAstafyev, line 40: Let's add here `self.search(...)` with empty filter to init scores

application/client/src/app/ui/elements/recent/state.ts
```diff
@@ -47,13 +49,15 @@ export class State {
...
+        this.matcher.search(this.filter.value());
+        this.actions.sort((a: WrappedAction, b: WrappedAction) => b.getScore() - a.getScore());
         this.update.emit();
     }
 
     public getFilteredActions(): WrappedAction[] {
-        return this.actions.filter((a) => a.filtered);
+        if (this.filter.value() === '') {
+            return this.actions;
+        }
```
DmitryAstafyev, line 60: if score would be inited (on rust level or even on abstract level), we don't need to check filter value

application/client/src/app/ui/elements/recent.mini/state.ts
```diff
@@ -5,17 +5,19 @@ import { recent } from '@service/recent';
...
-export class State {
+export class State extends Holder {
     public filter: Filter;
     public actions: WrappedAction[] = [];
     public update: Subject<void> = new Subject<void>();
     public selected: string = '';
     protected close: CloseHandler | undefined;
 
     constructor(ilc: IlcInterface & ChangesDetector) {
+        super();
```
DmitryAstafyev, line 20: BTW, if you need to destroy matcher as soon as state's parent component was killed, you can add this:
```
ilc.life().destroy(() => {
            this.matcher.destroy();
            // it will be called as soon as component would be destroyed.
});
```

DmitryAstafyev (changes requested):

application/client/src/app/ui/elements/recent.mini/state.ts
fur-gan: We don't have a dedicated function to destroy the matcher.
An alternative that comes to my mind would be to call `delete this.matcher` but that is not really necessary, because that's being done internally anyway.

application/apps/webassembly/matcher/src/lib.rs
```diff
@@ -25,81 +26,81 @@ impl Matcher {
...
+    pub fn set_item(&mut self, item: String) -> Result<usize, String> {
+        match serde_json::from_str(&item) {
+            Ok::<HashMap<String, String>, _>(item) => {
+                self.items_initial.push(item);
+                match self.search(String::new(), None) {
+                    Ok(_) => {}
+                    Err(err) => {
+                        return Err(format!("Initializing with empty search failed: {}", err))
+                    }
+                }
```
DmitryAstafyev, line 44:
```
match self.search(String::new(), None) {
                    Ok(_) => {}
                    Err(err) => {
                        return Err(format!("Initializing with empty search failed: {}", err))
                    }
                }
```

                
this block can be replaced with: `self.search(String::new(), None).map_err(|e| format!("Initializing with empty search failed: {}", e))?;`

DmitryAstafyev (approved):

PR #1495, opened 2022-09-02
Concatenation feature

fur-gan:


application/client/src/app/ui/tabs/sources/multiplefiles/structure/component.ts
```diff
@@ -0,0 +1,191 @@
...
+        file.reverseSelect();
+        this.context.next(null);
+    }
+
+    public ngSize(size: number): string {
+        return bytesToStr(size);
+    }
+
+    public ngContextMenu(event: any, file?: FileHolder) {
+        event.stopImmediatePropagation();
```
DmitryAstafyev, line 86: is it really needed? as Soon as it's listener of `contextmenu` event, I think no need to stop it.

application/client/src/app/ui/tabs/sources/multiplefiles/structure/component.ts
```diff
@@ -0,0 +1,191 @@
...
+    public ngOnRowSelect(file: FileHolder) {
+        file.reverseSelect();
+        this.context.next(null);
+    }
+
+    public ngSize(size: number): string {
+        return bytesToStr(size);
+    }
+
+    public ngContextMenu(event: any, file?: FileHolder) {
```
DmitryAstafyev, line 85: event here is `event: MouseEvent` not  any

application/client/src/app/ui/tabs/sources/multiplefiles/structure/component.ts
```diff
@@ -0,0 +1,191 @@
...
+                },
+            },
+            {
+                caption: 'Reverse Select All',
+                handler: () => {
+                    this.data.data.forEach((f: FileHolder) => f.reverseSelect());
+                    this.context.next(null);
+                },
+            },
+        );
```
DmitryAstafyev, line 119: you are welcome to use delimiters in the menu to make it more organized, just add an empty object
```
items.push(
            {
                caption: 'Select All',
                handler: () => { ... },
            },
            {
                caption: 'Unselect All',
                handler: () => { ... },
            },
            { /* This is delimiter */},
            {
                caption: 'Reverse Select All',
                handler: () => { ... },
            },
        );
```
I don't ask to put a delimiter like in the example, just to show how you can do it. You can add delimiters between groups of commands. Like selections items, removing items, and so on.

application/client/src/app/ui/tabs/sources/multiplefiles/component.ts
```diff
@@ -0,0 +1,115 @@
...
+                            });
+                        break;
+                    case FileType.Dlt:
+                        this.ilc()
+                            .services.system.opener.file(file.file)
+                            .dlt()
+                            .catch((err: Error) => {
+                                this.log().error(`Fail to open dlt file; error: ${err.message}`);
+                            });
+                        break;
```
DmitryAstafyev, line 76: ohh it would be nightmare ). If I select 10 DLT files it will open 10 tabs for setup options )... I don't have yet solution in mind what to do with it.

DmitryAstafyev: Probably at the moment, command open each should be available only for text files

application/client/src/app/ui/tabs/sources/multiplefiles/component.ts
```diff
@@ -0,0 +1,115 @@
...
+    public ngSelectedCount: number = 0;
+    public ngSelectedSize: string = '0 b';
+    public ngDestination: string = '';
+    public ngFiles: FileHolder[] = [];
+
+    private _files: FileHolder[] = [];
+
+    constructor() {
+        super();
+    }
```
DmitryAstafyev, line 39: not necessary to define constructor in this case

application/client/src/app/ui/tabs/sources/multiplefiles/component.ts
```diff
@@ -0,0 +1,115 @@
...
+                            .services.system.opener.file(file.file)
+                            .dlt()
+                            .catch((err: Error) => {
+                                this.log().error(`Fail to open dlt file; error: ${err.message}`);
+                            });
+                        break;
+                }
+            }
+        });
+        // TODO - Close tab
```
DmitryAstafyev, line 80: To close tab:
```
import { TabControls } from '@service/session';
...
@Input() tab!: TabControls;
...
private _example(): void {
     this.tab.close();
}
...
```
It will close a tab.
Also, I would put your attention to the next available methods in `TabControlls`:
```
// Will save your stat in a storage of tab
this.tab.storage<YOUR_STATE_TYPE>().set(this.state)
// Will extract a state
const state: YOUR_STATE_TYPE = this.tab.storage<YOUR_STATE_TYPE>.get()
```
You can use tab's storage to save some state of your view. If user will switch a tab and back to your tab again you will have a chance to restore view like it was before user "jumped" out

application/client/src/app/ui/tabs/sources/multiplefiles/component.ts
```diff
@@ -0,0 +1,115 @@
...
+    private _files: FileHolder[] = [];
+
+    constructor() {
+        super();
+    }
+
+    public ngAfterContentInit(): void {
+        this._files = this.files.map((file: File) => new FileHolder(this.matcher, file));
+        this.ngFiles = this._files;
+        this.ngOnContext(null);
```
DmitryAstafyev, line 44: good place to restore state of your view (for example if it was opened, but user "jumped" to another tab and back
in `ngOnDestroy` you can save state. See API bellow in comments

application/client/src/app/ui/tabs/sources/multiplefiles/component.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    public action: Action = new Action();
+    public ngCompatible: boolean = false;
+    public ngSelectedCount: number = 0;
+    public ngSelectedSize: string = '0 b';
+    public ngDestination: string = '';
+    public ngFiles: FileHolder[] = [];
+
+    private _files: FileHolder[] = [];
```
DmitryAstafyev, line 35: all these stuff can be grouped into State, which you can use after to save and load (to support tab's switching). In layout you also can directly access to state, like `{{state.size}}` instead `ngSelectedSize`

application/client/src/app/ui/tabs/module.ts
```diff
@@ -14,6 +15,7 @@ import { TimezoneSelectorModule } from '@elements/timezones/module';
         TabSourceDltStreamModule,
         TimezoneSelectorModule,
         TabSourceTextStreamModule,
+        TabSourceMultipleFilesModule,
```
DmitryAstafyev, line 18: add `TabSourceMultipleFilesModule` also into `bootstrap` section bellow to prevent production bug

application/client/src/app/ui/layout/workspace/no-tabs-content/component.ts
```diff
@@ -53,32 +54,47 @@ export class LayoutWorkspaceNoContent {
...
+                        .services.system.session.getTabsService()
+                        .add({
+                            name: 'Multiple Files',
+                            active: true,
+                            closable: true,
+                            content: {
+                                factory: TabSourceMultipleFiles,
+                                inputs: {
+                                    files: files,
+                                },
```
DmitryAstafyev, line 94: This is just a note. Nothing to change/fix. 
It's become so complicated to be in component, probably we should move it right into service. We can discuss this question on a call.

DmitryAstafyev (changes requested):

application/client/src/app/ui/layout/workspace/no-tabs-content/component.ts
DmitryAstafyev: In general. I think as a minimum solution we also should support drag and drop. I mean the user should have a possibility to drop file(s) right into view from his file explorer.

DmitryAstafyev:
In general. I think as a minimum solution we also should support drag and drop. I mean the user should have the possibility to drop file(s) right into view from his file explorer.

application/client/src/app/ui/tabs/sources/multiplefiles/structure/component.ts
fur-gan: We use `ngContextMenu` in both `<table>` (line 1) and `<tr>` (line 40). If we were to call the context menu on the `<tr>`, it would be called on the table row first and then on the table as well, because the table itself is the parent.

application/client/src/app/ui/layout/workspace/no-tabs-content/component.ts
```diff
@@ -53,32 +54,47 @@ export class LayoutWorkspaceNoContent {
...
+                                break;
+                        }
+                    });
+                } else {
+                    this.ilc()
+                        .services.system.session.getTabsService()
+                        .add({
+                            name: 'Multiple Files',
+                            active: true,
+                            closable: true,
```
DmitryAstafyev, line 89: use more high level API:
```
this.ilc().services.system.session.add().tab(
                      {
                            name: 'Multiple Files',
                            active: true,
                            closable: true,
                            content: {
                                factory: TabSourceMultipleFiles,
                                inputs: {
                                    files: files,
                                },
                       }
);
```
it will give you more `inputs` (including `TabControls`) in your component

DmitryAstafyev (changes requested):

application/client/src/app/ui/tabs/sources/multiplefiles/component.ts
```diff
@@ -0,0 +1,201 @@
...
+                this.ilc()
+                    .services.system.bridge.files()
+                    .getByPath([item.path])
+                    .then((result) => {
+                        if (result.length !== 1) {
+                            this.log().error(`Invalid file stat info`);
+                            return;
+                        }
+                        if (item != null) {
+                            this.state.files.push(
```
DmitryAstafyev, line 99: you check two time `item !== null`. To make Typescript happy and have single check you can reassign the value:
```
if (item === null) {
    continue;
}
let file = item;
// do all stuff
```
As soon as you created new variable `file` you don't need to double check in in callbacks bellow in promisses.

DmitryAstafyev: In general I would reorganize a little this method.
1. Here looks pretty suitable usage of `Promise.all` to understand when all files processed
2. I would add progress bar to indicate start and finish processing

application/client/src/app/ui/tabs/sources/multiplefiles/component.ts
```diff
@@ -0,0 +1,201 @@
...
+                    this._selectedTypes.push(file.type);
+                }
+            }
+        });
+        this.state.selectedSize = bytesToStr(size);
+
+        this.state.openable = !this._selectedTypes.includes(FileType.Dlt);
+        this.state.concatable =
+            this._selectedTypes.length === 1 && this._selectedTypes.includes(FileType.Text);
+    }
```
DmitryAstafyev, line 188: in this method, you almost manipulate with `state`. Would it be better to move this method just into state?

application/client/src/app/ui/tabs/sources/multiplefiles/component.ts
DmitryAstafyev: BWT `getByPath(filenames: string[])` accept an array of files, well you can get info for all just with one call, so no need `Promise.all`, no need loop

application/client/src/app/ui/tabs/sources/multiplefiles/file.holder.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export interface IFileInfo {
+    filename: string;
+    modificationDate: number;
+    name: string;
+    path: string;
+    size: number;
+    type: FileType | string;
+}
```
DmitryAstafyev, line 26: I am not sure, why you need one more type definition for file?
`application/platform/types/files.ts: File` looks like has everything what you need:
```
export interface File {
    name: string; //name of file
    ext: string; //extention
    path: string; //path
    filename: string; //full file name
    type: FileType; //type of file (DLT, pcap, text)
    stat: Stat; 
}
```

DmitryAstafyev (changes requested):

application/client/src/app/ui/tabs/sources/multiplefiles/component.ts
```diff
@@ -0,0 +1,201 @@
...
+        // TODO - Concatenate files
+        // TODO - Close tab - if DLT new Tab
+    }
+
+    public ngOnDestination() {
+        // TODO - Open file explorer to select path
+    }
+
+    public ngOnDrop(fileList: any) {
+        for (let i = 0; i < fileList.length; i++) {
```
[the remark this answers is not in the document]
fur-gan: Yeah, it's possible, included the solution in the next commit

DmitryAstafyev (approved):

PR #1501, opened 2022-09-15
Support sending messages on serial connection

fur-gan:


application/client/src/app/ui/elements/transport/details/serial/component.ts
```diff
@@ -16,8 +22,47 @@ export class TransportSerial extends ChangesDetector {
     @Input() public source!: SerialTransportSettings;
     @Input() public session!: Session;
 
+    @ViewChild('stdin') public input!: AutocompleteInput;
+
```
DmitryAstafyev, line 26: id = "stdin" was for commands as soon as it was sending to `stdin` on child process )
It isn't an error, but just could confuse developer (even you in a couple of months)

application/apps/indexer/sources/src/serial/serialport.rs
```diff
@@ -1,11 +1,46 @@
...
+
+impl Encoder<String> for LineCodec {
+    type Error = io::Error;
+
+    fn encode(&mut self, item: String, dst: &mut BytesMut) -> Result<(), Self::Error> {
+        dst.reserve(item.len() + 1);
+        dst.put(item.as_bytes());
+        dst.put_u8(b'\n');
+        Ok(())
+    }
```
DmitryAstafyev, line 42: I am not sure, but here you get as input `String` and after you "put" this string as bytes into buffer.
But... bellow in `income` income you gets bytes and convert it into string... Looks a little bit wierd. In general it looks like it:
- FE -> Rust:  sends bytes
- Rust (income method): convert bytes to string
- Rust (income method): send string into `write_stream`
- Rust (LineCodec / encode): convert string to bytes and write into buffer

If we are writing bytes into buffer, does it make sense to convert it into string?

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

marcmo (approved):

PR #1504, opened 2022-09-26
TCP stream UI

fur-gan:


application/client/src/app/ui/elements/transport/setup/component.ts
```diff
@@ -30,7 +30,9 @@ export class Transport extends ChangesDetector {
 
     public ngOnSourceChange() {
         this.state.switch();
-        this.action.defaults();
+        if (this.action !== undefined) {
+            this.action.defaults();
+        }
```
DmitryAstafyev, line 35: why `action` could be `undefined`? It should not be possible as soon as it's `@Input`. If it happens, it means there are some bug on parent component level

DmitryAstafyev (changes requested):

application/client/src/app/ui/elements/transport/setup/component.ts
fur-gan: In `application/client/src/app/ui/tabs/sources/dltstream/template.html:16` we have:
`<app-transport [state]="state.transport"></app-transport>`
I thought `action` wasn't provided on purpose which is why I implemented this "workaround".
Will just add `action` and remove the `if`-check instead

DmitryAstafyev (approved):

PR #1510, opened 2022-10-11
Charts feature

fur-gan:


application/client/src/app/ui/service/toolbar.ts
```diff
@@ -7,13 +7,15 @@ import { unique } from '@platform/env/sequence';
...
 import { Tabs } from './toolbar/tabs';
+import { Chart } from './toolbar/chart';
 
 @SetupService(ui['toolbar'])
 export class Service extends Implementation {
     private _emitter!: Emitter;
     private _channel!: Channel;
     private _services!: Services;
     private _available!: Tabs;
+    private _chart!: Chart;
```
DmitryAstafyev, line 18: Chart service should not be a part of the toolbar service. It should "live" in the scope of the chart view.

application/client/src/app/ui/service/toolbar.ts
```diff
@@ -137,6 +140,10 @@ export class Service extends Implementation {
         });
     }
 
+    public get chart(): Chart {
+        return this._chart;
+    }
+
```
DmitryAstafyev, line 146: should not be here. Toolbar service is only for managing of toolbar: add/remove, active/deactive.

application/client/src/app/ui/service/toolbar/chart.ts
```diff
@@ -0,0 +1,45 @@
+import { Observable, Subject } from 'rxjs';
+
+export interface IPositionChange {
+    left: number;
+    width: number;
+    full: number;
+}
+
+export class Chart {
```
DmitryAstafyev, line 9: this class "lives" in the incorrect place. The chart isn't related to the toolbar. Actually chart "doesn't know" where is "he" would be rendered.

application/client/src/app/ui/views/toolbar/chart/zoomer/cursor/component.ts
```diff
@@ -0,0 +1,50 @@
...
+@Ilc()
+export class ViewChartZoomerCursorCanvas extends ChangesDetector implements AfterViewInit {
+    @Input() getLeftOffset: () => number = () => 0;
+
+    public state: State = new State();
+
+    constructor(cdRef: ChangeDetectorRef, private _vcRef: ViewContainerRef) {
+        super(cdRef);
+        window.addEventListener('mousemove', this.state.onWindowMousemove);
+        window.addEventListener('mouseup', this.state.onWindowMouseup);
```
DmitryAstafyev, line 27: you are linking window events to state methods.In destroy handler bellow you are unsubscribing from that. Would it be better just move it into state? You have already `state.init(...)` well subscription could happen there

application/client/src/app/ui/views/toolbar/chart/zoomer/cursor/component.ts
```diff
@@ -0,0 +1,50 @@
...
+        this.state.onClick(event);
+    }
+
+    public ngAfterViewInit() {
+        this.state.init(this.getLeftOffset.bind(this), this.detectChanges.bind(this), this._vcRef);
+    }
+
+    public ngOnDestroy() {
+        window.removeEventListener('mousemove', this.state.onWindowMousemove);
+        window.removeEventListener('mouseup', this.state.onWindowMouseup);
```
DmitryAstafyev, line 46: could be replaced with something like `state.destroy()`

application/client/src/app/ui/views/toolbar/chart/zoomer/cursor/state.ts
```diff
@@ -0,0 +1,274 @@
+import { InternalAPI } from '@service/ilc';
+import { ViewContainerRef } from '@angular/core';
+import { Chart, IPositionChange } from '@ui/service/toolbar/chart';
+
+enum EChangeKind {
+    move = 'move',
+    left = 'left',
+    right = 'right',
+    undefined = 'undefined',
```
DmitryAstafyev, line 9: you are using the reserved word `undefined`... it could be a source of errors or at least confuse. Please choose another variant-name... like `notInited` or any other

application/client/src/app/ui/views/toolbar/chart/zoomer/cursor/state.ts
```diff
@@ -0,0 +1,274 @@
...
+}
+
+const CSettings = {
+    minSize: 20,
+};
```
DmitryAstafyev, line 14: for single constant doesn't make sense to create object / group... better would be 
```
const MIN_SIZE_OF_SOMETHING = 20;
```

application/client/src/app/ui/views/toolbar/chart/zoomer/cursor/state.ts
```diff
@@ -0,0 +1,274 @@
...
+
+export class State {
+    public width: number = -1;
+    public left: number = 0;
+
+    private _ilc!: InternalAPI;
+    private _chart!: Chart;
+    private _detectChanges!: () => void;
```
DmitryAstafyev, line 22: actually if you need many stuff from parent component you can just send into state component as it is.
In state
```
    public bind(parent: IlcInterface & ChangesDetector) {
        // and you will get access to all
        parent.ilc().services.system.bridge;
        parent.env().subscriber.register(...);
        parent.detectChanges();
    }

```
In parent component:
```
    public ngAfterContentInit(): void {
        ...
        this.state.bind(this);
        ...
    }
```

application/client/src/app/ui/views/toolbar/chart/zoomer/state.ts
```diff
@@ -0,0 +1,233 @@
...
+    private _session!: Session;
+    private _logger!: Logger;
+    private _labels: string[] = [];
+    private _datasets: any[] = [];
+    private _map: ISearchMap = [];
+    private _detectChanges!: () => void;
+    private _filters!: Chart;
+    private _loading: boolean = false;
+    private _env!: Env;
+    private _ilc!: InternalAPI;
```
DmitryAstafyev, line 21: see previous comment about accessing to parent component's features

application/client/src/app/ui/views/toolbar/chart/zoomer/state.ts
```diff
@@ -0,0 +1,233 @@
...
+            this._createChart();
+        } else {
+            this._filters.data = {
+                datasets: this._datasets,
+                labels: this._labels,
+            };
+            this._filters.update();
+        }
+        this._loading = false;
+        this._detectChanges();
```
DmitryAstafyev, line 148: this piece of code looks as synchronic code: started at line 136 and finished on 148. It isn't clear why you need flag `this._loading` here? It make sense if you are doing some async stuff. I'm confused a bit with it

application/client/src/app/ui/views/toolbar/module.ts
```diff
@@ -2,10 +2,11 @@ import { NgModule } from '@angular/core';
 import { CommonModule } from '@angular/common';
 import { SearchModule } from './search/module';
 import { HistoryModule } from './history/module';
+import { ChartModule } from './chart/module';
```
DmitryAstafyev, line 5: as soon as Chart component is `@Initial` (should be available for run-time injecting) it should be added also into `bootstrap` section of module
```
@NgModule({
    declarations: [...],
    imports: [...],
    ....
    bootstrap: [ChartModule],
})
```
and in `ChartModule` chart-component also should be included into `bootstrap`.

It will make Angular preload all related to chart stuff on booting stage

application/client/src/app/service/session/session.ts
```diff
@@ -68,6 +68,17 @@ export class Session extends Base {
...
+        this._toolbar.add({
+            name: 'Chart',
+            active: false,
+            closable: false,
+            content: {
+                factory: components.get('app-views-chart'),
+                inputs: {
+                    session: this,
+                },
+            },
```
DmitryAstafyev, line 80: `uuid` field is missed
you can add uuid into `application/client/src/app/service/session/session.ts`:
```
const TOOLBAR_TAB_SEARCH = unique();
const TOOLBAR_TAB_PRESET = unique();
const TOOLBAR_TAB_CHART = unique();

```

application/client/src/app/service/session/dependencies/search/store.ts
```diff
@@ -17,12 +17,12 @@ export type StoredEntity<T> = T & Hash & Recognizable & Updatable<EntityUpdateEv
...
 export abstract class Store<T> extends Subscriber {
     public subjects: Subjects<{
-        highlights: Subject<StoredEntity<T>[]>;
+        highlights: Subject<StoredEntity<T>>;
         value: Subject<StoredEntity<T>[]>;
         inner: Subject<StoredEntity<T>[]>;
         any: Subject<StoredEntity<T>[]>;
     }> = new Subjects({
-        highlights: new Subject<StoredEntity<T>[]>(),
+        highlights: new Subject<StoredEntity<T>>(),
```
DmitryAstafyev, line 25: what is the motivation to change the signature of `highlights` event? Would it break other parts of solution?

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/chart/chart/state.ts
```diff
@@ -0,0 +1,261 @@
...
+
+export class State {
+    private _ilc!: InternalAPI;
+    private _env!: Env;
+    private _session!: Session;
+    private _logger!: Logger;
+    private _labels: string[] = [];
+    private _datasets: any[] = [];
+    private _map: ISearchMap = [];
+    private _detectChanges!: () => void;
```
DmitryAstafyev, line 23: bellow you will find a comment, how to get access to all features of parent components (detectChages, ilc, etc)

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/chart/chart/state.ts
```diff
@@ -0,0 +1,261 @@
...
+        this._env.subscriber.register(
+            this._session.search
+                .store()
+                .filters()
+                .subjects.get()
+                .highlights.subscribe((entity: StoredEntity<FilterRequest>) => {
+                    this._changeColor(entity.definition.uuid, entity.definition.colors.background);
+                }),
+        );
+        this._ilc.channel.ui.sidebar.resize(this._resize(containerWidth)._onUpdate.bind(this));
```
DmitryAstafyev, line 82:
```
this._ilc.channel.ui.sidebar.resize(this._resize(containerWidth)._onUpdate.bind(this));
```
it would not work like you want probably.
you are executing `this._resize(containerWidth)` just once during subscribing. And resize never will be triggered on the event.
if you want to resize on event, you shouldn't bind here
```
this._ilc.channel.ui.sidebar.resize(() => {
    // how to get actual value for "containerWidth"?
    this._resize(containerWidth)._onUpdate();
})
```

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/chart/chart/state.ts
```diff
@@ -0,0 +1,377 @@
...
+            },
+        });
+    }
+
+    private _fetch(width: number): Promise<void> {
+        return this._session.search
+            .getScaledMap(width)
+            .then((map) => {
+                this._map = map;
+                this._labels = new Array(this._map.length).fill('');
```
DmitryAstafyev, line 370: may be I've missed something, but looks like `this._labels` has empty strings only. Do we need to keep it as property in this case? Or it's performance question (to avoid recreating array each time)?

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/chart/chart/state.ts
fur-gan: Yes, it's for performance only. We need labels or the data won't be drawn in the chart. Because we won't show the axis anyway, it's just filled with empty strings.
Maybe it makes sense to recreate the array every time anew since it's rare that the array is reused.

application/client/src/app/ui/views/toolbar/chart/zoomer/cursor/state.ts
```diff
@@ -0,0 +1,280 @@
...
+
+export class State {
+    public _ng_width: number = -1;
+    public _ng_left: number = 0;
+
+    private _parent!: IlcInterface & ChangesDetector;
+    private _session!: Session;
+    private _service!: Service;
+    private _leftOffset = () => 0;
+    private _vcRef!: ViewContainerRef;
```
DmitryAstafyev, line 25: you can store is as `HTMLElement` no need to store reference to Angular container

application/client/src/app/ui/views/toolbar/chart/zoomer/cursor/state.ts
```diff
@@ -0,0 +1,280 @@
...
+        }
+        this._mouse.x = event.x;
+        this._mouse.kind = EChangeKind.move;
+    }
+
+    public onLeft(event: MouseEvent) {
+        this._mouse.x = event.x;
+        this._mouse.kind = EChangeKind.left;
+        event.preventDefault();
+        event.stopImmediatePropagation();
```
DmitryAstafyev, line 143: you can use method `stop` from `@ui\env\dom` to stop event

application/client/src/app/ui/views/toolbar/chart/zoomer/cursor/state.ts
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    public onRight(event: MouseEvent) {
+        this._mouse.x = event.x;
+        this._mouse.kind = EChangeKind.right;
+        event.preventDefault();
+        event.stopImmediatePropagation();
```
DmitryAstafyev, line 150: you can use method `stop` from `@ui\env\dom` to stop event

application/client/src/app/ui/views/toolbar/chart/zoomer/cursor/state.ts
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    public preventDefault(event: MouseEvent) {
+        stop(event);
+        return false;
+    }
```
DmitryAstafyev, line 156: you can use method `stop` from `@ui\env\dom` to stop event. Actually you don't need this method as a part of state.

application/client/src/app/ui/views/toolbar/chart/zoomer/cursor/state.ts
```diff
@@ -0,0 +1,280 @@
...
+                width: this._ng_width,
+                full: this._width,
+            },
+        });
+    }
+
+    private _restore() {
+        if (this._vcRef === undefined) {
+            return;
+        }
```
DmitryAstafyev, line 267: if it's possible please defile right time for field
```
private _vcRef: HTMLelement | undefined;
```

DmitryAstafyev: it will allow TypeScript compiler also help us to cover all use-cases

application/client/src/app/ui/views/toolbar/chart/zoomer/state.ts
```diff
@@ -0,0 +1,230 @@
...
+import { getChartService, Service } from '../service';
+
+export class State {
+    public isCursorVisible: boolean = true;
+
+    private _parent!: IlcInterface & ChangesDetector;
+    private _filters!: Chart;
+    private _session!: Session;
+    private _service!: Service;
+    private _vcRef!: ViewContainerRef;
```
DmitryAstafyev, line 19: You can store `HTMLElement` and pass it into state from component... No need to manipulate in state with angular-component reference

application/client/src/app/ui/views/toolbar/chart/zoomer/state.ts
```diff
@@ -0,0 +1,230 @@
...
+    private _map: ISearchMap = [];
+    private _loading: boolean = false;
+    private _width: number = 0;
+
+    public bind(
+        parent: IlcInterface & ChangesDetector,
+        session: Session,
+        vcRef: ViewContainerRef,
+        logger: Logger,
+    ) {
```
DmitryAstafyev, line 32: BTW, as soon as you pass `parent` as `IlcInterface`, you don't need `logger: Logger,`, because you can write logs directly `parent.log().error('bla-bla')`

DmitryAstafyev: one more field out - less complexity, less bugs

application/client/src/app/ui/views/toolbar/chart/zoomer/state.ts
```diff
@@ -0,0 +1,230 @@
...
+                    }
+                });
+        });
+        this._filters.update();
+    }
+
+    private _resize(): State {
+        this._parent.detectChanges();
+        this._width = this._vcRef.element.nativeElement.clientWidth;
+        return this;
```
DmitryAstafyev, line 111: Use please correct way to get size of node element

application/client/src/app/ui/views/toolbar/chart/zoomer/state.ts
```diff
@@ -0,0 +1,230 @@
...
+            this._createChart();
+        } else {
+            this._filters.data = {
+                datasets: this._datasets,
+                labels: new Array(this._labelCount).fill(''),
+            };
+            this._filters.update();
+        }
+        this._loading = false;
+        this._parent.detectChanges();
```
DmitryAstafyev, line 146: Not clear how flag `loading` can "update" something as soon as here no any async code. Maybe I'm wrong :/

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/chart/zoomer/state.ts
fur-gan: No no, you're right. I wanted to move the `this._loading = true` into the async method  `_fetch(width: number)` where we update our data. Must've missed it when I was resolving the comments

DmitryAstafyev (approved):

PR #1516, opened 2022-10-14
docs: update macos install method

nhymxu:
I added chipmunk to Homebrew cask repo.
So user can install using Homebrew

https://github.com/Homebrew/homebrew-cask/pull/133440


DmitryAstafyev:
Thank you for your cooperation! It will be merged pretty soon

PR #1549, opened 2022-12-05
Fixes

fur-gan:
**Final checklist to Chipmunk 3**
B.17. Flags on search don't work
B.19. Matches marker not aligned with scrollbar

**Bugs & Quirks Chipmunk 3**
S.4. the cross symbol should either be a button or be removed
B.10. enter will only work once
B.13. removing multicast sections not possible
Q.2. what is “Show matches”?

### DIFF
--- a/application/client/src/app/service/session/dependencies/search/state.ts
+++ b/application/client/src/app/service/session/dependencies/search/state.ts
@@ -11,11 +11,13 @@ export class State {
         collection: Subject<FilterRequest[]>;
         start: Subject<void>;
         finish: Subject<number>;
+        showMatches: Subject<IFilter>;
     }> = new Subjects({
         active: new Subject<IFilter | undefined>(),
         collection: new Subject<FilterRequest[]>(),
         start: new Subject<void>(),
         finish: new Subject<number>(),
+        showMatches: new Subject<IFilter>(),
     });
 
     private _search: Search;
@@ -115,5 +117,9 @@ export class State {
         this._hash = undefined;
         this.subjects.get().active.emit(undefined);
         return this.filters();
+    }
+
+    public showMatches(filter: IFilter) {
+        this.subjects.get().showMatches.emit(filter);
     }
 }
--- a/application/client/src/app/service/session/dependencies/search/state.ts
+++ b/application/client/src/app/service/session/dependencies/search/state.ts
@@ -1,28 +1,40 @@
-import { FilterRequest } from './filters/request';
 import { IFilter } from '@platform/types/filter';
 import { Subjects, Subject } from '@platform/env/subscription';
 import { Search } from '@service/session/dependencies/search';
 
 import * as obj from '@platform/env/obj';
 
+export interface IFinish {
+    found: number;
+    error?: string;
+}
+
+export interface ISearchInfo {
+    state: ESearchState;
+    found?: number;
+}
+
+export enum ESearchState {
+    running,
+    finished,
+    dropped,
+}
+
 export class State {
     public readonly subjects: Subjects<{
         active: Subject<IFilter | undefined>;
-        collection: Subject<FilterRequest[]>;
-        start: Subject<void>;
-        finish: Subject<number>;
-        showMatches: Subject<IFilter>;
+        start: Subject<IFilter[]>;
+        finish: Subject<IFinish>;
     }> = new Subjects({
         active: new Subject<IFilter | undefined>(),
-        collection: new Subject<FilterRequest[]>(),
-        start: new Subject<void>(),
-        finish: new Subject<number>(),
-        showMatches: new Subject<IFilter>(),
+        start: new Subject<IFilter[]>(),
+        finish: new Subject<IFinish>(),
     });
 
     private _search: Search;
     private _active: IFilter | undefined;
     private _hash: string | undefined;
+    private _searchInfo!: ISearchInfo;
 
     constructor(search: Search) {
         this._search = search;
@@ -40,22 +52,33 @@ export class State {
         return new Promise((resolve, reject) => {
             this._active = obj.clone(filter);
             this._hash = undefined;
+            if (this._searchInfo === undefined) {
+                this._searchInfo = {
+                    state: ESearchState.running,
+                };
+            } else {
+                this._searchInfo.state = ESearchState.running;
+            }
             this._search
                 .drop()
                 .then(() => {
-                    this.subjects.get().start.emit();
+                    this.subjects.get().start.emit([filter]);
                     this.subjects.get().active.emit(obj.clone(filter));
                     this._search
                         .search([filter])
                         .then((found: number) => {
-                            this.subjects.get().finish.emit(found);
+                            this.subjects.get().finish.emit({ found: found });
+                            this._searchInfo.found = found;
                             resolve(found);
                         })
                         .catch((err: Error) => {
                             this._active = undefined;
                             this.subjects.get().active.emit(undefined);
-                            this.subjects.get().finish.emit(0);
+                            this.subjects.get().finish.emit({ found: 0, error: err.message });
                             reject(err);
+                        })
+                        .finally(() => {
+                            this._searchInfo.state = ESearchState.finished;
                         });
                 })
                 .catch((err: Error) => {
@@ -87,29 +110,51 @@ export class State {
         }
         this._hash = hash;
         return new Promise((resolve, reject) => {
+            if (this._searchInfo === undefined) {
+                this._searchInfo = {
+                    state: ESearchState.running,
+                };
+            } else {
+                this._searchInfo.state = ESearchState.running;
+            }
             this._search
                 .drop()
                 .then(() => {
                     if (filters.length === 0) {
                         return resolve();
                     }
-                    this.subjects.get().start.emit();
+                    this.subjects.get().start.emit(filters);
                     this._search
                         .search(filters)
                         .then((found: number) => {
-                            this.subjects.get().finish.emit(found);
+                            this.subjects.get().finish.emit({ found: found });
+                            this._searchInfo.found = found;
                             resolve();
                         })
                         .catch((err: Error) => {
-                            this.subjects.get().finish.emit(0);
+                            this.subjects.get().finish.emit({ found: 0, error: err.message });
                             reject(err);
+                        })
+                        .finally(() => {
+                            this._searchInfo.state = ESearchState.finished;
                         });
                 })
                 .catch(reject);
         });
     }
 
+    public get searchInfo(): ISearchInfo | undefined {
+        return this._searchInfo;
+    }
+
     public reset(): Promise<void> {
+        if (this._searchInfo === undefined) {
+            this._searchInfo = {
+                state: ESearchState.dropped,
+            };
+        } else {
+            this._searchInfo.state = ESearchState.dropped;
+        }
         if (this._active === undefined) {
             return Promise.resolve();
         }
@@ -117,9 +162,5 @@ export class State {
         this._hash = undefined;
         this.subjects.get().active.emit(undefined);
         return this.filters();
-    }
-
-    public showMatches(filter: IFilter) {
-        this.subjects.get().showMatches.emit(filter);
     }
 }
--- a/application/client/src/app/service/session/dependencies/search/state.ts
+++ b/application/client/src/app/service/session/dependencies/search/state.ts
@@ -9,17 +9,6 @@ export interface IFinish {
     error?: string;
 }
 
-export interface ISearchInfo {
-    state: ESearchState;
-    found?: number;
-}
-
-export enum ESearchState {
-    running,
-    finished,
-    dropped,
-}
-
 export class State {
     public readonly subjects: Subjects<{
         active: Subject<IFilter | undefined>;
@@ -34,7 +23,7 @@ export class State {
     private _search: Search;
     private _active: IFilter | undefined;
     private _hash: string | undefined;
-    private _searchInfo!: ISearchInfo;
+    private _searching!: boolean;
 
     constructor(search: Search) {
         this._search = search;
@@ -52,13 +41,7 @@ export class State {
         return new Promise((resolve, reject) => {
             this._active = obj.clone(filter);
             this._hash = undefined;
-            if (this._searchInfo === undefined) {
-                this._searchInfo = {
-                    state: ESearchState.running,
-                };
-            } else {
-                this._searchInfo.state = ESearchState.running;
-            }
+            this._searching = true;
             this._search
                 .drop()
                 .then(() => {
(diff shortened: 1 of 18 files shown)


application/client/src/app/service/session/dependencies/search/state.ts
```diff
@@ -11,11 +11,13 @@ export class State {
         collection: Subject<FilterRequest[]>;
```
DmitryAstafyev, line 11: this event could be removed as soon as isn't used

application/client/src/app/service/session/dependencies/search/state.ts
```diff
@@ -11,11 +11,13 @@ export class State {
...
         start: Subject<void>;
         finish: Subject<number>;
+        showMatches: Subject<IFilter>;
```
DmitryAstafyev, line 14: this is not a pure event, that's action trigger.We can do things without it

application/client/src/app/service/session/dependencies/search/state.ts
```diff
@@ -116,4 +118,8 @@ export class State {
         this.subjects.get().active.emit(undefined);
         return this.filters();
     }
+
+    public showMatches(filter: IFilter) {
+        this.subjects.get().showMatches.emit(filter);
+    }
```
DmitryAstafyev, line 124: no need to have this method

DmitryAstafyev: and also this is logical mistake.
a) life-time of state = session
b) life-time of single listener of event `showMathces` is `app-views-search-input`-lifetime
if `app-views-search-input` doesn't exist, `showMathces` never will do any

application/client/src/app/ui/elements/transport/setup/udp/multicast/template.html
```diff
@@ -1,19 +1,21 @@
...
-    <input matInput placeholder="Network Interface" [(ngModel)]="multicast.interface" (ngModelChange)="_ng_onChanges()" [errorStateMatcher]="errors.interface">
-    <mat-error *ngIf="errors.interface.getErrorCode() === 'INVALID'">
-        Please define valid interface
-    </mat-error>
-    <mat-error *ngIf="errors.interface.getErrorCode() === 'REQUIRED'">
-        <strong>required</strong>
-    </mat-error>
-</mat-form-field>
-
+<div class="form-fields">
```
DmitryAstafyev, line 1: could we refuse of `<div class="form-fields">` as container? and use root-container `app-transport-udp-multicast`

application/client/src/app/ui/views/toolbar/search/input/component.ts
```diff
@@ -79,6 +80,14 @@ export class ViewSearchInput
                 this.input.focus();
             }),
         );
+        this.env().subscriber.register(
+            this.session.search
+                .state()
+                .subjects.get()
+                .showMatches.subscribe((filter: IFilter) => {
+                    this._updateSearch(filter);
+                }),
```
DmitryAstafyev, line 89: no needs to do it via events

application/client/src/app/ui/views/toolbar/search/input/component.ts
```diff
@@ -93,6 +102,9 @@ export class ViewSearchInput
                     this.active = new ActiveSearch(filter);
                     this.input.drop();
```
DmitryAstafyev, line 103: better setup `this.active` with event `finish` (event context should be changed to reflect results: `{ found: number; error?: string; filter: IFilter}`

application/client/src/app/ui/views/toolbar/search/input/component.ts
```diff
@@ -111,6 +123,10 @@ export class ViewSearchInput
         this.input.actions.recent.subscribe(() => {
             this.markChangesForCheck();
         });
+        this.input.actions.reaccept.subscribe(() => {
+            this._updateSearch(undefined);
+        });
```
DmitryAstafyev, line 128: in general with `accept` handler and `reaccept` happens same things - we are triggering a search. Do we need extra-event `reaccept` then?

application/client/src/app/ui/views/toolbar/search/input/input.ts
```diff
@@ -23,12 +23,14 @@ export class SearchInput {
...
         recent: Subject<void>;
         edit: Subject<void>;
+        reaccept: Subject<void>;
     } = {
         drop: new Subject(),
         clear: new Subject(),
         accept: new Subject(),
         recent: new Subject(),
         edit: new Subject(),
+        reaccept: new Subject(),
```
DmitryAstafyev, line 33: motivation of event `reaccept` isn't clean

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/search/input/input.ts
fur-gan: Was for updating search results once the flags are changed. The code has been adjusted and behavior for it is included into `accept` instead

application/client/src/app/ui/views/toolbar/search/input/component.ts
```diff
@@ -83,16 +89,30 @@ export class ViewSearchInput
...
+                    return;
+                }
+                filter = {
+                    filter: this.active.filter.filter,
+                    flags: this.input.flags,
+                };
+            } else {
+                filter = this.input.asFilter();
+                this.recent.update(filter.filter);
+            }
```
DmitryAstafyev, line 105: Okay, I see what you want to achieve here. But I think you've found a more fundamental mistake in the solution. `this.recent.update(filter.filter);` should not happen in component, it should happen in a place of search triggering, in `this.session.search.state().setActive(filter)`. Why? Because we can trigger search from different places - context menu, from here and others. 

Well, I would suggest leaving the event `accept` without changes, but move the saving of recent to the executor of search - it will put things into the right places.

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/search/input/component.ts
DmitryAstafyev, line 105: No need to add extra-flag to `accept` event. If you are changing a flags with active search, you will have `this.active !== undefined` it will give you understanding - no needs to save it as recent. Well
```
this.acrive === undefined && this.recent.update(filter.filter);
```
should work instead of additional flags, but check me please for sure.

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/search/input/input.ts
```diff
@@ -95,6 +96,27 @@ export class SearchInput {
...
+                this.flags.word = !this.flags.word;
+                this.actions.accept.emit();
+            },
+            regex: () => {
+                this.flags.reg = !this.flags.reg;
+                this.actions.accept.emit();
+            },
+        };
+    }
+
```
DmitryAstafyev, line 119: The general issue is: you are not tracking a state of active search. If you do not create a search (leave input empty) and switch flags - an empty search would be triggered. Reproducing:

1. Open any file
2. Do not write any in search-input, but just click on any flag-icon

Result: empty search is triggered.

Also from a code point of view, it still looks not accomplished. We have state of input `application/client/src/app/ui/views/toolbar/search/input/input.ts` and we have active search `application/client/src/app/ui/views/toolbar/search/input/active.ts`. I think if active-search was changed, we should change also `ActiveSearch`, probably with some kind of `update()` method, maybe. 

Probably solution could be addition event in `application/client/src/app/ui/views/toolbar/search/input/input.ts` - `flags`.

```
export class SearchInput {
    ...
    public flags: IFilterFlags = {
        word: false,
        cases: false,
        reg: true,
    };
    public actions: {
        ...
    } = {
        drop: new Subject(),
        clear: new Subject(),
        accept: new Subject(),
        recent: new Subject(),
        edit: new Subject(),
        flags: new Subject(),
    };

    ...

    // (!) This method can be refactored to have all setters in one place
    
    // public set(value: string | IFilter) {
    //     if (typeof value === 'string') {
    //         this.control.setValue(value);
    //         this._prev = value;
    //     } else {
    //         this.control.setValue(value.filter);
    //         this.flags = obj.clone(value.flags);
    //     }
    // }

    ...

    public set(): {
        value(value: string | IFilter): void;
        caseSensitive(): void;
        wholeWord(): void;
        regex(): void;
    } {
        return {
            value: (value: string | IFilter): void => {
                if (typeof value === 'string') {
                    this.control.setValue(value);
                    this._prev = value;
                } else {
                    this.control.setValue(value.filter);
                    this.flags = obj.clone(value.flags);
                }
            },
            caseSensitive: () => {
                this.flags.cases = !this.flags.cases;
                this.actions.flags.emit();
            },
            wholeWord: () => {
                this.flags.word = !this.flags.word;
                this.actions.flags.emit();
            },
            regex: () => {
                this.flags.reg = !this.flags.reg;
                this.actions.flags.emit();
            },
        };
    }
}
```
In this case in `component` you can safely listen it:
```
    public ngAfterViewInit(): void {
        ...
        this.input.actions.accept.subscribe(() => {
            const filter = this.input.asFilter();
            this.recent.update(filter.filter);
            this.session.search
                .state()
                .setActive(filter)
                .then(() => {
                    this.active = new ActiveSearch(filter);
                    this.input.drop();
                })
                .finally(() => {
                    this.markChangesForCheck();
                });
        });
        this.input.actions.flags.subscribe(() => {
            if (this.active === undefined) {
                return;
            }
            this.active.setFlags(this.input.flags);
            this.session.search
                .state()
                .setActive(this.active.filter)
                .finally(() => {
                    this.markChangesForCheck();
                });
        });
        ...
    }
```

Please, take into account (!) That's only suggestion of a solution, you can do another solution for sure.

Note (!). The code in my comment cannot be used as code of solution: it isn't checked, isn't tested and doesn't consider other goals of this PR. It's just an illustration of a possible solution.

DmitryAstafyev: While you are here, you can probably also take care restoring of search-progress. Situation is next: user started "heavy" search - we show progress; user left tab and back again - no search progress even it's still in progress. I guess we need a progress flag in search state.

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/search/input/component.ts
```diff
@@ -86,13 +106,55 @@ export class ViewSearchInput
...
+                .apply()
+                .catch((err: Error) => {
+                    this.log().error(`Fail to set flags on search search: ${err.message}`);
+                })
+                .finally(() => {
+                    this.markChangesForCheck();
+                });
+        });
+
+        this.input.actions.accept.subscribe(() => {
```
DmitryAstafyev, line 137: is this a mistake? You have 2 subscribers to `accept` (`this.input.actions.accept.subscribe`) on lines:
- 137
- 106

application/client/src/app/service/session/dependencies/search/state.ts
```diff
@@ -1,26 +1,40 @@
...
+
+export interface ISearchInfo {
+    state: ESearchState;
+    found?: number;
+}
+
+export enum ESearchState {
+    running,
+    finished,
+    dropped,
```
DmitryAstafyev, line 20: value `dropped` isn't used anyhow. Do we really need it? 
I think it would be enough just to have `boolean`: are we searching or not. There are just 2 states. No?

application/client/src/app/service/session/dependencies/search/state.ts
```diff
@@ -1,26 +1,40 @@
...
 
+export interface IFinish {
+    found: number;
+    error?: string;
+}
+
+export interface ISearchInfo {
+    state: ESearchState;
+    found?: number;
+}
```
DmitryAstafyev, line 15: Doesn't make sense to store in the state `found`, because the amount of found matches will be changes after the initial search. I think actual `found` you always can take from `map` (`session.search.map`), but check it please

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/search/input/component.ts
fur-gan: Yes, I messed up on the merge conflict as it seems

DmitryAstafyev (approved):

application/client/src/app/ui/views/toolbar/search/input/component.ts
```diff
@@ -51,6 +53,16 @@ export class ViewSearchInput
...
     public ngAfterContentInit(): void {
         this.progress = new Progress(this.session);
+        const searching: boolean | undefined = this.session.search.state().searching;
+        if (searching !== undefined) {
+            if (searching) {
+                this.progress.start();
+            } else {
+                this.progress.hidden = false;
+                this.progress.setFound(this.session.search.map.len());
+                this.detectChanges();
```
DmitryAstafyev, line 63: view isn't created yet (you are in `AfterContentInit`), you should not call detection of changes

application/client/src/app/ui/views/toolbar/search/input/component.ts
```diff
@@ -61,17 +73,21 @@ export class ViewSearchInput
             this.session.search
                 .state()
                 .subjects.get()
-                .start.subscribe(() => {
+                .start.subscribe((filters: IFilter[]) => {
                     this.progress.start();
+                    if (filters.length === 1) {
+                        this.active = new ActiveSearch(this.session.search, filters[0]);
+                    }
```
DmitryAstafyev, line 80: why we need it?

DmitryAstafyev:
will fix it by myself

application/client/src/app/service/session/dependencies/search/state.ts
```diff
@@ -1,26 +1,29 @@
...
-        start: new Subject<void>(),
-        finish: new Subject<number>(),
+        start: new Subject<IFilter[]>(),
+        finish: new Subject<IFinish>(),
     });
 
     private _search: Search;
     private _active: IFilter | undefined;
     private _hash: string | undefined;
+    private _searching!: boolean;
```
DmitryAstafyev, line 26: no need to have `undefined` as possible value... state is created as soon as session is created. On new session we don't have search in progress. Well, it should be just
```
private _searching: boolean = false;
```

application/client/src/app/ui/views/toolbar/search/input/component.ts
fur-gan: You're right. Didn't consider that. Should've moved it into `ngAfterViewInit` instead

application/client/src/app/ui/views/toolbar/search/input/component.ts
fur-gan: Do you mean the check or the whole thing?
The check is because we might have more than one element in the array. The assignment is because we create the active search there

application/client/src/app/service/session/dependencies/search/state.ts
fur-gan: Makes sense, yeah. That'd also get rid of a few extra lines in code.

Issue #1568, opened 2022-12-20
[chipmunk 3] Indicator for currently viewed area of file


fur-gan:
**Describe what you would like to have**

Add a colored transparent rectangle into the zoomable area of the chart view to indicate which part of the file is being viewed in the main view.

**More Information**

![image](https://user-images.githubusercontent.com/56543045/208615289-23f2d703-dbd8-4bbc-816f-3795b770bfdb.png)

PR #1569, opened 2022-12-21
Chart Features and fixes

fur-gan:
- #1568 
- Enhanced crosshair for better orientation in chart view

### DIFF
--- a/application/client/src/app/ui/views/toolbar/chart/chart/state.ts
+++ b/application/client/src/app/ui/views/toolbar/chart/chart/state.ts
@@ -6,6 +6,12 @@ import { IRange } from '@platform/types/range';
 import { AdvancedState, EChartName } from '../abstract/advanced';
 
 export class State extends AdvancedState {
+    public crosshairLeft: number = 0;
+    public tooltip: string = '';
+    public tooltipTop: number = 0;
+    public showLeftTooltip: boolean = true;
+    public mouseEnter: boolean = false;
+
     private _range!: IRange;
     private _loading: boolean = true;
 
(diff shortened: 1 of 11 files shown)


application/client/src/app/ui/views/toolbar/chart/chart/state.ts
```diff
@@ -39,8 +45,54 @@ export class State extends AdvancedState {
     }
 
     public onClick(event: MouseEvent) {
+        const position: number = this._calculatePosition(event);
+        position >= 0 && this._session.cursor.select(position, Owner.Chart);
+    }
+
+    public onMouseMove(event: MouseEvent) {
+        const rect: DOMRect = this._element.getBoundingClientRect();
```
DmitryAstafyev, line 53: `getBoundingClientRect` is a very-very expensive operation. That's not a good idea to do it with `mousemove` event.

application/client/src/app/ui/views/toolbar/chart/chart/state.ts
```diff
@@ -39,8 +45,54 @@ export class State extends AdvancedState {
...
+        const offsetX = event.offsetX;
+        if (offsetX > 0) {
+            this.crosshairLeft = offsetX;
+            this.showLeftTooltip = offsetX <= rect.width / 2;
+        }
+        if (event.offsetY > 0) {
+            this.tooltipTop = rect.height / 2;
+            const position: number = this._calculatePosition(event);
```
DmitryAstafyev, line 61: and inside `this._calculatePosition` one more call of `getBoundingClientRect`.
Be careful with this method it should be used as soon as it's really needed. As usual, we can guess a moment when the size could be changed and at this moment we can measure the size.

application/client/src/app/ui/views/toolbar/chart/chart/state.ts
```diff
@@ -39,8 +45,54 @@ export class State extends AdvancedState {
...
+            })
+            .catch((err: Error) => {
+                this._parent.log().error(err);
+            })
+            .finally(() => {
+                this._loading = false;
+            });
+    }
+
+    private _calculatePosition(event: MouseEvent): number {
```
DmitryAstafyev, line 93: In general, this method isn't trustable. Could it be NaN? or infinitive? If it's possible, I would use as output `number | Error` to make it more predictable

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/chart/chart/state.ts
fur-gan: I'm not sure at which place we could get something else than a number. I mean I make sure that we always calculate the position without failure, or did I miss something?

fur-gan: I see it now, yeah I've changed the code accordingly.

PR #1584, opened 2023-01-02
Fixes

fur-gan:
- Fix functionality for "Clear all" on Recent Actions
- Update list when elements deleted in `CTRL+SHIFT + P` view
- Fix chart view
- Refactoring and small fixes
- Allow canceling export of data

### DIFF
--- a/application/client/src/app/service/recent.ts
+++ b/application/client/src/app/service/recent.ts
@@ -19,7 +19,12 @@ const STORAGE_KEY = 'user_recent_actions';
 @DependOn(bridge)
 @SetupService(services['recent'])
 export class Service extends Implementation {
-    private _entryUpdate: Subject<void> = new Subject();
+    public readonly updated: Subject<void> = new Subject();
+
+    public override destroy(): Promise<void> {
+        this.updated.destroy();
+        return super.destroy();
+    }
 
     public get(): Promise<Action[]> {
         return new Promise((resolve, reject) => {
@@ -70,7 +75,7 @@ export class Service extends Implementation {
                 .entries({ key: STORAGE_KEY })
                 .delete(uuids)
                 .then(() => {
-                    this._entryUpdate.emit();
+                    this.updated.emit();
                     resolve();
                 })
                 .catch((err: Error) => {
@@ -116,10 +121,6 @@ export class Service extends Implementation {
                 };
             },
         };
-    }
-
-    public get entryUpdate(): Subject<void> {
-        return this._entryUpdate;
     }
 }
 export interface Service extends Interface {}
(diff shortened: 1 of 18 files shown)


application/client/src/app/ui/elements/recent.mini/state.ts
```diff
@@ -86,6 +86,17 @@ export class State extends Holder {
...
     }
 
+    public removeAll() {
+        recent
+            .get()
+            .then((actions: Action[]) => {
+                this.remove(actions.map((action: Action) => action.uuid));
+            })
+            .catch((err: Error) => {
+                console.error(`Fail to remove all recent actions: ${err.message}`);
```
DmitryAstafyev, line 96: ```this.log().error instead```

application/client/src/app/ui/elements/recent/state.ts
```diff
@@ -78,6 +78,21 @@ export class State extends Holder {
...
+                this.remove(actions.map((action: Action) => action.uuid));
+            })
+            .catch((err: Error) => {
+                console.error(`Fail to remove all recent actions: ${err.message}`);
+            });
+    }
+
+    public entryUpdate(): Subscription {
+        return recent.entryUpdate.subscribe(this.reload.bind(this));
+    }
```
DmitryAstafyev, line 94: no needs to share subscription with parent component, because you have `ilc` in state already
```ilc: IlcInterface & ChangesDetector```

File: application/client/src/app/ui/elements/recent/component.ts [hunk 3]
```diff
@@ -45,6 +45,7 @@ export class RecentActions extends ChangesDetector implements AfterContentInit {
                 this.detectChanges();
             }),
         );
+        this.env().subscriber.register(this.state.entryUpdate());
```
DmitryAstafyev, line 48: no need to have here (see below)

DmitryAstafyev (changes requested):

application/client/src/app/service/recent.ts
```diff
@@ -62,7 +70,18 @@ export class Service extends Implementation {
...
     public delete(uuids: string[]): Promise<void> {
-        return bridge.entries({ key: STORAGE_KEY }).delete(uuids);
+        return new Promise<void>((resolve, reject) => {
+            bridge
+                .entries({ key: STORAGE_KEY })
+                .delete(uuids)
+                .then(() => {
+                    this.updated.emit();
+                    resolve();
+                })
```
DmitryAstafyev, line 80: I cannot select line 69, but looks like `updated` event should happen not only remove, but on adding too. It make sense to emit event also with it. See line 69

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

PR #1666, opened 2023-01-28
Breadcrumbs

DmitryAstafyev:


application/apps/indexer/session/Cargo.toml
```diff
@@ -6,18 +6,22 @@ edition = "2021"
...
 serde_json = "1.0"
 sources = { path = "../sources" }
 thiserror = "1.0"
 tokio = { version = "1.24", features = ["full"] }
 tokio-stream = "0.1"
 tokio-util = "0.7"
-uuid = { version = "1.2", features = ["serde", "v4"] }
-dirs = "4.0"
\ No newline at end of file
+uuid = { version = "1.0", features = ["serde", "v4"] }
```
marcmo, line 24: why go back to an old version? current version should be `"1.3"`

application/apps/indexer/session/src/state/indexes/keys.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub fn add(&mut self, position: u64) {
+        self.keys.push(position);
+        self.sorted = false;
+    }
+
+    pub fn remove(&mut self, position: &u64) {
+        if let Ok(index) = self.keys.binary_search(position) {
+            self.keys.remove(index);
+        }
+    }
```
marcmo, line 26: not correct:
``` rust
mod test {
    #[test]
    fn test_keys() {
        let mut keys = super::Keys::new();
        assert_eq!(keys.first(), None);
        keys.add(13);
        keys.add(100);
        keys.add(12);
        keys.add(11);
        keys.add(200);
        keys.add(1);
        keys.remove(&100);
        let res = keys.clone();
        assert_eq!(res, vec![1, 11, 12, 13, 200]);
    }
}
```
will fail

application/apps/indexer/session/src/state/indexes/keys.rs
```diff
@@ -0,0 +1,91 @@
...
+        self.sort();
+        self.keys.first()
+    }
+
+    pub fn last(&mut self) -> Option<&u64> {
+        self.sort();
+        self.keys.last()
+    }
+
+    pub fn clone(&mut self) -> Vec<u64> {
```
marcmo, line 87: better not call a function `clone` if it is different from the expected clone

marcmo: (`clone` has a different signature)
``` rust
fn [clone](https://doc.rust-lang.org/std/clone/trait.Clone.html#tymethod.clone)(&self) -> Self;
```

application/apps/indexer/session/src/state/indexes/keys.rs
```diff
@@ -0,0 +1,91 @@
+use crate::events::{NativeError, NativeErrorKind};
+use indexer_base::progress::Severity;
+
+#[derive(Debug)]
+pub struct Keys {
```
marcmo, line 5: tests missing for a new data structure

application/apps/indexer/session/src/state/indexes/tests_controller.rs
```diff
@@ -0,0 +1,1783 @@
...
+                Action::AddBookmark(pos) => format!("AddBookmark - {pos}"),
+                Action::RemoveBookmark(pos) => format!("RemoveBookmark - {pos}"),
+                Action::Frame((range, _)) => format!("Frame - {range:?}"),
+            }
+        )
+    }
+}
+
+#[test]
+fn test() {
```
marcmo, line 158: much better implemented as `proptests`

application/apps/indexer/session/src/state/indexes/tests_map_cases.rs
```diff
@@ -0,0 +1,1729 @@
...
+                Action::Insert((pos, nature)) => format!("Insert - {nature:?} into {pos}"),
+                Action::Frame((range, _)) => format!("Frame - {range:?}"),
+            }
+        )
+    }
+}
+
+#[test]
+fn test() {
+    let mut tests: Vec<MapTest> = vec![
```
marcmo, line 170: better implemented as `proptests`

application/apps/indexer/session/src/state/indexes/tests_map_performance.rs
```diff
@@ -0,0 +1,106 @@
+use super::{map::Map, nature::Nature};
```
marcmo, line 1: performance tests must not be part of the unit tests. 
1. needs to be built as `release`
2. takes way too much time

application/apps/indexer/session/Cargo.toml
DmitryAstafyev: Corrected to 1.2. as most recent (1.2.2)

application/apps/indexer/session/src/state/indexes/keys.rs
DmitryAstafyev: yes, sure. But this test doesn't reflect real use-case and context. Our streams are increasing only. Well as for new keys we always will get it incrementally from 1 to 2. Exception - bookmarks (in these keys we do sorting for sure). But in anyway I've added extra-checks for sorted or not.

application/apps/indexer/session/src/state/indexes/tests_map_performance.rs
DmitryAstafyev: commented out

application/apps/indexer/session/src/state/indexes/keys.rs
DmitryAstafyev: Changed to `get()`

application/apps/indexer/session/src/state/indexes/keys.rs
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+    pub fn add(&mut self, position: u64) {
+        self.keys.push(position);
+        self.sorted = false;
+    }
+
+    pub fn remove(&mut self, position: &u64) {
+        if let Ok(index) = self.keys.binary_search(position) {
+            self.keys.remove(index);
```
marcmo, line 24: not sure why this should work...`binary_search` only works on sorted vector...when calling remove, the vector is not necessarily sorted.

application/apps/indexer/session/src/state/indexes/keys.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn test_keys_001() {
+        let mut keys = super::Keys::new();
+        assert_eq!(keys.first(), None);
+        keys.add(13);
+        keys.add(100);
+        keys.add(12);
+        keys.add(11);
+        keys.add(200);
+        keys.add(1);
+        keys.sort();
```
marcmo, line 105: that is cheating ;)

DmitryAstafyev: yes, it is ) But `keys` doesn't use as it is. It's used from `Map` and map cares about sorting. actually map always sort `keys` if it's needed. Well at the end, `keys` stays sorted always

application/apps/indexer/session/src/state/indexes/keys.rs
DmitryAstafyev: and `map` (on level upper) keeps `keys` always stored. That's why `Map` works fast - we are sorting as soon as we need it.

PR #1674, opened 2023-02-05
Up to 3.5.0

DmitryAstafyev:
- fix updating an indexed map on change of stream len
- remove breadcrumb separator from main view
- fix calculation of count of hidden rows

### DIFF
--- a/application/holder/package.json
+++ b/application/holder/package.json
@@ -1,6 +1,6 @@
 {
     "name": "chipmunk",
-    "version": "3.4.24",
+    "version": "3.5.0",
     "chipmunk": {
         "versions": {}
     },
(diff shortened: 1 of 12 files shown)


application/holder/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "chipmunk",
-    "version": "3.4.24",
+    "version": "3.5.0",
```
marcmo, line 3: let's stick to semantic versioning: increase last digit if no API was changed and if no major new features are introduced.
If you take a look at rust crates, they often have hight 2-digits in the last version number (e.g.1.0.78)

DmitryAstafyev: I've shifted major part because `breadcrumbs` looks like a new feature, but not patch

marcmo (approved):

PR #1681, opened 2023-02-13
Refactor filter validation

fur-gan:


application/client/src/app/service/session/dependencies/search/filters/request.ts
```diff
@@ -77,16 +77,41 @@ export class FilterRequest
     private _hash!: string;
     public found: number = 0;
 
-    static isValid(request: string | undefined): boolean {
-        if (request === undefined) {
-            return false;
-        }
-        if (!regex.isValid(request)) {
+    static isValid(
+        request: string | undefined,
```
DmitryAstafyev, line 81: `request` cannot be `undefined`. That's already error if someone tries to create RegEx from `undefined`

application/client/src/app/service/session/dependencies/search/filters/request.ts
```diff
@@ -77,16 +77,41 @@ export class FilterRequest
...
+            request === undefined ||
+            getFilterError(request, caseSensitive, wholeWord, regex) !== undefined
+        ) {
             return false;
         }
         return true;
     }
 
+    static isValidErrorMessage(
+        request: string | undefined,
```
DmitryAstafyev, line 96: `request` cannot be `undefined`. That's already error if someone tries to create RegEx from `undefined`

application/client/src/app/service/session/dependencies/search/filters/request.ts
```diff
@@ -77,16 +77,41 @@ export class FilterRequest
...
+        if (
+            request === undefined ||
+            getFilterError(request, caseSensitive, wholeWord, regex) !== undefined
+        ) {
             return false;
         }
         return true;
     }
 
+    static isValidErrorMessage(
```
DmitryAstafyev, line 95: would be better probably something like `getValidationError` because if we are using `is` it sound like it should return `bool`

application/client/src/app/ui/views/sidebar/search/filters/filter/component.ts
```diff
@@ -20,6 +20,12 @@ import { ChangesDetector } from '@ui/env/extentions/changes';
 import { State } from './state';
 import { stop } from '@ui/env/dom';
 
+export enum EFlag {
+    cases = 'cases',
+    word = 'word',
+    reg = 'reg',
+}
+
```
DmitryAstafyev, line 28: that's better to place to `@platform/types/filter`

application/client/src/app/service/session/dependencies/search/filters/request.ts
```diff
@@ -77,16 +77,41 @@ export class FilterRequest
     private _hash!: string;
     public found: number = 0;
 
-    static isValid(request: string | undefined): boolean {
-        if (request === undefined) {
-            return false;
-        }
-        if (!regex.isValid(request)) {
+    static isValid(
```
DmitryAstafyev, line 80: Can it get as an argument just `IFilter`? Almost everywhere you are passing `IFilter` into it

application/client/src/app/service/session/dependencies/search/filters/request.ts
DmitryAstafyev, line 95: Can it get as an argument just `IFilter`? Almost everywhere you are passing `IFilter` into it

application/client/src/app/ui/views/toolbar/search/input/component.ts
```diff
@@ -95,6 +97,16 @@ export class ViewSearchInput
...
                 }),
         );
+        this.env().subscriber.register(
+            this.session.search
+                .state()
+                .subjects.get()
+                .active.subscribe((filter: IFilter | undefined) => {
+                    if (filter !== undefined) {
+                        this.active = new ActiveSearch(this.session.search, filter);
+                    }
```
DmitryAstafyev, line 107: just don't know (didn't test) - do we need here `this.detectChanges()`?

application/client/src/app/ui/views/toolbar/search/input/component.ts
```diff
@@ -159,6 +157,7 @@ export class ViewSearchInput
             this.active = new ActiveSearch(this.session.search, active);
             this.input.drop();
         }
+        this._restoreNonActiveSearch();
```
DmitryAstafyev, line 160: the function is used just in one place... no sense to exclude `this.input.set().value(this.session.search.state().nonActive);` into standalone method

application/client/src/app/ui/views/toolbar/search/input/component.ts
```diff
@@ -203,5 +202,9 @@ export class ViewSearchInput
             },
         };
     }
+
+    private _restoreNonActiveSearch() {
+        this.input.set().value(this.session.search.state().nonActive);
+    }
```
DmitryAstafyev, line 208: no needs to create a method with 1 line of code just for calls from 1 place.

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/search/input/component.ts
fur-gan: I've recently read about how to write code more understandable and clean in which it is recommended to avoid writing comments and instead create functions with the function name describing what the action does. It might be obvious now what the function does, but at a later point storing it in a function might help us to understand the code easier. In my opinion, writing a function here has any downsides. But I can remove it if you still want, sure.

DmitryAstafyev: yes, it makes sense; self-explanation code is always good. But in this case, you should create a method `restore` and put there all, what related to restoring
```
protected restore() {
        const active = this.session.search.state().getActive();
        if (active !== undefined) {
            this.active = new ActiveSearch(this.session.search, active);
            this.input.drop();
        }
        this.input.set().value(this.session.search.state().nonActive);
}
```
and call it from `ngAfterViewInit`. In this case all looks good. Because current solution has retore-related stuff in `ngAfterViewInit` and in method `_restoreNonActiveSearch`. That's why there are 2 optional wayt:
- put all restore related to `ngAfterViewInit` (my first suggestion)
- exclude from `ngAfterViewInit` all restore related stuff to own method and call it from `ngAfterViewInit`

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/search/input/component.ts
fur-gan: I've tested it just now and it works without

DmitryAstafyev (approved):

PR #1687, opened 2023-02-22
Shells info extracting support

DmitryAstafyev:
## Features
- import envvars from available shells
- selecting shell 
- remember selected shell
## Fixes
- fix spawning of process (windows)
- fix About dialog
- add URLs in About dialog

### DIFF
--- a/application/apps/rustcore/rs-bindings/Cargo.toml
+++ b/application/apps/rustcore/rs-bindings/Cargo.toml
@@ -28,7 +28,7 @@ node-bindgen = {git = "https://github.com/infinyon/node-bindgen.git", branch = "
 # node-bindgen = { version = "5.0", features = ["serde-json"] }
 processor = { path = "../../indexer/processor" }
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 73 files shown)


application/apps/rustcore/rs-bindings/Cargo.toml
```diff
@@ -28,11 +28,12 @@ node-bindgen = {git = "https://github.com/infinyon/node-bindgen.git", branch = "
 # node-bindgen = { version = "5.0", features = ["serde-json"] }
 processor = { path = "../../indexer/processor" }
 serde = { version = "1.0", features = ["derive"] }
-serde_json = "1.0"
+serde_json = "^1.0"
```
marcmo, line 31: what does this? this is usually not used in cargo.toml files

DmitryAstafyev:
```
Caret requirements are an alternative syntax for the default strategy, ^1.2.3 is exactly equivalent to 1.2.3.
```
That's equal to `1.0` as far as I understand

PR #1689, opened 2023-03-01
Add saved chart request and more

fur-gan:
- Add data of saved chart requests
- Fixes concerning the chart in the toolbar
- etc.


DmitryAstafyev (changes requested):

application/client/src/app/service/session/dependencies/search/charts/store.ts
```diff
@@ -1,13 +1,22 @@
 import { Key, Store, StoredEntity } from '../store';
 import { ChartRequest } from './request';
 import { IFilter } from '@platform/types/filter';
-import { DisableConvertable } from '../disabled/converting';
-
+import { Subject } from '@platform/env/subscription';
+import { ISelectEvent } from '@ui/views/sidebar/search/providers/definitions/provider';
```
DmitryAstafyev, line 5: service referred to ui

application/client/src/app/service/session/dependencies/search/charts/request.ts
```diff
@@ -179,30 +184,66 @@ export class ChartRequest
...
+                    typeof desc.borderWidth === 'string' &&
+                    this.set(true).borderWidth(desc.borderWidth)
+                ) {
+                    event.on().borderWidth();
+                }
+                if (
+                    typeof desc.pointRadius === 'string' &&
+                    this.set(true).pointRadius(desc.pointRadius)
+                ) {
+                    event.on().pointRadius();
```
DmitryAstafyev, line 208: Why numeric props (`borderWidth`, `desc.pointRadius` etc.) are checked for `string` type? `typeof desc.borderWidth === 'string'`

application/client/src/app/service/session/dependencies/search/charts/store.ts
DmitryAstafyev, line 5: service should not refer to any UI component directly, it's break a logic of service (as independent on UI entity)

application/client/src/app/service/session/dependencies/search/charts/store.ts
```diff
@@ -18,20 +27,20 @@ export class ChartsStore extends Store<ChartRequest> {
...
     }
+
+    public select(event: ISelectEvent) {
+        this._selectedGuid = event.guids[0];
+        this.chartSelected.emit(event.entity === undefined ? '' : event.guids[0]);
+    }
+
+    public get selectedGuid(): string {
+        return this._selectedGuid;
+    }
```
DmitryAstafyev, line 45: I guess this is UI actions, but it's a `store`, it's just storing information about session's charts. These methods related to state of target UI compenent.

application/client/src/app/service/session/dependencies/search/charts/store.ts
```diff
@@ -1,13 +1,22 @@
...
 import { ChartRequest } from './request';
 import { IFilter } from '@platform/types/filter';
-import { DisableConvertable } from '../disabled/converting';
-
+import { Subject } from '@platform/env/subscription';
+import { ISelectEvent } from '@ui/views/sidebar/search/providers/definitions/provider';
 export { ChartRequest } from './request';
 
 export class ChartsStore extends Store<ChartRequest> {
+    public chartSelected: Subject<string> = new Subject<string>();
```
DmitryAstafyev, line 9: this is not an event of `Store` and entity. This is UI event. It should not "lives" here

application/client/src/app/service/session/dependencies/search/store.ts
```diff
@@ -75,6 +75,12 @@ export abstract class Store<T> extends Subscriber {
         return Array.from(this._entities.values());
     }
 
+    public getActive(): StoredEntity<T>[] {
+        return Array.from(this._entities.values()).filter((entity: T) => {
+            return (entity as any).definition.active;
```
DmitryAstafyev, line 80: This Store is an abstraction `Store<T>`, and type `T` also has requirements: 
```
export type StoredEntity<T> = T & Hash & Recognizable & Updatable<EntityUpdateEvent<any, T>>;
```
The idea behind is:  preventing any type related issues. It means, applying of `any` is in conflict with this conception.
Also `Store` just doesn't know what is `active` / `inactive`. 
If we should to teach `Store` recognize it, we should add requirements, like
```
interface ActiveState {
     isActive(): boolean;
}
export abstract class Store<T extends ActiveState> extends Subscriber {
```
But this is will not work, because `Store<T>` stays  behind all stores, including `disabled`, but disabled entities doesn't have property `active`/`inactive`
Well, it means - it should be removed from here.

application/client/src/app/ui/styles/material.less
```diff
@@ -297,6 +297,20 @@ mat-snack-bar-container {
     }
 }
 
+& mat-slider {
+    width: 96%!important;
+    margin-left: 0px!important;
```
DmitryAstafyev, line 302: Please use some class name for modifications of mat-components:
```
mat-slider.material-modification { ... }
```

DmitryAstafyev (changes requested):

application/client/src/app/service/session/dependencies/search/charts/request.ts
fur-gan: Can't find any specific reason for it, must've been a copy&paste mistake

application/client/src/app/ui/views/sidebar/search/filters/provider.ts
```diff
@@ -297,38 +319,37 @@ export class ProviderFilters extends Provider<FilterRequest> {
...
+                        .update([
+                            new FilterRequest({
+                                filter: {
+                                    filter: chartRequest.definition.filter,
+                                    flags: {
+                                        word: false,
+                                        cases: false,
+                                        reg: true,
+                                    },
+                                },
```
DmitryAstafyev, line 343: also static method on `FilterRequest` - `fromChart` will make code better

application/client/src/app/ui/views/sidebar/search/filters/provider.ts
```diff
@@ -297,38 +319,37 @@ export class ProviderFilters extends Provider<FilterRequest> {
...
-                //                         regexp: true,
-                //                     },
-                //                 },
-                //                 event.currentIndex,
-                //             );
-                //     }
+                this.session.search
+                    .store()
+                    .filters()
+                    .update([disabled.entity() as StoredEntity<FilterRequest>]);
```
DmitryAstafyev, line 325: we have method `tryRestore` to convert from `Disabled` to something

application/client/src/app/ui/views/sidebar/search/filters/provider.ts
```diff
@@ -297,38 +319,37 @@ export class ProviderFilters extends Provider<FilterRequest> {
...
+            } else {
+                if (data.entries !== undefined && data.entries.length > 0) {
+                    const entry: Entity<ChartRequest> = data.entries[0] as Entity<ChartRequest>;
+                    const chartRequest: ChartRequest = entry.extract();
+                    this.session.search.store().filters().delete([chartRequest.uuid()]);
+                    this.session.search
+                        .store()
+                        .filters()
+                        .update([
+                            new FilterRequest({
```
DmitryAstafyev, line 335: Update method gets `StoredEntity<T>`, but you are sending `FilterRequest`. That's mistake

application/client/src/app/ui/views/sidebar/search/filters/provider.ts
```diff
@@ -172,15 +175,34 @@ export class ProviderFilters extends Provider<FilterRequest> {
...
+            ChartRequest.isValid(request.definition.filter.filter) &&
+                items.push({
+                    caption: `Convert to Chart`,
+                    handler: () => {
+                        this.session.search.store().filters().delete([request.uuid()]);
+                        this.session.search
+                            .store()
+                            .charts()
+                            .update([
+                                new ChartRequest({
```
DmitryAstafyev, line 197: Update method gets `StoredEntity<T>`, but you are sending `ChartRequest`. That's mistake

application/client/src/app/ui/views/toolbar/chart/chart/component.ts
```diff
@@ -33,7 +33,7 @@ export class ViewChartCanvas extends ChangesDetector implements AfterViewInit, O
 
     public ngAfterViewInit() {
         this.state.bind(this, this.session, this.vcRef.element.nativeElement as HTMLElement);
-        this.state.init();
+        this.state.init(this.ilc());
```
DmitryAstafyev, line 36: with `bind` you already sent `this`, which includes `ilc()` as well. No needs to send `this.ilc()` again with `init`

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/chart/chart/chart.ts
```diff
@@ -0,0 +1,411 @@
...
+    public switchScaleType() {
+        (this._chart.options as any).scales['y'].type = this.reverseScaleType;
+        this._chart.update();
+    }
+
+    public zoom(zoomedRange: IRange) {
+        this._zoomedRange = zoomedRange;
+        [
+            (this._chart.config.options as any).scales['x'].min,
+            (this._chart.config.options as any).scales['x'].max,
```
DmitryAstafyev, line 79: is it possible to get rid of `any`? chartJS should have some type description, no?

application/client/src/app/ui/views/toolbar/chart/chart/chart.ts
```diff
@@ -0,0 +1,411 @@
...
+        this._chart.data.labels = [];
+        for (const [line, values] of this._session.search.values.get()) {
+            values.forEach((sValue: string, id: number) => {
+                const value: number = parseInt(sValue);
+                const scaleX = (this._chart.config.options as any).scales['x'];
+                if (line < scaleX.min) {
+                    scaleX.min = line;
+                } else if (line > scaleX.max) {
+                    scaleX.max = line;
+                }
```
DmitryAstafyev, line 210: is condition `line === scaleX`  missed?

application/client/src/app/ui/views/toolbar/chart/chart/chart.ts
```diff
@@ -0,0 +1,411 @@
...
+            );
+            if (entity === undefined) {
+                delete (this._chart.options as any).scales[axis];
+            } else {
+                if (axis === selectedUuid && entity.definition.active) {
+                    (this._chart.options as any).scales[axis].display = true;
+                    (this._chart.options as any).scales[axis].color;
+                }
+                (this._chart.options as any).scales[axis].display =
+                    axis === selectedUuid && entity.definition.active;
```
DmitryAstafyev, line 264: so many `any` :((

application/client/src/app/ui/views/toolbar/chart/chart/chart.ts
```diff
@@ -0,0 +1,411 @@
...
+            }
+        });
+    }
+
+    private _updateLabels(): [number, number] {
+        const labels = this._chart.config.data.labels as number[];
```
DmitryAstafyev, line 270: why you have to force type
```
....ata.labels as number[];
```
Isn't labels numbers?

application/client/src/app/ui/views/toolbar/chart/chart/state.ts
```diff
@@ -54,81 +90,30 @@ export class State extends AdvancedState {
...
-        let pos: IPosition = this._service.getPosition(this._session.uuid()).position;
-        if (!this._isPositionViable(pos)) {
-            pos = {
-                left: 0,
-                width: 0,
-                full: this._width,
-            };
-        }
+        let pos: IPosition | undefined = this._service.getPosition(this._session.uuid());
+        pos ??= this._defaultPosition;
```
DmitryAstafyev, line 114: can be done in 1 line
```
let pos: IPosition = this._service.getPosition(this._session.uuid()) ?? this._defaultPosition;
```

DmitryAstafyev: I've seen in somewhere else in code of PR

application/client/src/app/ui/views/toolbar/chart/chart/state.ts
```diff
@@ -138,60 +123,62 @@ export class State extends AdvancedState {
...
-                  full: this._width,
-                  left: 0,
-                  width: this._width,
-              };
-        const range: IRange = {
-            from: Math.round((position.left / position.full) * streamLen),
-            to: Math.round(((position.left + position.width) / position.full) * streamLen),
+        const streamLength: number = this._session.stream.len();
+        let position: IPosition | undefined = positionChange.position;
+        position ??= this._defaultPosition;
```
DmitryAstafyev, line 155:
```
        let position: IPosition = positionChange.position ?? this._defaultPosition;
```

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/toolbar/chart/chart/chart.ts
fur-gan: Technically it is, but the actual type is: `labels?: unknown[] | undefined`
The problem with ChartJS is, that almost everything is either abstract or optional, that all the checks just bloat the code.

fur-gan: In our case it's always `number[]`, but it could also be `string[]` or even `object[]`, you're free to choose. Since we always have `number[]`, I just cheated by casting it to `number[]` so we skip all the checks.

application/client/src/app/ui/views/toolbar/chart/chart/chart.ts
fur-gan: Well just for line 260 we'd have to write like this:
```
                    const options = this._chart.options;
                    if (options !== undefined) {
                        const scales = options.scales;
                        if (scales !== undefined) {
                            const scale = scales[axis];
                            if (scale !== undefined) {
                                scale.display = true;
                            }
                        }
```

application/client/src/app/ui/views/toolbar/chart/chart/chart.ts
fur-gan: No, if `line === scaleX.min` or `line  === scaleX.max`, we can just skip. This is because we don't need to update `min` or `max` in that case. What we do here is update the `min` and `max` of the scale so if for example: `max = 10`, 'there is no point to reassign `max` to `10` again.

PR #1762, opened 2023-04-09
Add contribution guide.

sudeeptarlekar:
Added contribution guide for new user.

### DIFF
--- a/contribution.md
+++ b/contribution.md
@@ -1,0 +1,148 @@
+# Contributing
+
+Chipmunk uses [Rust](https://www.rust-lang.org/) for processing the log files while frontend application is built using the [ElectronJS](https://www.electronjs.org/)
+
+## Pre-requisite
+
+To build and run chipmunk on local you will need following languages installed on your system.
+1. Rust
+2. NodeJS
+3. Ruby
+
+To check if you have all the pre-requisite installed or not, chipmunk provides the shell
+script for this purpose. After cloning the repo run following command in your preferred terminal.
+
+```
+sh developing/check.sh
+```
+
+If everything is installed then script should print success messages.
+If all pre-requisite are satisfied, [install](#installing-dependencies) dependencies.
+Follow steps below for installing the missing dependencies.
+
+### Installing Rust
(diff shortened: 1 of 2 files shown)


sudeeptarlekar:
@DmitryAstafyev can you please review.
cc: @marcmo 

Thanks in advance

contribution.md
```diff
@@ -0,0 +1,125 @@
...
+We are using [yarn](https://yarnpkg.com/) as a package manager. Installing yarn is simple
+as you just have to run following command in the terminal.
+
+```
+npm install -g yarn
+```
+
+## Build
+As mentioned earlier backend is built in Rust which takes care of reading the log files
+and frontend app is built using Electron JS.
```
DmitryAstafyev, line 77: I would add here, something like
### Completed build / rebuild
Completed build/rebuild could be done with 
```
rake developing:clean_rebuild_all
```
We are recommend use this command for the first time of building and also in cases when instance of chipmunk was updated/rebased.

contribution.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Changes in both backend and frontend
+```
+cd application/holder
+rake developing:holder_platform_bindings
+yarn run electron
+```
+
+With this you can take a look at the open issues and open a PR to fix them.
+
```
DmitryAstafyev, line 122: Probably make sense to add here something like:
### Build/developing commands
Full list of commands you can see with 
```
rake -T
```
Also you can use interactive mode with starting `rake` without any arguments
Peek 2023-04-10 10-09

DmitryAstafyev (changes requested):
In general make sense to add section about checking of code. Like "before creating of pull-request you should:":
- make sure, there are no issues related to `TypeScript`: `rake lint:all`. This command will check all `TypeScript` related parts of application
- make sure, there are no issues related to `Rust`: `rake clippy:all`. This command will check of `Rust` related parts of application
- make sure, basic tests are good with `rake test:all`

DmitryAstafyev (approved):

PR #1808, opened 2023-05-25
Fix DLT Fibex and Attachment index

kruss:


DmitryAstafyev (approved):

application/apps/indexer/parsers/src/dlt/attachment.rs
```diff
@@ -257,23 +256,23 @@ impl FtScanner {
                             size: ft_start.size as usize,
                             created_date: Some(ft_start.created),
                             modified_date: None,
-                            messages: vec![self.index],
+                            messages: vec![self.index - 1],
```
marcmo, line 259: better than first incrementing i and then indexing into `[i - 1]`
would be indexing into [i] and then incrementing in the end.

kruss: done.

Issue #1852, opened 2023-07-13
Chipmunk3 missing column colorisation feature


sudeeptarlekar:
**Prerequisites**

affected version: 3

make sure to check this before filing an issue:

- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

**Describe what you would like to have**

In chipmunk2 user can add a color to particuar column but this feature is missing from Chipmunk3
![Screenshot 2023-07-13 at 2 09 34 PM](https://github.com/esrlabs/chipmunk/assets/23317691/f9525e3f-f3e0-4f83-85cd-31b1f561ec02)

DmitryAstafyev:
consider please also #1628

PR #1854, opened 2023-07-14
Add ability to change column colors in DLT file

sudeeptarlekar:
Add column color change functionality and ability to show or hide columns from DLT file only.

Resolves: #1852


application/client/src/app/ui/views/sidebar/settings/modules.ts
```diff
@@ -0,0 +1,11 @@
+import { NgModule } from "@angular/core";
+import { CommonModule } from "@angular/common";
+
+@NgModule({
+    imports: [
+        CommonModule
+    ],
+    declarations: [],
```
sudeeptarlekar, line 8: declarations: [Settings],
exports: [Settings]

Issue #1833, opened 2023-06-16
Add export/import filters to menu


marcmo:
also add it to the context menu for filters
<img width="227" alt="image" src="https://github.com/esrlabs/chipmunk/assets/99541/d3ab10aa-24de-4a79-8e75-cafe963d5cf6">

DmitryAstafyev:
@sudeeptarlekar the general solutions would be:
- into context-menu (calls on any filter/chart) introduce item - `Export into File`
- by selecting it - opens dialog to select destination file
- by selecting of file - save all filters/charts/disabled into given file

PR #1856, opened 2023-07-19
Add import-export options for sidebar filters

sudeeptarlekar:
Add ability to import and export all filters, Charts from sidebar. At the moment user can still export all filters from history in toolbar but to enhance the accessibility this PR add same feature in sidebar.

Fixes: #1833

### DIFF
--- a/application/client/src/app/ui/views/sidebar/search/providers/providers.ts
+++ b/application/client/src/app/ui/views/sidebar/search/providers/providers.ts
@@ -503,19 +503,22 @@ export class Providers {
             }
         });
 
-        this.contextMenuOptions(event.items);
+        this.injectGeneralMenuItems(event.items);
 
         this.subjects.get().context.emit(event);
+
     }
 
 
-    public contextMenuOptions(items: IMenuItem[]): IMenuItem[] {
+    public injectGeneralMenuItems(items: IMenuItem[]): void {
         const historySession = history.get(this.session);
         if(historySession === undefined) {
             this.logger.error('History session is not defined');
-            return items;
+            return;
         }
-        items.push({ /* Delimiter */ });
+        items.push({
+            /* Delimiter */
+        });
         const store = this.session.search.store();
         const showExport: boolean = (store.filters().get().length + store.charts().get().length + store.disabled().get().length) !== 0;
         showExport && items.push(
@@ -528,10 +531,20 @@ export class Providers {
             caption: 'Import from File',
             handler: () => this.filters(historySession).import(),
         });
-        return items;
     }
 
     protected filters(historySession: HistorySession): { import(): void; export(): void } {
+        const logAndNotifyError = (error: Error): void => {
+            this.logger.error(error.message);
+            notifications.notify(
+                new Notification({
+                    message: error.message,
+                    session: this.session.uuid(),
+                    actions: []
+                })
+            );
+        };
+
         return {
             import: (): void => {
                 bridge.files().select.text()
@@ -558,9 +571,9 @@ export class Providers {
                             historySession.apply(collection);
                         }
                     })
-                    .catch(error => this.logAndNotifyError(error));
+                    .catch(error => logAndNotifyError(error));
                 })
-                .catch(error => this.logAndNotifyError(error))
+                .catch(error => logAndNotifyError(error))
             },
             export: (): void => {
                 bridge.files().select.save()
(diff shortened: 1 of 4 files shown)


application/client/src/app/ui/views/sidebar/search/providers/providers.ts
```diff
@@ -500,6 +506,67 @@ export class Providers {
...
+                    return;
+                }
+                if (uuids.length > 1) {
+                    this.session.switch().toolbar.presets();
+                    return;
+                } else {
+                    const collection = history.collections.get(uuids[0]);
+                    if (collection === undefined) {
+                        this.logger.error(`Cannot find imported collection with UUID: ${uuids[0]}`);
+                        return
```
DmitryAstafyev, line 562: missed `;`

```
return;
```

application/client/src/app/ui/views/sidebar/search/providers/providers.ts
```diff
@@ -500,6 +506,67 @@ export class Providers {
         this.subjects.get().context.emit(event);
     }
 
+
+    public contextMenuOptions(items: IMenuItem[]): IMenuItem[] {
+        const historySession = history.get(this.session);
+        if(historySession === undefined) {
+            this.logger.error('History session is not defined');
+            return items
```
DmitryAstafyev, line 514: missed `;`

```
return items;
```

application/client/src/app/ui/views/sidebar/search/providers/providers.ts
```diff
@@ -485,6 +488,9 @@ export class Providers {
                     });
                 },
             });
+
+        this.contextMenuOptions(event.items);
+
```
DmitryAstafyev, line 493: It would be better to put at the end, to show menu items (related to export/import) and the end of context-menu.

application/client/src/app/ui/views/sidebar/search/providers/providers.ts
```diff
@@ -500,6 +506,67 @@ export class Providers {
...
+        });
+        return items;
+    }
+
+    private _exportFilters = (historySession: HistorySession) => {
+        bridge.files().select.save()
+        .then((filename: string | undefined) => {
+            if (filename === undefined)
+                return;
+            history.export([historySession.collections.uuid], filename);
```
DmitryAstafyev, line 537: `history.export(...)` is a promise; we always should handle `catch` of promises.

application/client/src/app/ui/views/sidebar/search/providers/providers.ts
```diff
@@ -500,6 +506,67 @@ export class Providers {
...
+        });
+        items.push(
+        {
+            caption: 'Import from File',
+            handler: () => this._importFilterFile(historySession),
+        });
+        return items;
+    }
+
+    private _exportFilters = (historySession: HistorySession) => {
```
DmitryAstafyev, line 532: I would group methods `_exportFilters` and `_importFilterFile` into something like:
```
protected history(historySession: HistorySession): { import(): void; export(): void } => {
    return {
        export: (): void => {...},
        import: (): void => {...},
    };
}
```

application/client/src/app/ui/views/sidebar/search/providers/providers.ts
```diff
@@ -500,6 +506,67 @@ export class Providers {
...
+        bridge.files().select.save()
+        .then((filename: string | undefined) => {
+            if (filename === undefined)
+                return;
+            history.export([historySession.collections.uuid], filename);
+        })
+        .catch(error => this.logger.error(error.message));
+    }
+
+    private _importFilterFile = (historySession: HistorySession) => {
```
DmitryAstafyev, line 542: please use regular declaraction of methods:

```
private _importFilterFile(historySession: HistorySession): void {
    ....
}
```
or 
```
protected importFilterFile(historySession: HistorySession): void {
    ....
}
```

this `private _importFilterFile = (historySession: HistorySession) => {}` declaration of property (with function as value), but not declaration of method

But see also comment above, I would group methods `_importFilterFile` and `_exportFilters`

DmitryAstafyev (changes requested):

application/client/src/app/ui/views/sidebar/search/providers/providers.ts
sudeeptarlekar: But then wouldn't it create confusion with https://github.com/esrlabs/chipmunk/blob/master/application/client/src/app/service/history.ts#L75?

sudeeptarlekar: How aboout `protected filters....`

DmitryAstafyev:
In general also in case of an error (import/export) we should inform the user about it. Search via solution be `new Notification(` to find necessary API

application/client/src/app/ui/views/sidebar/search/providers/providers.ts
```diff
@@ -497,9 +502,90 @@ export class Providers {
...
+        items.push(
+        {
+            caption: 'Import from File',
+            handler: () => this.filters(historySession).import(),
+        });
+        return items;
+    }
+
+    protected filters(historySession: HistorySession): { import(): void; export(): void } {
+        return {
```
DmitryAstafyev, line 535: method `logAndNotifyError` can be present as arrow-function here (as soon as it's used only in this scope)
```
const logAndNotifyError = () => {...};
return {...}
```

DmitryAstafyev (approved):

PR #1874, opened 2023-07-27
Fix attaching files after concat (close #1866)

### LINKED ISSUE #1866: Attach new text file after concat operation on text files
## To reproduce:
- concat a couple of text files
- open "observing" tab in sidebar
- select "Attach New"
- select text file

## Result
- new file is visible in "observing" tab in tail state
- but in main view - no content of new file

## Expected result
- new file should give new content into main view

DmitryAstafyev:
- Allows attaching to same session text & binary files after concat operation was done
- Fires error on attempt to attach a file to session with linked session file

### DIFF
--- a/application/apps/indexer/session/src/state/session_file.rs
+++ b/application/apps/indexer/session/src/state/session_file.rs
@@ -47,20 +47,32 @@ pub enum SessionFileState {
 }
 
 #[derive(Debug, Clone)]
-pub enum SessionFileStage {
-    NotCreated,
-    Linked,
-    NotLinked,
+pub enum SessionFileOrigin {
+    Linked(PathBuf),
+    Generated(PathBuf),
+}
+
+impl SessionFileOrigin {
+    pub fn filename(&self) -> PathBuf {
+        match self {
+            Self::Linked(filename) => filename,
+            Self::Generated(filename) => filename,
+        }
+        .clone()
+    }
+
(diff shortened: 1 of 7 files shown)


application/apps/indexer/session/src/state/session_file.rs
```diff
@@ -46,13 +46,21 @@ pub enum SessionFileState {
     NoChanges,
 }
 
+#[derive(Debug, Clone)]
+pub enum SessionFileStage {
+    NotCreated,
+    Linked,
+    NotLinked,
+}
+
```
marcmo, line 55: one possibility. but makes matching a little complex.
other possibility could be:
``` rust
#[derive(Debug)]
pub struct SessionFile {
    pub grabber: Option<Box<Grabber>>,
    pub filename: Option<PathBuf>,
    pub writer: Option<BufWriter<File>>,
    pub last_message_timestamp: Instant,
    pub sources: SourceIDs,
    pub origin: Option<SessionFileOrigin>,
}
...
pub enum SessionFileOrigin {
    TextFile(PathBuf),
    Streams(streams),
}
```

marcmo: looks good

DmitryAstafyev: yes, make sense... I've updated it

PR #1875, opened 2023-07-28
Consider file type with export as raw (close #1867)

### LINKED ISSUE #1867: Exporting content as raw from PcapNg file
## Reproduce:
- select PcapNg file with DLT and open it
- select a couple of rows and call (via context menu) - "Export Rows as Raw"
- save into file

## Result
- exported file should be DLT binary or PcapNg with binary, but it isn't possible to open file with exported data not as binary, not as PcapNg

## Expected result
- file with exported data should be readable
- file with exported data should fit to original file format

DmitryAstafyev:
- Creates suitable source for exporting as raw
- Allows export SomeIp into binary from PcapNg
- Fixes issue with exporting DLT from PcapNg


marcmo:
looks good

PR #1883, opened 2023-08-04
[DRAFT] Add someip statistics api to unbound session

kruss:


application/apps/rustcore/ts-bindings/spec/defaults.json
```diff
@@ -58,7 +58,7 @@
         },
         "jobs": {
             "regular": {
-                "execute_only": [1,2,3,4,5],
```
marcmo, line 61: why did this change? does it mean we are not executing those?

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -337,6 +348,250 @@ impl Display for SomeipLogMessage {
...
+type StatisticDetailMap = FxHashMap<usize, usize>;
+
+fn add_statistic(item_map: &mut StatisticItemMap, item_id: usize, detail_id: usize) {
+    if let Some((item_value, detail_map)) = item_map.get_mut(&item_id) {
+        *item_value += 1;
+        if let Some(detail_value) = detail_map.get_mut(&detail_id) {
+            *detail_value += 1;
+        } else {
+            detail_map.insert(detail_id, 1);
+        }
```
marcmo, line 517: this is usually done with the `entry` function (https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry)
if possible use that here as well

AmmarAbouZor:
Feature is implemented in Chipmunk 4

PR #1885, opened 2023-08-07
Bug fixing 01

DmitryAstafyev:


application/apps/indexer/session/src/state/indexes/keys.rs
```diff
@@ -15,6 +15,28 @@ impl Keys {
...
+            }
+            to = *value;
+        }
+        if (!ranges.is_empty() && ranges[ranges.len() - 1].start() != &from)
+            || (ranges.is_empty() && !self.keys.is_empty())
+        {
+            ranges.push(RangeInclusive::new(from, to));
+        }
+        ranges
+    }
```
marcmo, line 38: AFAICT this is the only functionality added in the PR. but would be good to add a unit test for it.

Issue #1841, opened 2023-07-04
File type recognize


DmitryAstafyev:
On rustcore level we should have a method, which allows determining a basic type of file: binary or text. This method will be used to suggest a user suitable parser during opening the file.

DmitryAstafyev:
@marcmo I thing it also can include functionality by detecting the coding of text files. What do you think?

Issue #1921, opened 2023-10-06
Support other than UTF8 coding


DmitryAstafyev:
Add support other than UTF8 coding

ddrm86:
As commented on https://www.reddit.com/r/rust/comments/1tln8fd/comment/onxlk61/ at my job we have .log files (plain text with ANSI encoding, probably Windows-1252 or ISO-8859-1, generated by some old Apache L4J version) that are available through HTTPS.

I can easily dump these files on a Windows terminal with "curl -N https_address", but it does not work with Chipmunk 4.0.0-alpha.2:

```
[2026-05-26T09:38:30Z ERROR session::handlers::observing] Producer Error: Data Source Error: Unrecoverable source error: Unable to decode input as UTF8
Unrecoverable error during producer session: Data Source Error: Unrecoverable source error: Unable to decode input as UTF8

```
The attached file is a small sample of one of those logs.

[expeco-APPSEC_server1.log.2026-04-22.txt](https://github.com/user-attachments/files/28258706/expeco-APPSEC_server1.log.2026-04-22.txt)

AmmarAbouZor:
@ddrm86 Thanks for the sample. I'll try to look into it soon once the needed features for Chipmunk new version 4 are ported from Chipmunk 3.

Meanwhile it's possible to have a plugin parser as workaround until we try to resolve this issue as native behavior

PR #1935, opened 2023-10-18
Detect file type

sudeeptarlekar:
Rustcore should have functionality to detect the file type. If file type is binary or text, and if text what is the encoding of file.

Resolves: #1841


DmitryAstafyev:
## Location

We have a nice suitable place for it `application/apps/indexer/addon`, but now it's just one module. I would suggest (if @marcmo agree):
- rename `application/apps/indexer/addon` into `application/apps/indexer/addons`
- current lib in `application/apps/indexer/addon` rename into `dlt-tools` and add as submodule
- same for your tool, create `files-tools` and add there

It will be something like:
```
-- addons
---- dlt-tools
------ src
------ Cargo.tolm
---- files-tools
------ src
------ Cargo.tolm
```
Create it as `lib` and add tests of course.

## Artifacts for tests

Also, take into account, we have this place `developing/resources`. Here you can store resources for testing. You can left there some example files, which you would use for testing.

## General idea

I think reading some parts of target-file and searching "specific binary or not binary" symbols is a good idea and the right direction.  Just a question, would it be simple just try some chunks of file with [std::str::from_utf8](https://doc.rust-lang.org/std/str/fn.from_utf8.html#). Because if it isn't UTF8: you will get error or result string `len` will be `!=` to the `len` of given byte's slice. That's just an idea, I don't know what is a better.

application/apps/indexer/indexer_cli/src/main.rs
```diff
@@ -1799,3 +1799,60 @@ async fn detect_messages_type(input: &Path) -> Result<bool, DltParseError> {
...
+    };
+    // println!("{:?}", buffer);
+    let result: bool = is_file_binary(buffer);
+
+    return String::from("Still in progress...");
+}
+
+fn fetch_starting_chunk(file_path: &Path) -> Result<Vec<u8>>{
+    let file = File::open(file_path)?;
+    let mut file = file.take(1024);
```
DmitryAstafyev, line 1822: In general, any static/configuration parameters keep on a top as const. It helps to easy manipulate with such kind of parameters 

```
const BYTES_TO_READ: usize = 1024;
...
let mut file = file.take(BYTES_TO_READ);
```

DmitryAstafyev (changes requested):

DmitryAstafyev:
`DO NOT MERGE` in the name of PR scares me )) sounds like a bomb in our solution)) If you PR in the `draft` state it will not be merged - no worries )

DmitryAstafyev:
linked to: #1921, #1841

marcmo:
`indexer/addon` is a good place...no need to rename to `addons`

application/apps/indexer/indexer_cli/src/main.rs
```diff
@@ -1799,3 +1799,60 @@ async fn detect_messages_type(input: &Path) -> Result<bool, DltParseError> {
...
+    let mut buffer = Vec::new();
```
marcmo, line 1823: for such a small amount of data you can put the data on the stack which is much faster.
```
let mut buffer = [0; BYTES_TO_READ];
```

application/apps/indexer/indexer_cli/src/main.rs
marcmo: better read in more then just 1k...better for the heuristics. maybe 10k

application/apps/indexer/addons/file-tools/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+        String::from("The file is likely binary with content\n") + content
+    } else {
+        String::from("The file is not likely binary with content\n") + content
+    }
+}
+
+fn fetch_starting_chunk(file_path: &Path) -> Result<Vec<u8>> {
+    let file = File::open(file_path)?;
+    let mut file = file.take(BYTES_TO_READ.try_into().unwrap());
+    let mut buffer = Vec::new();
```
DmitryAstafyev, line 44: Unwrap is dangerous, it will panic in case of error. Never panic in libs
Also what will happen is `file.size < BYTES_TO_READ`?

application/apps/indexer/addons/file-tools/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+use std::{fs::File, io::Read, path::Path};
+
+const BYTES_TO_READ: usize = 10240;
+
+pub fn detect_file_type(file_path: &Path) -> String {
+    let chunks = fetch_starting_chunk(file_path);
+    let buffer = match chunks {
+        Ok(buffer) => buffer,
+        // TODO: add correct error handling
+        Err(error) => panic!("Error while reading chunks from file: {}", error),
```
DmitryAstafyev, line 16: No panic please in libs )

application/apps/indexer/addons/file-tools/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+// read the first chunks from the file and return String
+// 'binary' or 'text' as of now. Later we can add enum for file
+// type.
+
+use anyhow::Result;
+use std::{fs::File, io::Read, path::Path};
+
+const BYTES_TO_READ: usize = 10240;
+
+pub fn detect_file_type(file_path: &Path) -> String {
```
DmitryAstafyev, line 11: Any IO related operations returns should return some `Result<>`. It's better then panic

DmitryAstafyev, line 11: Actually you are not detecting type of file... It's a little confuse, because by type we understand as usual `executable, image, video etc.`... I would suggest to have as public methods `is_text()` and `is_binary()`

application/apps/indexer/addons/file-tools/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    let buffer_length: f32 = buffer.len() as f32;
+
+    let non_printable_chars_count = buffer.iter().filter(|x| !is_printable(x)).count() as f32;
+    let non_ascii_ratio = non_printable_chars_count / buffer_length;
+
+    let high_ascii_count = buffer.iter().filter(|x| !is_high_ascii(x)).count() as f32;
+    let high_ascii_ratio = high_ascii_count / buffer_length;
+
+    (non_ascii_ratio > 0.3 && high_ascii_ratio < 0.05)
+        || (non_ascii_ratio > 0.8 && high_ascii_ratio > 0.8)
```
DmitryAstafyev, line 71: 0.3, 0.05 and etc, what is it about :/? Maybe better to put it into named constants to give an explanation of what is it?

application/apps/indexer/addons/file-tools/src/lib.rs
sudeeptarlekar: I will add error handling, and yes I added the code to check the file size and take that into consideration, but somehow when I read entire file, result of the function is wrong and when I go back to `1024` I am getting correct result.

application/apps/indexer/addons/file-tools/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    let file_length: u64 = metadata(file_path)?.len() - 1;
+    let file_length = if BYTES_TO_READ < file_length {
+        BYTES_TO_READ
+    } else {
+        file_length
+    };
+
+    let mut file = file.take(file_length);
+    let mut buffer = Vec::new();
+    file.read_to_end(&mut buffer)?;
```
DmitryAstafyev, line 36: not sure about this method... from doc:

> Read all bytes until EOF in this source, placing them into buf.

we don't need to read all bytes, just chuck with len = BYTES_TO_READ

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

application/apps/indexer/addons/file-tools/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    Ok(buffer)
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
```
DmitryAstafyev, line 42: Please add tests for error/specific use cases:
- the wrong path is given (file doesn't exist)
- file has a size 0 bytes

application/apps/indexer/session/src/unbound/commands/file.rs
```diff
@@ -0,0 +1,10 @@
+use super::{CommandOutcome, CommandOutcome::Finished};
+use crate::events::{ComputationError, ComputationError::OperationNotSupported};
+use file_tools::is_binary;
+
+pub fn is_file_binary(file_path: String) -> Result<CommandOutcome<bool>, ComputationError> {
+    match is_binary(file_path) {
+        Ok(is_binary) => Ok(Finished(is_binary)),
+        Err(err) => Err(OperationNotSupported(err.to_string())),
+    }
```
DmitryAstafyev, line 9: Can be written with mapping, for example:
```
func_call(few).map(|res| process success).map_err(|e| converting error)
```

application/client/src/app/ui/elements/navigator/providers/provider.files.ts
```diff
@@ -15,6 +15,7 @@ const DEFAULT_LEN = 20000;
 export class Provider extends Base<IFileDescription> {
     protected count: number = 0;
     protected roots: string[] = [];
+    protected logger = scope.getLogger('ProviderLogger');
```
DmitryAstafyev, line 18: We don't need extra logger. Logger comes with embedded `ilc`:
```
this.ilc.log().warn(...);
this.ilc.log().debug(...);
```

application/client/src/app/ui/elements/navigator/providers/provider.files.ts
```diff
@@ -139,32 +140,44 @@ export class Provider extends Base<IFileDescription> {
...
+                                        .guessParser()
+                                        .get(),
+                                )
+                                .catch((err: Error) => {
+                                    this.ilc.log().error(`Fail to open text file; error: ${err.message}`);
+                                });
+                        }
+                    })
+                    .catch((error) => {
+                        this.logger.error(error);
```
DmitryAstafyev, line 177: Use embedded logger: `this.ilc.log()`

application/holder/src/service/bridge/file/file.ts
```diff
@@ -19,15 +19,20 @@ export const handler = Requests.InjectLogger<
...
+                            }
+                            resolve(
+                                new Requests.File.File.Response({
+                                    files
+                                })
+                            )
+                        })
+                        .catch(error => {
+                            reject(error);
+                        });
```
DmitryAstafyev, line 35: can be just `.catch(reject);`

application/holder/src/service/bridge/file/select.ts
```diff
@@ -4,22 +4,32 @@ import { electron } from '@service/electron';
 import { File } from 'platform/types/files';
 import { FileType } from 'platform/types/observe/types/file';
 import { getFileEntities } from '@env/fs';
+import { scope } from "platform/env/scope";
 
 import * as Requests from 'platform/ipc/request';
 
 function any(ext?: string): Promise<File[]> {
+    const logger = scope.getLogger('SelectFile');
```
DmitryAstafyev, line 12: Creating of logger is not much, but still costs something... let's create in `catch` section. I hope it will be used not very often ;)

application/holder/src/service/bridge/file/select.ts
```diff
@@ -4,22 +4,32 @@ import { electron } from '@service/electron';
...
+                    if (entities instanceof Error) {
+                        reject(entities);
+                    } else {
+                        resolve(entities);
+                    }
+                })
+                // TODO: Add error log
+                .catch((error) => {
+                    logger.error(`Error while opening a file; ${error}`);
+                    throw error;
```
DmitryAstafyev, line 30: It will create an exception... let's reject here `catch(reject)`... probably logger will not be needed

application/holder/src/service/bridge/file/select.ts
```diff
@@ -32,12 +42,14 @@ function dlt(): Promise<File[]> {
...
-                    resolve(entities);
-                }
+                getFileEntities(files)
+                    .then((entities) => {
+                        if (entities instanceof Error) {
+                            reject(entities);
+                        } else {
+                            resolve(entities);
+                        }
+                    });
```
DmitryAstafyev, line 52: catch section is missed

application/holder/src/service/bridge/file/select.ts
```diff
@@ -50,12 +62,14 @@ function pcapng(): Promise<File[]> {
...
```
DmitryAstafyev, line 72: catch section is missed

application/holder/src/service/bridge/file/select.ts
```diff
@@ -68,12 +82,14 @@ function pcap(): Promise<File[]> {
...
```
DmitryAstafyev, line 92: catch section is missed

application/holder/src/service/bridge/folder/select.ts
```diff
@@ -19,12 +19,14 @@ function any(ext?: string): Promise<File[]> {
...
-                    resolve(entities);
-                }
+                getFileEntities(files)
+                    .then((entities) => {
+                        if (entities instanceof Error) {
+                            reject(entities);
+                        } else {
+                            resolve(entities);
+                        }
+                    });
```
DmitryAstafyev, line 29: catch section is missed

application/holder/src/service/bridge/folder/select.ts
```diff
@@ -34,12 +36,14 @@ function dlt(): Promise<File[]> {
...
```
DmitryAstafyev, line 46: catch section is missed

application/holder/src/service/bridge/folder/select.ts
```diff
@@ -49,12 +53,14 @@ function pcapng(): Promise<File[]> {
...
```
DmitryAstafyev, line 63: catch section is missed

application/holder/src/service/bridge/folder/select.ts
```diff
@@ -64,12 +70,14 @@ function pcap(): Promise<File[]> {
...
```
DmitryAstafyev, line 80: catch section is missed

application/holder/src/service/cli/open.ts
```diff
@@ -34,7 +34,7 @@ export class Action extends CLIAction {
         return this.error;
     }
 
-    public execute(cli: Service): Promise<void> {
+    public async execute(cli: Service): Promise<void> {
```
DmitryAstafyev, line 37: I don't remember (please check it), but if `fucntion` is async, we can return just `Error`, instead `Promise.reject(Error)`... but please check it.

DmitryAstafyev (changes requested):

application/apps/indexer/addons/file-tools/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+    } else {
+        file_length
+    };
+
+    let mut file = file.take(file_length);
+    let mut buffer = Vec::new();
+    file.read_to_end(&mut buffer)?;
+    Ok(buffer)
+}
+
```
DmitryAstafyev, line 44: This is right. And this is work. I do not mind if this block will no be changed, but it still can be visually optimized:
```
fn fetch_starting_chunk(file_path: &Path) -> Result<Vec<u8>> {
    let bytes_to_read: u64 = (metadata(file_path)?.len().max(1) - 1).min(BYTES_TO_READ);
    let mut buffer = Vec::new();
    File::open(file_path)?
        .take(bytes_to_read)
        .read_to_end(&mut buffer)?;
    Ok(buffer)
}
```

application/apps/rustcore/ts-bindings/src/native/native.jobs.ts
```diff
@@ -11,6 +11,8 @@ export abstract class JobsNative {
 
     public abstract destroy(): Promise<void>;
 
+    public abstract isFileBinary(sequence: number, filePath: string): Promise<void>;
+
```
DmitryAstafyev, line 15: Why `Promise<void>;`? But not `Promise<boolean>;`

application/client/src/app/ui/elements/navigator/providers/provider.files.ts
```diff
@@ -139,32 +138,44 @@ export class Provider extends Base<IFileDescription> {
...
+                                        .get(),
+                                )
+                                .catch((err: Error) => {
+                                    this.ilc.log().error(`Fail to open text file; error: ${err.message}`);
+                                });
+                        }
+                    })
+                    .catch((error) => {
+                        this.ilc.log().error(error);
+                        throw error;
```
DmitryAstafyev, line 176: Why `throw`? What motivation is?

application/holder/src/env/fs/index.ts
```diff
@@ -74,27 +76,27 @@ export function getFolders(paths: string[]): string[] | Error {
...
+                    throw entity
+                }
+                entities.push(entity);
+            }
+            return entities.filter((f) => f !== undefined) as File[];
         } catch (e) {
             return new Error(error(e));
         }
     }
 }
```
DmitryAstafyev, line 97: Section `try {} catch {}` there are from the previous solution (sync solution). As soon as now it's becoming `async` code, at least return type should be changed.

We cannot return `Promise<File[] | Error>` error in promise traditionally means rejecting. I think it should be `Promise<File[]>`.

This piece of code 
```
                const entity: File | Error | undefined = await getFileEntity(files[i])
                if (entity instanceof Error) {
                    throw entity
                }
```
would not work, because in case of rejecting (error) of `getFileEntity` it will throw exception, because no catch section.

application/holder/src/env/fs/index.ts
```diff
@@ -74,27 +76,27 @@ export function getFolders(paths: string[]): string[] | Error {
...
 
-export function getFileEntity(filename: string): File | undefined | Error {
+export async function getFileEntity(filename: string): Promise<File | undefined | Error> {
```
DmitryAstafyev, line 99: as soon as it becomes `async` we cannot resolve promise with errors. 
`Promise<File | undefined | Error>` this is should be `Promise<File | undefined>` and error will go to rejection

application/holder/src/env/fs/index.ts
```diff
@@ -74,27 +76,27 @@ export function getFolders(paths: string[]): string[] | Error {
...
     try {
         const stat = fs.statSync(filename);
```
DmitryAstafyev, line 101: Please make sure statements `try/catch` are still actual after switching to `async` code

application/holder/src/service/cli/open.ts
DmitryAstafyev: Here is a good example to better understand promises and `async` stuff. 
```
async function a() {
    return new Error('Test');
}

async function b() {
    return Promise.reject(new Error('Test'));
}

async function c() {
    throw new Error('Test');
}

async function d() {
    return new Promise(() => {
        throw new Error('Test');
    })
}

a().then((_) => {
    console.error(`[A]: CATCH: isn't handled`);
}).catch((_) => {
    console.log('[A] - ok');
});

b().then((_) => {
    console.error(`[B]: CATCH: isn't handled`);
}).catch((_) => {
    console.log('[B] - ok');
});

c().then((_) => {
    console.error(`[C]: CATCH: isn't handled`);
}).catch((_) => {
    console.log('[C] - ok');
});

d().then((_) => {
    console.error(`[D]: CATCH: isn't handled`);
}).catch((_) => {
    console.log('[D] - ok');
});
```

console:
```
[A]: CATCH: isn't handled
[C] - ok
[B] - ok
[D] - ok
```
Well returning `new Error` will not reject promise, but return result with `resolve`

application/holder/src/service/cli/open.ts
```diff
@@ -34,25 +34,23 @@ export class Action extends CLIAction {
...
```
DmitryAstafyev, line 37: Please also make sure, all places of this function's usage are checked and corrected also. It's quite radical change (sync -> async) and we should make sure - all calls of changed functions are corrected too.

DmitryAstafyev (changes requested):

application/apps/indexer/addons/file-tools/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+    let buffer = match chunks {
+        Ok(buffer) => buffer,
+        Err(err) => return Err(err),
+    };
+
+    let result = from_utf8(&buffer);
+    match result {
+        Ok(_file_content) => Ok(false),
+        Err(_err) => Ok(true),
+    }
```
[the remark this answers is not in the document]
sudeeptarlekar: I think `map`, `map_err` won't work here, as I was reading the [docs](https://doc.rust-lang.org/std/result/enum.Result.html#method.map) and map takes generic argument and returns that argument in Result, but here our argument is `&str` and we are returning `bool`.

But I found `map_or` which we can use...

DmitryAstafyev (approved):

Issue #1966, opened 2023-11-22
Create a chipmunk cli tool for developers


marcmo:
chipmunk has evolved to be more than a product, it's a platform + a specific use case.
we have long used rake for all developer related tasks.
especially building and re-building chipmunk is challenging.
we are a rust + typescript project and do not want to rely on ruby/rake on the developer side.
so we want to create a tool similar to the node-bindgen cli (https://github.com/infinyon/node-bindgen) that takes care of those tasks and replaces rake.
the tool should be implemented as a rust application

AmmarAbouZor:
This is a great idea!
Would it be specific for chipmunk or for rust + typescript apps in general? if it's gonna have a general use I would suggest creating a separate repo for it so we can publish it to `crates.io` when it's done

AmmarAbouZor:
I think this one is already done. Nice to see first issue I worked on again :grinning:

PR #1985, opened 2024-02-17
CLI Tool for Build System

AmmarAbouZor:
This PR closes the issue #1966.
This PR continue the great work in the PR  #1967.

At the end of this drafts we should have a working CLI tool to build, lint, clean and test the project to replace the current solution using rake

Here is the road map for this PR:
- [x] Support build, lint, clean and test commands
- [x] Providing an option to report all the build logs to stdout or to the given file
- [x] Test the tool across platforms 
- [x] Find a better way to get the root repository path. (Using Git?)
- [x] Complete missing test commands
- [x] Install the command as a cargo sub command
- [x] Shell-completion
- [x] Add check for development dependencies as a command and this check must always run before all other commands too
- [x] Rebuild only the changed parts from the last run (using Checksum?)
- [x] Keep track on the running task to prevent duplicated and parallel runs to the same task
- [x] Add Run command that builds the necessary parts then runs the app
- [x] Replace the async trait `Manager` with the enum `Target`
- [x] Provide a command to print the needed environment tools to build the app besides checking if they are installed
- [x] Provide a command to print a `print_dots` to visualize the dependencies graph
- [x] Add the `Update` target (Build, Clean...)
- [x] Resolve tasks tree upfront and show all of them in progress bar, keeping the execution running concurrently
- [x] Apply new changes after switching to yarn 4


DmitryAstafyev:
Originally created [PR](https://github.com/esrlabs/chipmunk/pull/1967) with useful discussion

AmmarAbouZor:
@DmitryAstafyev I was working on the prettifying the error messages and wanted to eliminate the app panic when it starts outside of chipmunk repo and replace it with an error message, which lead me to use OnceCell from Tokio to achieve that. After implementing it for `LOCATION`, I've implemented it for `TRACKER` as well to remove the dependency `Lazy-Statics` since the same functionality is provided by tokio

AmmarAbouZor:
I ran into a strange error while implementing the test command for TS bindings, where I need some help. Here are the error infos and what I've found out after some debugging and investigations:

Running the test for the spec `jobs` leads to the program receiving `SIGTRAP` because of " Fatal Javascript out of memory" from the operating system even if we run the spec `jobs` only.  In the most cases the program get terminated without any error but in some cases I've got the following error:
```bash
Test Wrapper:
<--- Last few GCs --->
[65036:0x3d58002fc000]       82 ms: Mark-Compact (reduce) 0.8 (3.2) -> 0.8 (2.0) MB, 9.49 / 0.00 ms  (average mu = 0.222, current mu = 0.064) last resort; GC in old space requested
[65036:0x3d58002fc000]       86 ms: Mark-Compact (reduce) 0.8 (2.0) -> 0.8 (2.0) MB, 3.55 / 0.00 ms  (average mu = 0.163, current mu = 0.005) last resort; GC in old space requested
<--- JS stacktrace --->
#
# Fatal JavaScript out of memory: CALL_AND_RETRY_LAST
#
~/chipmunk/application/apps/rustcore/ts-bindings/node_modules/electron/dist/electron exited with signal SIGTRAP 
```

After debugging and investigations I've found the responsible lines of codes for this error. There are two positions worth further investigation:
- We use `Stdio::piped()` when executing the commands in the file `spawner.rs` line 74. But if we use `Stdio::inherit()` we don't get this error, but the message will be printed on the console.
- The error is produced by calling the method `jobs.getShellProfiles()` in the file `session.jobs.spec.ts` line 173. When this line is commented out the command runs in the CLI app with `Stdio::pipe()` without any problem.
- Currently I've commented the spec `jobs' out and wrote the information about this problem in a TODO inside the app in the file `wrapper.rs` line 122.

I think the problem is with method `jobs.getShellProfiles()` from typescript having problem to run inside a `Stdio::piped()` in rust (with tokio runtime).

Since my experience with typescript is very basic I would appreciate it if a typescript expert can have a look on it. @DmitryAstafyev @marcmo

DmitryAstafyev:
@AmmarAbouZor but the jobs test works with the current build sys based on ruby. 
```
rake bindings:test:jobs
```

AmmarAbouZor:
@DmitryAstafyev Sorry I didn't mention that I've had the same observation while I was investigating in it. It works with Rust too if we set the stdout in Rust command to `Stdio::inherit()` instead `Stdio::piped()` in  the file`spawner.rs` line 74.

The problem happens only if the typescript function runs inside a piped stdout in Rust runtime.

AmmarAbouZor:
I've finally implemented a work-around for the problem with Stdio with TS bindings tests.
- I've created a new method in the tracker that suspends the progress bars and run the giving blocking command.
- All TS command run sequentially with Stdio::inherit() printing their outputs direct to the command line while the bars are suspended.
- After the commands are finished the progress bar will be shown again.

AmmarAbouZor:
The CLI tool can now be installed and invoked as cargo extension or sub-command
To install it we can use `cargo install --path=cli`
And after the installation we can invoke the command via cargo `cargo chipmunk {args}`

AmmarAbouZor:
Builds will be tracked in a shared map to prevent duplications

When a build is called the shared map will be checked and it will proceed as following:
  - If the build already done then return early with its result
  - If the build is running then wait for it to be done then return its result
  - If the build hasn't started then start it and update the build states map

cli/src/modules/mod.rs
```diff
@@ -0,0 +1,347 @@
...
+                    let res = self.after(prod).await?;
+                    if let Some(result) = res {
+                        results.push(result);
+                    }
+                    if matches!(self.kind(), Kind::Ts) && prod {
+                        let clean_res = self.clean().await?;
+                        results.push(clean_res);
+                        let install_res = self.install(prod).await?;
+                        results.push(install_res);
+                    }
```
AmmarAbouZor, line 249: @DmitryAstafyev 
I was comparing the current ruby implementation and I couldn't find the equivalent of this code there, and I want to double check if we need it here.

AmmarAbouZor: Hello @DmitryAstafyev Thanks for the clarification. 
I've noticed that this code will be called after calling `self.after()` which would actually do the manual copying. Do we need to move it to above the `self.after()` block?

AmmarAbouZor:
- I made a large refactoring to replace the trait `Manager` with the enum `Target` to improve the structure and the readability of the code.
- This refactor solved the problem with the duplicated install calls to install TS bindings tools between TS and RS bindings
- The crate `async-trait` has been removed since it's not used anymore
- I had to solve the problem that (recursive async calls aren't supported in rust currently) with Boxed Futures

AmmarAbouZor:
I've tested the tool on Windows and made some fixes to make work there and it's working.
I think this PR is ready for review now since it has all the needed functionalities. I've marked at as ready for review and I'll continue working on the auto completion while it's being reviewed

AmmarAbouZor:
Shell Completion has been added

cli/src/modules/mod.rs
AmmarAbouZor: I moved that part to before copying the data in order to get the benefits from it

marcmo (changes requested):
Hi Ammar,
your PR is quite impressive, I'm very happy with your first PR!
But I still think we should implement a different algorithm for executing the individual build/test/... steps than the one dmitry suggested.
Currently, we introduce parallism only in 1 place, where we join all the futures to build the individual targets.
We also have to recursively call our build function and thus need to resort to boxing the future which is not nice.
I'd rather suggest the following algorithm:
starting from the target(s) we need to build (more general: the tasks we need to execute), we initially build up the dependency tree in a hash-map.
once we have this tree, we implement a function that computes only the tasks that do not have any more dependencies.
this set of tasks we push to a task-queue from where a task-executer can pick them up and execute concurrently. As soon a task is finished, we re-calculate the "ready" tasks and add them to our build queue.
That way we can
* know in advance how many tasks need to be performed in total
* can always run as many tasks in parallel as is appropriate
* avoid any kind of recursion and boxed futures
* all the code we need to execute in parallel becomes infrastructure and will not depend on the kind of task we will performe (build/test/lint/...)
We can discuss this later today, but that's the way I think we should go.

For the rest of the code it's already pretty good! Just as a rule of thumb: we avoid any unhandled panic (unwraps/expects) and rather deal with them as an error that will be handled all the way to the top.
About documentation: we try to have at least every public API documented

AmmarAbouZor:
I've started with the re-write to calculate the jobs upfront and I marked this PR as draft again until the re-write is done

AmmarAbouZor:
Following steps in the re-write to app logic has been done: 
- Jobs dependencies will be resolved upfront which make it possible to write unit-tests for them too
- All jobs will be viewed on the status bar from the start with the new stat `wait` for the pending jobs that are waiting for their dependencies
- Jobs will still run concurrently
- Linting TS targets pulls build jobs for all involved targets in their dependencies because TS linting doesn't check for type correctness which is only possible with build. This was solved before by calling build command from linting function which led to errors when linting is called after clean.
- New print dot CLI argument has been added to print all the tasks alongside with the current print-dot command that prints the targets only
- Missing paths for clean command has been added.
- When an install job fails, their corresponding build and copy jobs will be automatically skipped but the other jobs will still run
- Yarn 4 Integration form master are already built within this PR

AmmarAbouZor:
Changes tests on Windows at it seems it's working fine

AmmarAbouZor:
Hi @DmitryAstafyev
Here is a list of all included features that needed to be tested on this PR:
- All commands (build, clean, test, lint, run) should run cross-platform.
- Checksum computation to figure out if something has changed on some target, then all the targets who are dependent on it must be built too.
- Build functions will be called in the linting for TypeScript and the Checksums + Dependencies build will be applied here as well. 
- Print dot function to print dependencies between the main targets and their tasks as well.
- Shell completion.

AmmarAbouZor:
@DmitryAstafyev @marcmo 
- I've created integration tests for all provided commands on the CLI tool + Tests for the checksum implementation for detect code changes and for the dependencies between build targets.
- I've also fixed unit-tests from wasm-bindens that weren't working on Windows and added Clippy fixes for Windows.
- The integration tests are tested on Linux and Windows only

AmmarAbouZor:
@DmitryAstafyev @marcmo 
I've integrated the CLI tool in a separate GitHub Actions for reference now. 
These steps are just for testing for now and would need more improvements once we want to replace the current ones with them

AmmarAbouZor:
Hi @marcmo, 
As we had spoken yesterday, the following changes have been added:
- `mathces!()` macros and catch-all place holders in match statements have been replaced with explicit types, this will keep the compiler assistance powerful when we add more targets or job types in the future
- Boolean in Checksum comparison map has been replaced with an enum to enforce logic correctness with rust type system
- While making the changes I've found out that we have been running build steps for all test targets which is needed for typescript and WASM targets only. This has been fixed as well.

cli/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+console = "0.15.7"
+fs_extra = "1.3.0"
+futures = "0.3.28"
+futures-lite = "1.13.0"
+git2 = "0.18.2"
+indicatif = "0.17.7"
+tokio = { version = "1.36.0", features = ["full"] }
+dir_checksum = { path = "./dir_checksum" }
+which = "6.0"
+clap_complete = "4.5"
```
marcmo, line 28: looks good, just make sure that for stable crates we leave out the bugfix part of the version.
tokio: do we need full here?

cli/dir_checksum/src/input.rs
```diff
@@ -0,0 +1,108 @@
...
+            // fills one buffer while the worker threads are hashing the other
+            // one. We might implement that in the future, but since this is
+            // the slow path anyway, it's not high priority.
+            Self::File(file) => {
+                copy_wide(file, &mut hasher)?;
+            }
+        }
+        Ok(hasher.finalize())
+    }
+}
```
marcmo, line 46: super nice❤️

cli/dir_checksum/tests/integration_tests.rs
marcmo, line 1: why do we have an
`tests/integration_tests.rs` file and a folder
`integration_tests` ?

cli/dir_checksum/tests/integration_tests.rs
```diff
@@ -0,0 +1,178 @@
...
+fn hash_combinations_file_identical() -> anyhow::Result<()> {
+    let (tmp_dir, _) = create_tmp_dir_with_file("comb_file_identical")?;
+
+    let original_hash = calc_combined_checksum(tmp_dir.path())?;
+
+    assert_eq!(
+        original_hash,
+        calc_combined_checksum(tmp_dir.path())?,
+        "Hashes of the same directory must be identical"
+    );
```
marcmo, line 29: I don't get this test. Just calling the function twice and comparing the results does not seem very interesting.

cli/dir_checksum/tests/integration_tests.rs
```diff
@@ -0,0 +1,178 @@
...
+#[test]
+fn hash_combinations_empty_file() -> anyhow::Result<()> {
+    let (tmp_dir, _) = create_tmp_dir_with_file("comb_empty_file")?;
+
+    let original_hash = calc_combined_checksum(tmp_dir.path())?;
+
+    // Create an empty file
+    let empty_file_path = tmp_dir.path().join("empty.txt");
+    let empty_file = File::create(&empty_file_path)?;
+    drop(empty_file);
```
marcmo, line 85: isn't that the same as just not assigning the result to a local variable?

cli/src/checksum_records.rs
```diff
@@ -0,0 +1,266 @@
...
+use anyhow::{anyhow, Context};
+use dir_checksum::{calc_combined_checksum, HashDigest};
+
+use crate::{job_type::JobType, location::get_root, target::Target};
+
+const FILE_NAME_DEV: &str = ".build_chksum_dev";
+const FILE_NAME_PROD: &str = ".build_chksum_prod";
+
+#[derive(Debug)]
+pub struct ChecksumRecords {
```
marcmo, line 18: pub struct missing documentation

marcmo (commented):
very impressive PR Ammar! Good job!!
it's huge so we should try to merge it ASAP and possibly create issues on top of it

cli/dir_checksum/tests/integration_tests.rs
AmmarAbouZor: I renamed the file to `mod.rs` to follow the conventions in rust for structuring integration tests.
Now it's `{root}/tests/mod.rs`

cli/dir_checksum/tests/integration_tests.rs
AmmarAbouZor: The point of this test, that if we call the function on the same dir without any changes then I should provide the same hash. But I can remove it since it's included in the next test `hash_combinations_add_then_remove_file()`

application/apps/rustcore/ts-bindings/spec/session.jobs.spec.ts
```diff
@@ -133,6 +133,7 @@ describe('Jobs', function () {
     it(config.regular.list[4], function () {
         return runner(config.regular, 4, async (logger, done, collector) => {
             const jobs = collector(await Jobs.create()) as Jobs;
+            //TODO: If we comment out these lines, build CLI with Stdio::piped() works.
```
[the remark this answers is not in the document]
AmmarAbouZor: This todo is related to the issue described in `cli/src/target/wrapper.rs`, which is that TS bindings tests errors when run them with a newly spawned piped stdout and only run when we inherit the stdout from the main process.
For now, we are suspending the progress bars temporally while running these tests printing out their output directly to stdout and then we continue with the progress bars. 
I think the main issue is in the crate [envvars](https://crates.io/crates/envvars) when we run it from piped shell.

I'll remove the comment here and update the comments inside `wrapper.rs` instead

cli/src/target/mod.rs
```diff
@@ -0,0 +1,551 @@
...
+            Target::Wrapper => ["application", "apps", "rustcore", "ts-bindings"].iter(),
+            Target::Client => ["application", "client"].iter(),
+            Target::Shared => ["application", "platform"].iter(),
+            Target::App => ["application", "holder"].iter(),
+            Target::Cli => ["cli"].iter(),
+            Target::Wasm => ["application", "apps", "rustcore", "wasm-bindings"].iter(),
+            Target::Updater => ["application", "apps", "precompiled", "updater"].iter(),
+        };
+
+        sub_parts.collect()
```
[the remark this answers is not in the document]
AmmarAbouZor: This could be implemented in the future in a separate PR to reduce the hard coded configuration and to read them form a separate configuration file

AmmarAbouZor:
@marcmo Requested changes are implemented/answered. The PR can be reviewed again :+1:

marcmo (approved):

Issue #1715, opened 2023-03-07
Searcher (rustcore)


DmitryAstafyev:
## Specification
- takes a list of files (text-plain) and regular expression
- make search and returns results (filename, count of matches)
- operation should be cancelable

PR #1986, opened 2024-02-18
Implementation for text_grep - Issue (Searcher (rustcore) #1715)

itsmesamster:


application/apps/text_grep/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+                let sender = sender.clone();
+                let error_sender_clone = error_sender.clone();
+                let cancel_token = cancel_token_clone.clone();
+                let file_path = Arc::clone(file_path);
+                thread::spawn(move || {
+                    if let Err(err) =
+                        process_file(&file_path, &patterns, chunk_size, &cancel_token, &sender)
+                    {
+                        if error_sender_clone.send(err.to_string()).is_err() {
+                            eprintln!("Error sending error message through channel");
```
DmitryAstafyev, line 63: crate should not post any to console, it should return error in case of error

application/apps/text_grep/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+
+        let thread_handles: Vec<_> = file_paths_arc
+            .par_iter()
+            .map(|file_path| {
+                let patterns = patterns_arc.clone();
+                let sender = sender.clone();
+                let error_sender_clone = error_sender.clone();
+                let cancel_token = cancel_token_clone.clone();
+                let file_path = Arc::clone(file_path);
+                thread::spawn(move || {
```
DmitryAstafyev, line 58: why we need to spawn system thread with async method? In general it's very expensive. If we really need threads here tokio's `task` should be used. But I don't think we need it at all.

application/apps/text_grep/src/wc_1_ahocorasick.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    Ok(results)
+  }
+}
+
+fn process_file(
+  file_path: &Arc<str>,
+  patterns: &[Arc<str>],
+  chunk_size: usize,
+  cancel_token: &CancellationToken,
```
DmitryAstafyev, line 93: Using `Arc` potentially will give more headaches than cloning.

application/apps/text_grep/src/wc_1_ahocorasick.rs
```diff
@@ -0,0 +1,165 @@
...
+  sender: &mpsc::Sender<Result<SearchResult, String>>,
+) -> Result<(), String> {
```
DmitryAstafyev, line 95: Crate should have own Error enum. Please take a look, for example, to crate `thiserror`, which helps to create it. See examples in other places of chipmunk's solution.

application/apps/text_grep/src/wc_1_ahocorasick.rs
```diff
@@ -0,0 +1,165 @@
...
+  let file_path = PathBuf::from(&**file_path);
+  if !is_text_file(&file_path) {
+    let error_msg = format!("File '{}' is not a text file", file_path.display());
+    if sender.send(Err(error_msg.clone())).is_err() {
+      eprintln!("Error sending search result through channel");
```
DmitryAstafyev, line 100: no output into terminal from crate

application/apps/text_grep/src/wc_1_ahocorasick.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+    return Ok(());
+  }
+
+  let mut pattern_counts = HashMap::new();
+  let start_time = Instant::now();
+  eprintln!("Trying to process file {:?}", file_path);
```
DmitryAstafyev, line 107: no output into terminal from crate

application/apps/text_grep/src/wc_1_ahocorasick.rs
```diff
@@ -0,0 +1,165 @@
...
+    .map(|pattern| {
+      aho_corasick::AhoCorasickBuilder::new()
+        .match_kind(aho_corasick::MatchKind::Standard)
+        .ascii_case_insensitive(true)
+        .build(vec![&**pattern])
+    })
+    .collect::<Result<_, _>>()
+    .map_err(|e| e.to_string())?;
+
+  eprintln!("Trying to Open file {:?}", file_path);
```
DmitryAstafyev, line 120: no output into terminal from crate

application/apps/text_grep/src/wc_1_ahocorasick.rs
```diff
@@ -0,0 +1,165 @@
...
+  let file = File::open(&file_path).map_err(|e| e.to_string())?;
+  let mut reader = BufReader::new(file);
+  let mut buffer = vec![0; chunk_size];
+
+  loop {
+    if cancel_token.is_cancelled() {
+      return Err("Operation cancelled".to_string());
```
DmitryAstafyev, line 127: I don't think in case of cancellation, we should return error... Probably `Result<Option<...>, E>` with `Ok(None)` in case of cancellation would be better

application/apps/text_grep/src/wc_reviewed.rs
```diff
@@ -0,0 +1,154 @@
...
+            if error_sender_clone.send(err.to_string()).is_err() {
+              eprintln!("Error sending error message through channel");
+            }
+          }
+        })
+      })
+      .collect();
+
+    for handle in thread_handles {
+      handle.join().unwrap();
```
DmitryAstafyev, line 64: `unwrap()` is acceptable only in tests and very-very specific use-cases. Please, check for error here and return some error if it is. For ex: `handle.join().map_err(|e| e.into())?;`

DmitryAstafyev (changes requested):

DmitryAstafyev:
@itsmesamster thanks for PR. I have a couple of comments in general:

- do not mix `async` and stuff from `std`. I'm about `treads`, `channels` etc from `std`... If you need to spawn something it should be really motivated and we should use `tokio` to do it (because it's major framework for chipmunk)
- make sure you do not post any into terminal... It's crate/lib - it should return status. We can post something only for debugging  
- please define `Error` enum for errors of your crate. Take a look to `thiserror` crate, which can be useful for string converting
- please add some information about measurements...

itsmesamster:
image

Some random measurements and proof that the count is correct as per ripgrep, depending on the case sensitivity flag.

DmitryAstafyev (approved):

Issue #1972, opened 2024-01-04
Replace the crate async_trait with rust native support


AmmarAbouZor:
**Prerequisites**

affected version: Master

make sure to check this before filing an issue:

- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

**Describe what you would like to have**

Native support for async traits has been added to rust form Version 1.75, which should be able to replace using the crate async_trait in the most cases, especially with a stand-alone application where no other apps have dependencies on the codes like with libraries to consider.

I want to try to replace the crate async_trait with the native rust solution, but at first I want to make sure that there are no problems to enforce the developers to use rust Version 1.75+. 
Besides that we need to adjust the GitHub Pipelines to make sure that the latest rust version is installed

DmitryAstafyev:
make sense

AmmarAbouZor:
@DmitryAstafyev Should I try to implement this in the CLI tool at first, or it's better to wait to make sure that there is no problem with forcing the minimum rust Version to be 1.75?

DmitryAstafyev:
@AmmarAbouZor hello. 
I think CLI stuff and this issue don't depend on each other. For CLI we already have PR. As for this issue, it can go as a separate PR and you are welcome to create it from your fork as well.

AmmarAbouZor:
@DmitryAstafyev I thought it would some restriction because some developers in the team can't install rust Version 1.75, then they can't build either the app or the CLI-tool. 
I'll provide a PR to CLI-tool at first to check if we can apply it since the native support has its restrictions with concurrency too.

AmmarAbouZor:
So I've tried to apply the native async-trait on the CLI-tool and got some weird build error about cycles inside the traits, then I've tried to build it using rust nightly to get error messages which make sense:
```bash
    --> src/target.rs:48:30
    |
48  |     pub fn get(&self) -> Box<dyn Manager + Sync + Send> {
    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^ `Manager` cannot be made into an object
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
   --> src/modules/mod.rs:78:14
    |
61  | pub trait Manager {
    |           ------- this trait cannot be made into an object...
...
78  |     async fn reset(&self) -> Result<SpawnResult, Error> {
[...]

PR #1987, opened 2024-02-19
Remove async_trait crate and use native method

sudeeptarlekar:
From Rust1.75 onwards async support was added so we do not need async-trait crate anymore. Removed this crate in this PR and used native async support.

Resolves: #1972


DmitryAstafyev:
Looks like at we moment we cannot get rid of `async_trait`. Let's postpone it for ~1 year ;)

PR #2005, opened 2024-04-15
Fix issue with unicode & not valid utf8

DmitryAstafyev:
# 3.12.1 (16.04.2024)

## Fixes
- Fix issue related to invalid UTF-8 and Unicode


application/apps/indexer/processor/src/text_source.rs
```diff
@@ -15,6 +15,46 @@ use tokio_util::sync::CancellationToken;
...
+/// # Arguments
+///
+/// * `bytes` - A byte's slice to be decoded.
+/// * `placeholder` - Char to replace with invalid UTF8 characters
+///
+/// # Returns
+///
+/// Returns valid UTF8 `String`.
+///
+fn bytes_to_utf8(bytes: &[u8], placeholder: Option<char>) -> String {
```
marcmo, line 30: that looks very complicated for something that should be part of the stdlib...and indeed there is an easier way:
``` Rust
String::from_utf8_lossy(bytes).to_string()
```
that will not allow us do specify our own replacement characters, but should be sufficient!
Bonus: will pass all your tests ;)

example:
``` rust
    #[test]
    fn test_invalid_utf8_easy() {
        let bytes = vec![0x68, 0x65, 0x6C, 0x6C, 0x6F, 0xC1, 0x80];
        assert!(
            std::str::from_utf8(&bytes).is_err(),
            "Should be invalid UTF-8"
        );
        let serialized = bytes_to_utf8(&bytes, Some('?'));
        assert_eq!(serialized, "hello��");
    }
```

application/apps/indexer/processor/src/text_source.rs
```diff
@@ -308,3 +347,41 @@ fn count_lines_up_to_last_newline(buffer: &[u8]) -> (u64, usize) {
...
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::bytes_to_utf8;
+    use proptest::prelude::*;
+    use rand::{thread_rng, Rng};
+
+    fn strategy() -> impl Strategy<Value = Vec<u8>> {
```
marcmo, line 357: better then to create a strategy from hand, use the proptest way to compose:
``` rust
    prop_compose! {
        fn sample_string()
                    (mut pre in prop::collection::vec(any::<u8>(), 0..=0x7F),
                     mut post in prop::collection::vec(any::<u8>(), 0x80..=0xBF))
                    -> Vec<u8> {
                        let mut res: Vec<u8> = vec![];
                        res.append(&mut pre);
                        res.push(0xC1);
                        res.append(&mut post);
                        res
        }
    }
```

application/apps/indexer/sources/src/command/process.rs
```diff
@@ -91,6 +91,7 @@ impl ProcessSource {
             .stdout(Stdio::piped())
             .stderr(Stdio::piped())
             .stdin(Stdio::piped())
+            .kill_on_drop(true)
```
marcmo, line 94: not related to PR

application/apps/indexer/sources/src/command/process.rs
```diff
@@ -239,12 +241,11 @@ async fn test_process() -> Result<(), ProcessError> {
                 .is_some()
             {
                 assert!(!process_source.current_slice().is_empty());
-                // println!(
-                //     "{}",
-                //     std::str::from_utf8(process_source.current_slice()).unwrap()
-                // );
                 process_source.consume(process_source.current_slice().len());
             }
+            // By some reasons during test sometimes process stay alive and as result
```
marcmo, line 246: not PR related

marcmo (changes requested):

application/apps/indexer/processor/src/text_source.rs
DmitryAstafyev: Thanks for it. Have no idea how I've missed it...With it indeed fix became trivial.

application/apps/indexer/processor/src/text_source.rs
DmitryAstafyev: Thanks for it. Have no idea how I've missed it... with it indeed the fix became trivial

DmitryAstafyev: No need to test (at least on this level) as soon as there are no manual implementation any more.

application/apps/indexer/sources/src/command/process.rs
DmitryAstafyev: moved into another PR

marcmo (approved):
good to go if no performance hits are observed

PR #2006, opened 2024-04-17
Fix spawning process test

DmitryAstafyev:
By some reason after recent updates (crates and probably linux), from time to time after the test spawned process stays alive; as a result, we are stuck without a resolved test.

Probably makes sense to make a deeper investigation of such behavior.


application/apps/indexer/sources/src/command/process.rs
```diff
@@ -109,6 +109,7 @@ impl ProcessSource {
             .stdout(Stdio::piped())
             .stderr(Stdio::piped())
             .stdin(Stdio::piped())
+            .kill_on_drop(true)
```
marcmo, line 112: why do this only when not on windows?

marcmo (approved):

PR #2023, opened 2024-05-07
Up to 3.12.5

DmitryAstafyev:
# 3.12.5 (08.05.2024)

## Changes
- Check filters/charts during importing
- Highlight invalid filters/charts in sidebar
- Prevent search with invalid conditions
- Better error reporting on search fail
- Correction of filters/charts styles (sidebar)
- Allow modification of invalid filters/charts

### DIFF
--- a/application/apps/indexer/processor/src/search/searchers/regular.rs
+++ b/application/apps/indexer/processor/src/search/searchers/regular.rs
@@ -45,6 +45,15 @@ pub type RegularSearchHolder = BaseSearcher<RegularSearchState>;
 
 impl RegularSearchHolder {
     pub fn setup(&mut self, filters: Vec<SearchFilter>) -> Result<(), SearchError> {
+        let invalid = filters
+            .iter()
+            .filter_map(|f| if f.valid() { None } else { Some(&f.value) })
+            .cloned()
+            .collect::<Vec<String>>()
+            .join("; ");
+        if !invalid.is_empty() {
+            Err(SearchError::Input(format!("Invalid filters: {invalid}")))?;
+        }
         let mut aliases = HashMap::new();
         let mut matchers = vec![];
         for (pos, filter) in filters.iter().enumerate() {
--- a/application/apps/indexer/processor/src/search/searchers/regular.rs
+++ b/application/apps/indexer/processor/src/search/searchers/regular.rs
@@ -47,7 +47,8 @@ impl RegularSearchHolder {
     pub fn setup(&mut self, filters: Vec<SearchFilter>) -> Result<(), SearchError> {
         let invalid = filters
             .iter()
-            .filter_map(|f| if f.valid() { None } else { Some(&f.value) })
+            .filter(|&f| f.valid())
+            .map(|f| &f.value)
             .cloned()
             .collect::<Vec<String>>()
             .join("; ");
--- a/application/apps/indexer/processor/src/search/searchers/regular.rs
+++ b/application/apps/indexer/processor/src/search/searchers/regular.rs
@@ -47,7 +47,7 @@ impl RegularSearchHolder {
     pub fn setup(&mut self, filters: Vec<SearchFilter>) -> Result<(), SearchError> {
         let invalid = filters
             .iter()
-            .filter(|&f| f.valid())
+            .filter(|&f| !f.valid())
             .map(|f| &f.value)
             .cloned()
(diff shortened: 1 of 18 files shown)


application/apps/indexer/session/src/handlers/search.rs
```diff
@@ -34,6 +34,22 @@ pub async fn execute_search(
...
+        Err(NativeError {
+            severity: Severity::ERROR,
+            kind: NativeErrorKind::OperationSearch,
+            message: Some(format!(
+                "Next {} filter(s) are invalid: {}",
+                invalid.len(),
+                invalid.join("; ")
+            )),
+        })?;
+    }
```
marcmo, line 52: why not move this check to `searchers.rs` and execute it when calling `holder.setup`?

application/apps/indexer/session/src/handlers/search_values.rs
```diff
@@ -28,6 +31,22 @@ pub async fn execute_value_search(
...
+        .iter()
+        .filter(|f| !is_valid(f))
+        .cloned()
+        .collect::<Vec<String>>();
+    if !invalid.is_empty() {
+        Err(NativeError {
+            severity: Severity::ERROR,
+            kind: NativeErrorKind::OperationSearch,
+            message: Some(format!(
+                "Next {} filter(s) are invalid: {}",
```
marcmo, line 44: same thing. if you put it to as a function to the searchers module, you can call the check on both the search-holder and value-holder setup

marcmo: that way you also do not have to make the value field public API since you just need it inside the process crate

application/apps/indexer/session/src/handlers/search.rs
DmitryAstafyev: Motivation is: holder isn't shared "above" threads, it stored in `SessionState` and before use it, it should be taken from state and after usage I have to put it back. To avoid these steps (1. take holder, 2. try setup. 3. put it back if error) I've putted check here. Also we are checking a little different things: filter and search condition for values-extraction. In the first case we are considering flags, in the second - only regex as a string.

Well if we put this check inside a holder we will need more steps todo then now.

application/apps/indexer/session/src/handlers/search_values.rs
DmitryAstafyev: Actually I've missed issue with handling error on setup... it should put back holder, but doesn't

application/apps/indexer/processor/src/search/searchers/values.rs
```diff
@@ -44,6 +59,15 @@ pub type ValueSearchHolder = BaseSearcher<ValueSearchState>;
...
     pub fn setup(&mut self, terms: Vec<String>) -> Result<(), SearchError> {
+        let invalid = terms
+            .iter()
+            .filter(|f| !is_valid(f))
+            .cloned()
+            .collect::<Vec<String>>()
+            .join("; ");
+        if !invalid.is_empty() {
+            Err(SearchError::Input(format!("Invalid filters: {invalid}")))?;
+        }
```
marcmo, line 70: you are using basically the same algorithm twice...why not move it to the search module?
BTW this version of the filtering is nicer than converting to None/Some

DmitryAstafyev: do you mean `processor/src/search/searchers/mod.rs`?

but we are checking different things: in case of regular search we are checking SearchFilter in case of values - String ... check would be different in any way  :/

also then we get holder from state, we get some of holders: `get_search_holder` or `get_search_values_holder`. So in any way we are "working" with target holder

application/apps/indexer/processor/src/search/searchers/regular.rs
```diff
@@ -45,6 +45,15 @@ pub type RegularSearchHolder = BaseSearcher<RegularSearchState>;
 
 impl RegularSearchHolder {
     pub fn setup(&mut self, filters: Vec<SearchFilter>) -> Result<(), SearchError> {
+        let invalid = filters
+            .iter()
+            .filter_map(|f| if f.valid() { None } else { Some(&f.value) })
```
marcmo, line 50: this looks clearer to me:
``` rust
            .filter(|&f| f.valid())
            .map(|f| &f.value)
```

PR #2045, opened 2024-06-18
Producer Improvements & Unit Tests

AmmarAbouZor:
This PR provide unit-tests for producer main loop in the function `read_next_segment()`.

- [x] Tests should cover each case in the parse returns 
- [x] Code of producer is checked and a couple of notes are written in the code itself to be discussed with the reviewers
- [x] Add Unit-Test for byte sources to uncover the bugs found and solve them after that.

This PR will stay as Draft until the comments in code are resolved.


application/apps/indexer/sources/src/binary/pcap/legacy.rs
```diff
@@ -99,6 +99,9 @@ impl<R: Read + Send + Sync> ByteSource for PcapLegacyByteSource<R> {
                         if actual_tp == *wanted {
                             Ok(Some(ReloadInfo::new(
                                 received_bytes,
+                                // BUG: This should represent all available bytes in `self.buffer`.
+                                // This assumes that the buffer will be empty on each parse call
+                                // which will fail silently when parser implementing changes.
```
marcmo, line 104: you are right...from the view of the ByteSource, not every reload-call might have discarded the previously loaded bytes.
In practice with the protocols on top of ethernet it always is a completely new payload. But the concept of the ByteSource requires us to keep track of the consumed bytes.
Please write a unit-test that can show this behavior along with another test that processes multiple UDP packets. The first one should already fail, the second one should be green. Then you can work on correcting this behavior.

application/apps/indexer/sources/src/binary/pcap/ng.rs
```diff
@@ -96,6 +96,9 @@ impl<R: Read + Send + Sync> ByteSource for PcapngByteSource<R> {
                             transport: Some(wanted),
                         }),
                     ) => {
+                        // BUG: This should represent all available bytes in `self.buffer`.
```
marcmo, line 99: Same thing as above

application/apps/indexer/sources/src/command/process.rs
```diff
@@ -188,6 +188,9 @@ impl ByteSource for ProcessSource {
             let stored = line.len() + 1;
             self.buffer.copy_from_slice(line.as_bytes());
             self.buffer.copy_from_slice(b"\n");
+            // BUG: This should represent all available bytes in `self.buffer`.
```
marcmo, line 191: also correct observation! please expose this bug with a unit test and then fix it

AmmarAbouZor:
Unit Tests for all byte source has been added to insure reload result matches the `ReloadResult` signature. 
These tests uncovered that some byte sources weren't returning the correct available bytes, which has been fixed and the tests are green for all sources expect `PCAP` legacy source because I couldn't provide an input that allows the reload method to be called twice.

@marcmo Is it possible to have input for the legacy PCAP source that allows calling reload method twice?
Besides that, there are other potential bugs that we need to clarify in the PR

marcmo:
is this ready for review? if so please mark it non-draft

AmmarAbouZor:
@marcmo We still have some points that we need to clarify here than I can write the rest of the unit tests then it can be ready for review

AmmarAbouZor:
@marcmo I've resolve the issue with creating valid input for all the tests to pass.
Now we still have three points marked is `BUG` in file `/indexer/sources/src/producer.rs` that we still need to clarify, then this PR would be ready for final review

DmitryAstafyev (approved):

AmmarAbouZor:
Hi @DmitryAstafyev @marcmo,

* I fixed the final potential bugs and covered them in unit tests as well. Currently the method load won't return None if the byte-source doesn't load more data, but it still has none-consumed bytes. In case of parse error, the newly loaded bytes will be checked in avoid unending loops if parser can't parse the last few bytes remaining on a source
* Comments about cancel-safety for `load()` method has been added to structs implementing `income()` method on byte-source.
* I renamed `relaod()` methods to `load()` since these methods load the next available bytes appending them to the available one but they don't re-load data that is already loaded.

All points are done and this PR is ready for the final review.

PR #2046, opened 2024-06-19
Group shared dependencies with 'workspace.dependencies' 

AmmarAbouZor:
Cargo provides the feature [workspace.dependencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table) to manage shared dependencies among all libraries sharing the same workspace.

Using this feature I've gathered all shared dependencies with the same major version in `Cargo.toml` file in the `indexer` directory and replaced the specified version for dependencies in all its child-libs.

This refactoring shouldn't change the current dependencies tree because there is no changes in any of `Cargo.lock` files

### DIFF
--- a/application/apps/indexer/Cargo.toml
+++ b/application/apps/indexer/Cargo.toml
@@ -14,6 +14,28 @@ members = [
     "sources",
 ]
 
+[workspace.dependencies]
+log = "0.4"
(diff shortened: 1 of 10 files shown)


kruss (approved):
Nice!

PR #2049, opened 2024-06-27
Sessions: Cancel other tasks on operation errors

AmmarAbouZor:
This PR makes sure that cancel call will be invoked on the cancellation token instance if running producer returns with an operation error to avoid having the app stuck in `join!()` calls since some tasks like `tail::track()` won't return until cancel is called.

For Example the current Implementation won't show any error message in `indexer_cli` tool if `run_source()` call returns an error because it will be waiting for `tail::track()` to finish, which won't finish in that case until the user closes the session manually, and then at that point the error message will be printed.

### DIFF
--- a/application/apps/precompiled/updater/src/env/args.rs
+++ b/application/apps/precompiled/updater/src/env/args.rs
@@ -20,7 +20,10 @@ pub struct Arguments {
     pub app_name: String,
     pub location: PathBuf,
     pub compressed: PathBuf,
+    // `pid` and `ppid` aren't used yet but they can be used to check shutdown status by parent process.
+    #[allow(dead_code)]
     pub pid: Option<usize>,
+    #[allow(dead_code)]
     pub ppid: Option<usize>,
 }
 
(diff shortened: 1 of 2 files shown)


DmitryAstafyev:
Thanks @AmmarAbouZor . That's a very good find, @marcmo. Actually if our observation method is stuck for some reason we are losing a way to report it back because of tailing. This issue is relevant for file cases, and other use cases not affected.

DmitryAstafyev (approved):
Just clippy issues should be resolved. Clippy issues aren't related to PR's changes, but fixes can be included right into this one

AmmarAbouZor:
Linting issues from the updated has been fixed by suppressing the `unused` warning since the value can be used in the futures

application/apps/precompiled/updater/src/env/args.rs
```diff
@@ -20,7 +20,10 @@ pub struct Arguments {
     pub app_name: String,
     pub location: PathBuf,
     pub compressed: PathBuf,
+    // `pid` and `ppid` aren't used yet but they can be used to check shutdown status by parent process.
+    #[allow(dead_code)]
     pub pid: Option<usize>,
+    #[allow(dead_code)]
```
DmitryAstafyev, line 26: @marcmo I believe that's okay to ignore these warns in this case. That's actually not an implemented feature of the updater, which should check the state of the parent process and start updating as soon as the process is down. As soon as we didn't find quick solution for Windows we postpone this feature. But the parent process sends to updater PID as well and it makes sense to leave it there.

PR #2050, opened 2024-07-03
Separate codesign logic from electron builder for macOS

itsmesamster:


AmmarAbouZor:
These changes should be included in the build CLI tool as following:
* Release Command should have an argument `code-sign` that should provide the path for the signing configuration file.
* Configuration file should include all possible paths and command + Anything that can be changed often to avoid direct changes on Build CLI code.
* Configuration file format can be anything serialized with `serde`. It would be nice to have it in JSON format.
* Changes can be push directly to this PR
* Program should fail if this argument is provided on Linux or Windows

AmmarAbouZor:
Hi @itsmesamster @DmitryAstafyev,
I've implemented the code signing and notarizing for MacOS as following:

* I've added the CLI argument `--code-sign` or `-c` which accept the path for the configuration file of code signing.
* The implementation of code signing is a part of release on MacOS now when the configuration file & Environment variables are provided. 
* The configuration file is in `toml` format because we can add comments to it unlike `json` format.
* I've included all paths, commands and environment variables in the configuration file and created a configuration file with the current configurations talken from rake implementation under the path `cli/codesign_config/macos.toml`. The file is well documented to explain the meaning of those configurations
* An example for running the code signing:

```bash
cargo chipmunk release -c cli/codesign_config/macos.toml
```

DmitryAstafyev (approved):

DmitryAstafyev:
@AmmarAbouZor thanks for PR. A couple of moments:
- as soon as `application/holder/hooks/afterSign.js` isn't used - makes sense to remove it.
- please also make with @itsmesamster tests with failed signing process to make sure we have in output relevant error messages. For example, put into credentials some error, like `APPLEIDPASS = FAKEPASS`.

PR #2052, opened 2024-07-08
Core: Use Multiple Parse Results

AmmarAbouZor:
This PR is a work of progress and currently blocked by #2045 

As preparation for the plugins system, we need to change the parser trait signature and the producer stream to return multiple parse results at once instead of returning them one by one. 
The idea behind that to let the parsers plugins parse all the available bytes and return all the results to the host at once.

The changes include:
- `Parser::parse()` returns `impl IntoIter<Item = ParseResult> + Send` to avoid affect the performance of current parsers, because we can use `iter::once()` with them, which help the compiler to do more optimization since it knows at compile time that the results are one item only.
- Producer Stream returns a vector of results because there is one possible implementation for it and it's using vectors to store the results.

#### Road Map:
- [x] Change `Parser::parse()` and `Producer::as_stream()` signatures to return multiple results ate once.
- [ ] Make sure all parsers stop on the first potential error they encounter.
- [ ] Cover all edge cases with unit-tests after #2045 is merged
- [ ] Apply benchmarks on the this PR to make sure there is no performance loss

### DIFF
--- a/application/apps/indexer/addons/dlt-tools/src/lib.rs
+++ b/application/apps/indexer/addons/dlt-tools/src/lib.rs
@@ -51,13 +51,15 @@ pub async fn scan_dlt_ft(
                         canceled = true;
                         break;
                     }
-                    item = tokio_stream::StreamExt::next(&mut stream) => {
-                        match item {
-                            Some((_, item)) => {
+                    items = tokio_stream::StreamExt::next(&mut stream) => {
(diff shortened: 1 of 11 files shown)


AmmarAbouZor:
This PR can be closed because it will be replaced with new implementation based on unit tests & local benchmarks

PR #2054, opened 2024-07-18
Add protobuf module as wasm module

DmitryAstafyev:
## Description

To avoid generating a code for both (rust and JS) sides, we can:

- generate only a rust representation of the protocol. 
- create wasm module with generated code
- use wasm module on JS side to decode/encode messages

In this way we can be sure: that messages on rust and JS will be encoded/decoded in the same way

## Motivation

Research shows: that using different libraries/modules to generate code based on `*.proto` at the end can give different results. For example same message can be encoded into different bytes on JS side and on Rust side. That makes work with protocol unpredictable.


DmitryAstafyev:
General note: code isn't documented at all. It will be before merging.

application/apps/protocol/binding/proto/src/lib.rs
```diff
@@ -0,0 +1,113 @@
+///
+/// # Entities:
+///
+/// * `Range`
+/// * `RangeInclusive`
+///
+pub mod common {
+    include!(concat!(env!("OUT_DIR"), "/common.rs"));
```
AmmarAbouZor, line 8: I don't know if the backslash will be working on Windows, I think we can test that on Windows first before adjusting it manually. Hopefully the rust compiler can resolve automatically.

application/apps/protocol/binding/wasm/src/msg/attachment.rs
```diff
@@ -0,0 +1,11 @@
+use crate::*;
+use extend::extend;
+use prost::Message;
+use serde_wasm_bindgen::{from_value, to_value};
+use wasm_bindgen::prelude::*;
```
AmmarAbouZor, line 5: I think we can at some point include the use statements inside the macro itself by writing the full paths to the types inside it.
This would make any future changes easier by changing the macro itself and not having to add using statement for each file at a time.

application/apps/rustcore/rs-bindings/src/js/converting/observe.rs
```diff
@@ -0,0 +1,182 @@
...
+                factory::ParserType::Dlt(factory::DltParserSettings {
+                    filter_config: opt.filter_config.map(|opt| factory::DltFilterConfig {
+                        min_log_level: Some(
+                            if opt.min_log_level <= u8::MAX as u32
+                                && opt.min_log_level >= u8::MIN as u32
+                            {
+                                opt.min_log_level as u8
+                            } else {
+                                0
+                            },
```
AmmarAbouZor, line 135: I think this can be replaced with the `u8::try_from().unwrap_or(0)`

application/apps/rustcore/rs-bindings/src/js/converting/observe.rs
```diff
@@ -0,0 +1,182 @@
...
+                        context_ids: if opt.context_ids.is_empty() {
+                            None
+                        } else {
+                            Some(opt.context_ids)
+                        },
+                        ecu_ids: if opt.ecu_ids.is_empty() {
+                            None
+                        } else {
+                            Some(opt.ecu_ids)
+                        },
```
AmmarAbouZor, line 153: I've found a one liner for this pattern:
```rust
  ecu_ids: opt.ecu_ids.is_empty().not().then_some(opt.ecu_ids)
// or
  ecu_ids: opt.ecu_ids.is_empty().not().then(|| opt.ecu_ids)

```

application/apps/rustcore/rs-bindings/src/js/jobs/converting.rs
```diff
@@ -0,0 +1,179 @@
...
+                    command_outcome::Finished {
+                        result: Some(Output {
+                            output: Some(output::Output::StringValue(v)),
+                        }),
+                    },
+                )),
+            })
+            .unwrap_or(proto::CommandOutcome {
+                outcome: Some(command_outcome::Outcome::Cancelled(Cancelled {})),
+            });
```
AmmarAbouZor, line 39: If these parts are performance-critical then, I would try if using `unwrap_or_else()` here would bring any improvements because of its lazy loading. Although I don't think it would matter that much since the structs inside the unwrap function aren't that heavy

AmmarAbouZor (commented):
Wow this a hug and -at the same time- very clean PR :100: 
 
I had a quick look on the rust part and wrote some comments about the point that caught my eyes. Most of them are points for the final phase if the other performance issues are resolved, and some of them could bring tiny improvements for the performance.

application/apps/rustcore/rs-bindings/src/js/jobs/converting.rs
marcmo: nice work Dmitry!
until we have figured out the performance part, please mark it as "Draft"

PR #2056, opened 2024-07-18
Add Plugins Support with WebAssembly

AmmarAbouZor:
This PR still works in progress...

It will provide a plugin system for Chipmunk using WASM with [wasmtime](https://docs.rs/wasmtime/latest/wasmtime) runtime and the [component model](https://component-model.bytecodealliance.org/) definitions.

The Communication API between Chipmunk and the plugins is defined via [WIT files](https://component-model.bytecodealliance.org/design/wit.html), and the plugins could be written basically in any language which supports compiling to web assembly and the component model proposal.

For Plugins written in Rust we will provide a crate which will take care of generating the bindings form the `wit` files and provides all the needed functionalities for development (like logging).

The plugin system will offer writing plugins in the following areas:
- **Parser**: The parser will be given a buffer of bytes that needed to be parsed to the data types defined in chipmunk to make this data useful for analyzing.
- **Byte Source**: This part is responsible of providing data to chipmunk as array of bytes, then the app can parse and analyze this data.
- **Provider**: This part is responsible of both getting the data from the given configurations, parse them and then provide Chipmunk with parsing results then it can analyze them.

## Roadmap:
- [x] Setup and initialize one WASM host inside chipmunk that will be used with all the plugins of different types providing needed access to the plugins using WASI.
- [x] Provide a separate crate for plugins developers in rust to take care of the generating the types and communicating with the host and expose simple trait for the developers to implement and then they can use the provided macros to export their types as Chipmunk Plugins
- [ ] Provide templates for the developers to start from.
- [ ] Intergrate these templates with the newly developed [Build CLI tool](https://github.com/esrlabs/chipmunk/pull/1985) to generate the templates vie CLI commands

### Parser:
- [x] WIT definitions for communication between Chipmunk and the plugins.
- [x] Create Parser Host that loads, validates and communicates with the parser plugins.
- [x] Basic integration in the system with env-variable for benchmarks and testing purposes.
- [x] **Crate for Plugin developers**: Plugins developers need to implement a simple trait `Parser` for their types and then export this type via the macro `parser_export!()` provided by the crate
- [ ] Set how the users of chipmunk can control which plugin parser they want to use and implement that on front- and back-end
- [ ] Finish Documentation and provide all the needed information for the plugin developers and the end users of chipmunk

### Byte Source:
- [x] WIT definitions for communication between Chipmunk and the plugins.
- [ ] Create Byte Source Host that loads, validates and communicates with the source plugins.
- [ ] Basic integration in the system with env-variable for benchmarks and testing purposes.
- [ ] **Crate for Plugin developers**: Plugins developers need to implement a simple trait `ByteSource` for their types and then export this type via the macro `bytesource_export!()` provided by the crate
- [ ] Set how the users of chipmunk can control which plugin parser they want to use and implement that on front- and back-end
- [ ] Finish Documentation and provide all the needed information for the plugin developers and the end users of chipmunk


AmmarAbouZor:
Will be continued in branch `plugins_support` in mainstream repo
#2196

PR #2059, opened 2024-07-22
Resolve issue 2053 - Update on MacOS seems not to install the correct architecture.

itsmesamster:


application/holder/src/env/os/platform.ts
```diff
@@ -28,13 +29,14 @@ export function getPlatform(win32Only = false): Platform {
             break;
         case Platform.openbsd:
             return Platform.linux;
-        case Platform.darwin:
-            if (os.arch() === 'arm64') {
+        case Platform.darwin: {
+            const result = execSync('sysctl -n machdep.cpu.brand_string').toString();
```
DmitryAstafyev, line 33: `execSync` from [docs](https://nodejs.org/api/child_process.html#child_processexecsynccommand-options) : 

> If the process times out or has a non-zero exit code, this method will throw. 

Very dangerous because:

- it will be called at a very start of an application
- if some error - we will get `throw` during starting, it means an application probably will not start at all
- this is sync method, if some "issues" on a host (magic update of OS for example), we have a risk to wait here forever

I do not mind using something a platform specific, but here we cannot spawn external programs with potentially unpredictable result. It should be wrapped somehow or we should find NodeJS API to do that. Or maybe rust method, which can be called from here

DmitryAstafyev (changes requested):

application/holder/src/env/os/platform.ts
itsmesamster: Hi Dmitry, updated the code to include the changes you suggested. Let me know if anything else is required.

application/holder/src/env/os/platform.ts
```diff
@@ -15,6 +16,23 @@ export enum Platform {
...
 }
 
+let cachedCpuBrandString: string | null = null;
+
+function safeExecSync(command: string, timeout: number): string {
+    try {
+        return execSync(command, { timeout }).toString().trim();
+    } catch (error) {
+        return '';
+    }
```
DmitryAstafyev, line 26: Some logs are needed here.
```
import { scope } from 'platform/env/scope';
import { error } from 'platform/log/utils';
...
try {
        return execSync(command, { timeout }).toString().trim();
    } catch (err) {
        scope.getLogger('PlatformChecker').warn(`Fail to detect arch for darwin. Command "..." gives error: ${error(err)}`);
        return '';
    }
...
```

DmitryAstafyev: also makes sense to make
```
execSync(command, { timeout }).toString().trim().toLowerCase();
                                                                              ^^^^^^^^^^^^^^
```
It will allow accept also different cases (Intel, intel, INTEL)

application/holder/src/env/os/platform.ts
```diff
@@ -25,16 +43,16 @@ export function getPlatform(win32Only = false): Platform {
...
                 return Platform.linux;
             }
-            break;
         case Platform.openbsd:
             return Platform.linux;
-        case Platform.darwin:
-            if (os.arch() === 'arm64') {
+        case Platform.darwin: {
+            const result = getCpuBrandString();
+            if (os.arch() === 'arm64' || !result.includes('Intel')) {
```
DmitryAstafyev, line 50:
```
!result.includes('intel'))
                          ^^^^
```
Just "intel", all low-cases

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):
This PR is ready to be merged. @itsmesamster please make the necessary tests to make sure the update workflow works as well and with your confirmation it will go to master.

PR #2061, opened 2024-07-24
Issue 1993 + Issue 2057 (Benchmarks provisioning)

itsmesamster:


application/apps/rustcore/ts-bindings/spec/session.search.spec.ts
```diff
@@ -7,9 +7,12 @@ import { initLogger } from './logger';
 initLogger();
 import { Session, Factory } from '../src/api/session';
 import { IGrabbedElement } from 'platform/types/content';
-import { finish, createSampleFile, runner } from './common';
+import { finish, createSampleFile, performanceReport, setMeasurement, runner } from './common';
 import { readConfigurationFile } from './config';
 
+import * as os from 'os';
+import * as path from 'path';
```
DmitryAstafyev, line 14: also unused

application/apps/rustcore/ts-bindings/spec/session.benchmark.spec.ts
```diff
@@ -0,0 +1,306 @@
+// tslint:disable
+
+// We need to provide path to TypeScript types definitions
+/// <reference path="../node_modules/@types/jasmine/index.d.ts" />
+/// <reference path="../node_modules/@types/node/index.d.ts" />
```
DmitryAstafyev, line 5: by the way, can be removed. It's rudiment from times when we were using ts-jasmine

application/apps/rustcore/ts-bindings/spec/session.benchmark.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+
+                        const stream = session.getStream();
+                        if (stream instanceof Error) throw stream;
+
+                        const events = session.getEvents();
+                        if (events instanceof Error) throw events;
+
+                        const search = session.getSearch();
+
+                        let home_dir = (process.env as any)['SH_HOME_DIR'];
```
DmitryAstafyev, line 53: Potential bug. Case if `home_dir` doesn't have the correct value (at least not an empty string) isn't handled

application/apps/rustcore/ts-bindings/spec/session.benchmark.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+                        const session = await Session.create();
+                        session.debug(true, testName);
+
+                        const stream = session.getStream();
+                        if (stream instanceof Error) throw stream;
+
+                        const events = session.getEvents();
+                        if (events instanceof Error) throw events;
+
+                        const search = session.getSearch();
```
DmitryAstafyev, line 51: `session.getSearch()` can return `Error`. Case isn't handeled

application/apps/rustcore/ts-bindings/spec/session.benchmark.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+
+                        let home_dir = (process.env as any)['SH_HOME_DIR'];
+
+                        // Handle based on the test type
+                        switch (test.alias) {
+                            case 'Observe - grab content (text)':
```
DmitryAstafyev, line 57: Not good solution make selection based on name of test. `alias` here isn't like `id` or `uuid`, but it's name. Probably makes more sense to check by property name at least:
```
                  "test4": {  // <== better to use
                  ^^^^^^
                        "alias": "Stream - startup measurement", // better rename to "name" or "title"
                        "id" | "alias": "SOME_UNIQUE_ID" // another approach to add addition field
                        "file": "",
                        "expectation_ms": 1000
                    }
```
But in any way we cannot do like it
```
case 'Observe - grab content (text)':
```
Constants should be used instead
```
case BENCH_TEST_CASE_A:
```
You can create a list of references at some specific place. Hardcoding IDs/UUIDs/Aliases - not good idea

application/apps/rustcore/ts-bindings/spec/session.benchmark.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+                                            cwd: process.cwd(),
+                                            envs: process.env as { [key: string]: string },
+                                        })
+                                        .get()
+                                        .sterilized()
+                                );
+                                const startupResults = measurement();
+                                const startupReport = performanceReport(testName, startupResults.ms, test.expectation_ms, `${home_dir}/${test.file}`);
+                                finish(startupSession, done, startupReport ? undefined : new Error(`${testName} is fail`));
+                                break;
```
DmitryAstafyev, line 132: you are leaving a test without closing `session` which was created above at line 42

application/apps/rustcore/ts-bindings/spec/session.benchmark.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+                                            cwd: process.cwd(),
+                                            envs: process.env as { [key: string]: string },
+                                        })
+                                        .get()
+                                        .sterilized()
+                                );
+                                const shutdownResults = measurement();
+                                const shutdownReport = performanceReport(testName, shutdownResults.ms, test.expectation_ms, `${home_dir}/${test.file}`);
+                                finish(shutdownSession, done, shutdownReport ? undefined : new Error(`${testName} is fail`));
+                                break;
```
DmitryAstafyev, line 160: you are leaving a test without closing `session` which was created above at line 42

application/apps/rustcore/ts-bindings/spec/session.benchmark.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+                                            })
+                                            .get()
+                                            .sterilized()
+                                    ).catch((err) => `File ${i} failed to open: ${err.message}`);
+                                    results.push(result);
+                                }
+                                const multiResults = measurement();
+                                const multiReport = performanceReport(testName, multiResults.ms, test.expectation_ms, `${home_dir}/${test.file}`);
+                                finish(undefined, done, multiReport ? undefined : new Error(`${testName} is fail`));
+                                break;
```
DmitryAstafyev, line 195: exit from test without closing `session` created on line 42. Also sessions created in this block aren't closed

application/apps/rustcore/ts-bindings/spec/session.benchmark.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+                                throw new Error(`Unsupported format or alias: ${test.alias}`);
+                        }
+                        events.FileRead.subscribe(() => {
+                            const results = measurement();
+                            const reportResult = performanceReport(testName, results.ms, test.expectation_ms, `${home_dir}/${test.file}`);
+                            finish(session, done, reportResult ? undefined : new Error(`${testName} is fail`));
+                        });
+
+                    } catch (err) {
+                        finish(undefined, done, new Error(`Failed to complete test "${testName}" due to error: ${err instanceof Error ? err.message : err}`));
```
DmitryAstafyev, line 300: ambiguous situation. "Throwout" can occur when creating a session, which requires its preliminary closure

application/apps/rustcore/ts-bindings/spec/session.benchmark.spec.ts
DmitryAstafyev, line 51: In general, this block of code 
```
                        const session = await Session.create();
                        session.debug(true, testName);

                        const stream = session.getStream();
                        if (stream instanceof Error) throw stream;

                        const events = session.getEvents();
                        if (events instanceof Error) throw events;

                        const search = session.getSearch();
                        if (search instanceof Error) throw events;
```
Can be extracted into standalone function. Maybe makes sense also update/refactor others tests.
```
async function getTestSessionEnv(testName: string): Promise<{
    session: Session;
    events: ISessionEvents;
    stream: SessionStream;
    search: SessionSearch;
}> {
    const session = await Session.create();
    session.debug(true, testName);

    const stream = session.getStream();
    if (stream instanceof Error) {
        await session.destroy();
        return Promise.reject(stream);
    }

    const events = session.getEvents();
    if (events instanceof Error) {
        await session.destroy();
        return Promise.reject(events);
    }

    const search = session.getSearch();
    if (search instanceof Error) {
        await session.destroy();
        return Promise.reject(search);
    }

    return { session, events, search, stream };
}

```
This function can be called each time as soon as you need session and its dependencies (events, search, stream)

Probably make sense to create this function in `common` to reused it not only in benchmarks tests, but in regulars also

application/apps/rustcore/ts-bindings/spec/session.indexes.spec.ts
```diff
@@ -6,10 +6,13 @@
 import { initLogger } from './logger';
 initLogger();
 import { Session, Factory } from '../src/api/session';
-import { createSampleFile, finish, runner } from './common';
+import { createSampleFile, finish, performanceReport, setMeasurement, runner } from './common';
```
DmitryAstafyev, line 9: `performanceReport`, `setMeasurement` are unused

application/apps/rustcore/ts-bindings/spec/session.indexes.spec.ts
```diff
@@ -6,10 +6,13 @@
...
 import { readConfigurationFile } from './config';
 import { Nature, IndexingMode, NatureTypes } from 'platform/types/content';
 
+import * as os from 'os';
+import * as path from 'path';
+
```
DmitryAstafyev, line 15: rudiments. Aren't used

application/apps/rustcore/ts-bindings/spec/session.observe.spec.ts
```diff
@@ -12,6 +12,8 @@ import { IAttachment } from 'platform/types/content';
 import { createSampleFile, finish, performanceReport, setMeasurement, runner } from './common';
 import { readConfigurationFile } from './config';
 import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
```
DmitryAstafyev, line 16: Aren't used

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,157 @@
...
+              repo: 'chipmunk',
+              pull_number: ${{ github.event.inputs.pr_id }}
+            });
+            const PR_OWNER = pr.data.user.login;
+            const PR_REPO = pr.data.base.repo.name;
+            console.log(`owner is ${PR_OWNER}`);
+            console.log(`repo is ${PR_REPO}`);
+            // Setting environment variables
+            fs.appendFileSync(process.env.GITHUB_ENV, `PR_OWNER=${PR_OWNER}\n`);
+            fs.appendFileSync(process.env.GITHUB_ENV, `PR_REPO=${PR_REPO}\n`);
```
DmitryAstafyev, line 92: is it possible to move script into file? Workflow in general very complicated, would be nice to have a way to make it more simple at least with extracting some scripts (bash, js etc) into files

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

DmitryAstafyev:
@itsmesamster great job Sameer! Let's also consider new build CLI sys to run benchmarks and let it go

DmitryAstafyev:
@AmmarAbouZor in the scope of this PR we would need an additional build CLI command. Right now we have the command `cargo chipmunk test [TARGET]` (targets: core, shared, binding, wrapper, wasm, client, updater, app, cli). But we would need a custom js test runner to replace this (with ruby):

``` ruby
Shell.sh "#{Paths::JASMINE} spec/build/spec/session.#{spec}.spec.js"
```

To something like this (with CLI):
```shell
cargo chipmunk test bindings --custom=full_path/spec/build/spec/benchmarks.spec.js
```
Could you please take care of it as soon as you will have some time?

AmmarAbouZor:
Hi @DmitryAstafyev @itsmesamster 
Selecting custom specs is implemented in CLI tool. Now it's possible to specify one or multiple tests as following: 
* `cargo chipmunk test wrapper -s spec/build/spec/session.jobs.spec.js`: Runs single specification.
* `cargo chipmunk test wrapper -s spec/build/spec/session.jobs.spec.js -s spec/build/spec/session.search.spec.js`: Runs multiple specifications.

I've increased the version of the tool and added the change to the CHANGELOG as well.

PR #2064, opened 2024-08-01
Resolve issue with insufficient access for updates - issue 2063

itsmesamster:


DmitryAstafyev:
@itsmesamster looks very promising. Very good findings! I will try to test it until end of the week and we can merge it. Please switch it to "ready" mode if it's ready.

application/client/src/app/ui/layout/module.ts
```diff
@@ -12,6 +12,7 @@ import { MatDividerModule } from '@angular/material/divider';
 import { MatSnackBarModule } from '@angular/material/snack-bar';
 import { MatIconModule } from '@angular/material/icon';
 import { MatMenuModule } from '@angular/material/menu';
```
DmitryAstafyev, line 14: We never used angular tooltips before. To keep UI/UX the same for the whole solution, I would exclude this change from this PR. If we would need to switch to angular tooltips, we should do this for everywhere tooltips are used.

DmitryAstafyev (changes requested):

application/client/src/app/ui/layout/module.ts
itsmesamster: Removed the tooltip and altered the message.

DmitryAstafyev (approved):

PR #2073, opened 2024-08-21
Parse DLT network-trace for SOME/IP

kruss:
This PR adds support for parsing SOME/IP network-trace from DLT messages.

- Applys SOME/IP parsing to payloads of IPC and Someip DLT Network-Trace messages.
- Uses fallback SOME/IP service lookup in case payload versions don't match Fibex model.


application/apps/indexer/parsers/src/dlt/fmt.rs
```diff
@@ -281,6 +285,17 @@ impl<'a> Serialize for FormattableMessage<'a> {
...
             }
+            PayloadContent::NetworkTrace(slices) => {
+                state.serialize_field("app-id", &ext_header_app_id)?;
+                state.serialize_field("context-id", &ext_header_context_id)?;
+                state.serialize_field("message-type", &ext_header_msg_type)?;
+                let arg_string = slices
+                    .iter()
+                    .map(|slice| format!("{:02X?}", slice))
+                    .collect::<Vec<String>>()
+                    .join("|");
```
AmmarAbouZor, line 296: I'm not sure if using `reduce()` here could solve this better:
```rust
                let arg_string = slices
                    .iter()
                    .map(|slice| format!("{:02X?}", slice))
                    .reduce(|mut acc, s| {
                        acc.push('|');
                        acc.push_str(&s);
                        acc
                    })
                    .unwrap_or_default();
```
If you think this is better, then we can apply it to `PayloadContent::Verbose` match arm on line 240

application/apps/indexer/parsers/src/dlt/fmt.rs
```diff
@@ -386,6 +402,19 @@ impl<'a> FormattableMessage<'a> {
                     payload_string,
                 ))
             }
+            PayloadContent::NetworkTrace(slices) => {
+                let payload_string = slices
+                    .iter()
+                    .map(|slice| format!("{:02X?}", slice))
+                    .collect::<Vec<String>>()
+                    .join("|");
+                Ok(PrintableMessage::new(
```
AmmarAbouZor, line 411: Same as above if we want to use `reduce()`.

application/apps/indexer/parsers/src/dlt/fmt.rs
```diff
@@ -559,6 +588,36 @@ impl<'a> fmt::Display for FormattableMessage<'a> {
...
+                                    return write!(
+                                        f,
+                                        "{}",
+                                        format!("{message}").replace(SOMEIP_COLUMN_SENTINAL, " ")
+                                    );
+                                }
+                            }
+                        }
+                    }
+                }
```
AmmarAbouZor, line 615: I tried to reduce the nesting here and came up with: 
```rust 
let is_someip_trace =
    self.message
        .extended_header
        .as_ref()
        .is_some_and(|ext_header| {
            matches!(
                ext_header.message_type,
                MessageType::NetworkTrace(NetworkTraceType::Ipc)
                    | MessageType::NetworkTrace(NetworkTraceType::Someip)
            )
        });

if is_someip_trace {
    if let PayloadContent::NetworkTrace(slices) = &self.message.payload {
        if let Some(Ok((_, message))) = slices.get(1).map(|slice| {
            SomeipParser::parse_message(self.fibex_someip_metadata, slice, None)
        }) {
            return write!(
                f,
                "{}",
                format!("{message}").replace(SOMEIP_COLUMN_SENTINAL, " ")
            );
        }
    }
}
```
You can apply part of it, if you think would have improvements

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -11,9 +11,34 @@ use someip_payload::{
...
+    pub fn from_fibex_files(paths: Vec<PathBuf>) -> Option<Self> {
+        let readers: Vec<_> = paths
+            .iter()
+            .filter_map(|path| FibexReader::from_file(path).ok())
+            .collect();
+
+        if !readers.is_empty() {
+            if let Ok(model) = FibexParser::try_parse(readers) {
+                return Some(FibexMetadata { model });
+            }
```
AmmarAbouZor, line 29: `map()` then `ok()` can be used here
```rust
            return FibexParser::try_parse(readers)
                .map(|model| FibexMetadata { model })
                .ok();
```

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -129,7 +142,23 @@ impl Parser<SomeipLogMessage> for SomeipParser {
...
+impl Parser<SomeipLogMessage> for SomeipParser {
+    fn parse<'a>(
+        &mut self,
+        input: &'a [u8],
+        timestamp: Option<u64>,
+    ) -> Result<(&'a [u8], Option<ParseYield<SomeipLogMessage>>), Error> {
+        match SomeipParser::parse_message(self.fibex_metadata.as_ref(), input, timestamp) {
+            Ok((rest, message)) => Ok((rest, Some(ParseYield::from(message)))),
+            Err(error) => Err(error),
+        }
```
AmmarAbouZor, line 157: `map()` can be used here
```rust
        SomeipParser::parse_message(self.fibex_metadata.as_ref(), input, timestamp)
            .map(|(rest, message)| (rest, Some(ParseYield::from(message))))
```

application/apps/indexer/session/src/handlers/observing/mod.rs
```diff
@@ -94,10 +94,17 @@ async fn run_source_intern<S: ByteSource>(
         }
         ParserType::Dlt(settings) => {
             let fmt_options = Some(FormatOptions::from(settings.tz.as_ref()));
+            let someip_metadata = match &settings.fibex_file_paths {
+                Some(paths) => {
+                    FibexSomeipMetadata::from_fibex_files(paths.iter().map(PathBuf::from).collect())
+                }
+                None => None,
+            };
```
AmmarAbouZor, line 102: `and_then()` can be used here.
The people who are working on rust features are really working on reducing nesting :grinning: 
 
```rust
            let someip_metadata = settings.fibex_file_paths.as_ref().and_then(|paths| {
                FibexSomeipMetadata::from_fibex_files(paths.iter().map(PathBuf::from).collect())
            });
```

AmmarAbouZor (changes requested):
I think the main point is to replace the two optional fibex meta data with enum since they can't be provided at the same time. I Add a suggestion in one of the review comments.
Other than that, I've added some hints to reduce nesting and reduce allocating strings.

application/apps/indexer/parsers/src/dlt/mod.rs
```diff
@@ -67,10 +70,11 @@ impl LogMessage for RawMessage {
 #[derive(Default)]
 pub struct DltParser<'m> {
     pub filter_config: Option<ProcessedDltFilterConfig>,
-    pub fibex_metadata: Option<&'m FibexMetadata>,
+    pub fibex_dlt_metadata: Option<&'m FibexDltMetadata>,
     pub fmt_options: Option<&'m FormatOptions>,
     pub with_storage_header: bool,
     ft_scanner: FtScanner,
+    fibex_someip_metadata: Option<&'m FibexSomeipMetadata>,
```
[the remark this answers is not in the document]
kruss: "I think the main point is to replace the two optional fibex meta data with enum since they can't be provided at the same time." 
=> This assumption is not correct, they could be present same time!

In case the Fibex does contain metadata about NonVerbose DLT messages and SomeIp Services, both metatdata items would be used to to decode a DLT file containing any NonVerbose DLT messages or SomeIp payload in DLT Network-Trace messages.

application/apps/indexer/parsers/src/dlt/fmt.rs
```diff
@@ -196,7 +199,8 @@ impl From<Option<&String>> for FormatOptions {
 /// A dlt message that can be formatted with optional FIBEX data support
 pub struct FormattableMessage<'a> {
     pub message: Message,
-    pub fibex_metadata: Option<&'a FibexMetadata>,
+    pub fibex_dlt_metadata: Option<&'a FibexDltMetadata>,
+    pub fibex_someip_metadata: Option<&'a FibexSomeipMetadata>,
     pub options: Option<&'a FormatOptions>,
 }
```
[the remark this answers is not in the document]
kruss: See comment above.

application/apps/indexer/session/src/handlers/observing/mod.rs
kruss: yes, but sometimes in same way as c-devs used to reduce their var names.. ;-)

application/apps/indexer/parsers/src/dlt/fmt.rs
kruss: Used parts of it, but also now checking on errors in SomeipParser::parse_message now.

application/apps/indexer/parsers/src/dlt/fmt.rs
kruss: I don't see how is reduce expressing the intent better here.

application/apps/indexer/parsers/src/dlt/fmt.rs
kruss: I don't see how is reduce expressing the intent better here.

application/apps/indexer/parsers/src/dlt/fmt.rs
AmmarAbouZor: The point of reduce here is to avoid allocation for the strings into a vector and then apply join on it, by starting with one string and appending each string to it. 
I think it wouldn't have a big impact on the performance though and we can leave like this to favor code readability here.

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -250,68 +279,84 @@ fn option_string(option: &SdEndpointOption) -> String {
...
+                                service.name,
+                                method.name,
+                                match payload.is_empty() {
+                                    true => Cow::Borrowed(""),
+                                    false => {
+                                        let mut som_parser = SOMParser::new(payload);
+                                        som_type
+                                            .map(|mut value| match value.parse(&mut som_parser) {
+                                                Ok(_) => Cow::Owned(
+                                                    format!("{value}").replace([' ', '\n'], ""),
```
AmmarAbouZor, line 325: `to_string()` can be used here directly since we don't need any custom formatting here.
```rust
value.to_string().replace([' ', '\n'], "")
```

application/apps/indexer/parsers/src/someip.rs
```diff
@@ -250,68 +279,84 @@ fn option_string(option: &SdEndpointOption) -> String {
...
+                                                ),
+                                                Err(error) => {
+                                                    format!("'{}' {:02X?}", error, *payload).into()
+                                                }
+                                            })
+                                            .unwrap_or(format!("{:02X?}", *payload).into())
```
AmmarAbouZor, line 331: `unwrap_or_else()` would be more suitable here to evaluate the inner value only if needed
```rust
unwrap_or_else(|| format!("{:02X?}", *payload).into())
```

application/apps/indexer/parsers/src/dlt/fmt.rs
kruss: Using 'join' on string concatenation seems indeed to be the way of choice within the code-base at multiple occasions, while 'reduce' seems currently only to be used within some arithmetic operations. Therefore I would prefer to stay with the present solution, yet recommend a separe PR for a global change hereon, if and only if a performance impact evaluation would show some measurable benefit for the tradeoff of code readability.

kruss:
> This looks great! I added two small comments that can be fixed right away without the need to another review, and added an explanation for the `reduce()` suggestion

Using 'join' on string concatenation seems indeed to be the way of choice within the code-base at multiple occasions, while 'reduce' seems currently only to be used within some arithmetic operations. Therefore I would prefer to stay with the present solution, yet recommend a separe PR for a global change hereon, if and only if a performance impact evaluation would show some measurable benefit for the tradeoff of code readability.

AmmarAbouZor (approved):

DmitryAstafyev (approved):

DmitryAstafyev (approved):

PR #2074, opened 2024-08-21
Allow nested parsers on MessageProducer level

DmitryAstafyev:

## Solution limitations:

### DIFF
--- a/application/apps/rustcore/ts-bindings/package.json
+++ b/application/apps/rustcore/ts-bindings/package.json
@@ -33,7 +33,7 @@
     "loglevel": "^1.6.6",
     "tmp": "^0.2.3",
     "ts-node": "^10.4.0",
-    "typescript": "^5.1.6"
+    "typescript": "5.1.6"
   },
   "dependencies": {
     "platform": "link:../../../platform",
--- a/application/apps/rustcore/ts-bindings/yarn.lock
+++ b/application/apps/rustcore/ts-bindings/yarn.lock
@@ -5,13 +5,6 @@ __metadata:
   version: 8
   cacheKey: 10c0
 
-"@aashutoshrathi/word-wrap@npm:^1.2.3":
-  version: 1.2.6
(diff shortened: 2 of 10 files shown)


AmmarAbouZor:
I think we need to define which case is the general one and which is the special one...
* If the case nested parsers won't be needed for the most cases and it's needed for the DLT with SomeIP parser, then it would be better to not change all the structure because of the it. For example, in the PR most of the lambdas are empty and most parsers are ignoring them.
* We need to keep in mind that, this change with parser signature will make the potential parser plugins more complex.
* Alternatively, we can inject the function to the needed parser only. For example, we can change the DLT parser to accept an optional nested function that it can use within its implementation of the parse method, keeping the structure of the producer and other parsers unchanged.

AmmarAbouZor:
I'll add here more information about why the complexity of this approach on the plugin system...

In WASM with the component model we can't inject a function from the host in the parse function itself. In the best case we can define that function on the host, document it and mention in the documentation that the implementer of the parse function must consider it inside their implementation of the parse method, which means that the implementers of any parse method must deal with this special case in their implementation.

Another solution is to define a `resource` structure in wasm world that has the inject function define in it, then we can provide the implementation of this function in the plugins host. This could be the best solution for this scenario but it still requires that the plugins developers are more involved in the internal design of Chipmunk.

We can make it a little bit simpler for developing plugins in rust since we'll be providing a crate with built-in macro for exporting the methods. With that we can call the parse on behalf on the plugin developer which will lead to a lot of confusion on the code fails on that part.

Another possible solution for rust Plugins with help from our library is to make the parse function for them accept a function too and inject the provided function to form the host within the macro, but this solution still has a lot of magic that is happening plugins developed in rust only and it still to get the developer of the plugin more involved in the internal system of Chipmunk. 

We can exclude the plugin system from this approach and ignore the provided function from in producer in the parser host of the plugins, which feels like a source of potential bugs.

DmitryAstafyev:
@AmmarAbouZor thanks for explanation. Totally makes sense.

PR #2077, opened 2024-08-27
Allow nested parsers on MessageProducer level

DmitryAstafyev:

## Solution limitations:

Recreated from #2074


AmmarAbouZor:

PR #2078, opened 2024-08-27
Allow nested parsers on MessageProducer level (v2)

DmitryAstafyev:

## Solution limitations:

The previous version is here #2077

### DIFF
--- a/application/apps/indexer/session/src/handlers/observing/mod.rs
+++ b/application/apps/indexer/session/src/handlers/observing/mod.rs
@@ -10,7 +10,7 @@ use parsers::{
     dlt::{fmt::FormatOptions, DltParser},
     someip::SomeipParser,
     text::StringTokenizer,
-    LogMessage, MessageStreamItem, ParseYield, Parser,
+    LogMessage, MessageStreamItem, ParseYield, Parser, ParserInstance, ParserInstanceAlias,
 };
 use sources::{
     factory::ParserType,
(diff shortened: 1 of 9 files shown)


application/apps/indexer/session/src/handlers/observing/mod.rs
```diff
@@ -138,7 +146,34 @@ async fn run_producer<T: LogMessage, P: Parser<T>, S: ByteSource>(
...
+                                                continue;
+                                            }
+                                            Err(e) => {
+                                                item.resolve(Some(Err(e.to_string())));
+                                                continue;
+                                            }
+                                            _ => {
+                                                item.resolve(Some(Err(String::from(
+                                                    "Fail to extract printable message",
+                                                ))));
```
DmitryAstafyev, line 169: missed `continue;`

DmitryAstafyev:
Something, that I didn't expect to be honest :/

Screenshot from 2024-08-27 15-55-06

Test inputs:
- same file
- same workflow
- same OS (linux)

Test workflow:
- open chipmunk
- open file
- reopen file 6 times (without closing chipmunk, same instance)

DmitryAstafyev:
The major approach is: that the parser doesn't own any nested parsers, but still can request additional (nested) parsing. At the same time, only the initial parser knows how a complete message should be represented. The Message Producer doesn't know it, the initial parser - knows. That's why the initial parser has a way to request an addition (nested) parsing. 

At the same time, this solution gives flexibility and can be easily scaled. 

To compare two ways:
**Parser owns nested parser(s)**
- addition (nested) parsing happens inside the initial parser. It might cause problems with error handling, because now the work of the initial parser depends not only on itself but also on the nested parser's work.
- the work of the nested parser is out of the control of MessageProducer. For example, if we would have some configuration for the nested parsing approach (like disable/enable) that logic will go inside the initial parser; in the other hand, if an initial parser doesn't own nested parsers and requires additional parsing - MessageProducer can keep such settings under control.

**Parsers doesn't own nested parser(s) and requires parsing**
- MessageProducer keeps control of error handling of nested parsers
- MessageProducer can be configured to enable/disable nested parsing

In the same time a weak point of suggested solution - an initial parser should keep in memory result of nested parsing

AmmarAbouZor:

AmmarAbouZor:

Issue #2071, opened 2024-08-20
Consider Errors while parsing messages payload 


AmmarAbouZor:
The current implementation of `Parser<T>` sets the trait bound `Display` on T which can't fail, but with current situation while parsing SomeIP messages it's possible for a payload to be partly parsed with errors. The current situation is by showing these errors in the message description as text which could be confusing for the users because they can confuse them with the actual log content.

To Solve this issue, we need a way to tell that parsing the payload of a message has some problems, then we can show these errors in the UI in a clean way, like Highlighting the faulty entries and having another tab for errors where the users can jump search the errors and jump to the according log when clicked etc...

To Start resolving this on Rust level we need a way to tell that something went wrong by parsing the payload. One suggestion is replacing the trait bound `Display` on the Type T in `Parser<T>` to another trait that have a method that can return a String besides an optional problem field like
```rust
/// Trait to replace `Display` for T in parser
trait ToText {
    /// This can increase the performance for parser result types, that never fails since
    /// we can check on it in compile time
    const CAN_ERROR: bool;

    /// Method to be used instead of `to_string()`
    fn to_text(&self) -> ToTextReturn;
}

/// Return Type from to_text() method that have the message beside an optional problem value.
struct ToTextReturn {
    msg: String,
    issue: Option<Problem>,
}

/// Information about the value.
struct Problem {
    severity: Severity,
    msg: Stirng,
}
```

After that we can keep track on the errors by assigning the line number for each error - similar to attachment - and saving them in memory or to a separate file.

## Performance considerations:
With this check we are adding an extra check for each line, therefore we need to check the benchmark to measure the impact on that check on the performance.
- One suggested improvement is to attach a constant boolean for items that can't error to roll their checks out at compile time
- Another suggestion is to have the optional Problem boxed ` issue: Option<Box<Problem>>` to save some bytes on each return item in cost of saving them on the heap. This suggestion wouldn't have that much of optimizing the memory consumption since we'll check for the result and drop the empty one directly anyway.

## Another Suggestion: 
One more possible option could be shifting nested parsing out of `FormattableMessage` to somewhere "upper" in the parser itself if that possible.

Issue #2070, opened 2024-08-19
Integrate Tracing for better Logging and Debugging 


AmmarAbouZor:
It would be helpful to improve logging functionality in the Rust part of Chipmunk since this part is really hard to debug using a debugger and since using normal logging function in Async world is missing the context of the log information.

To solve these problems, I'll suggest using the [crate tracing](https://crates.io/crates/tracing) for the following benefits:
* It's designed for async scenarios, providing much more context for each log.
* Tracing level can be controlled in build time, giving us the option to remove lower tracing levels completely in release builds.
* Data from tracing can be used for more than normal logs with all the provided subscribers for telemetry and performance etc..
* Can be bridged with the normal logs as well.

AmmarAbouZor:
We can start with implementing it in the build CLI tool as a logging system to get a better feeling and some experience with it before implementing it in Chipmunk.

PR #2084, opened 2024-08-30
Allow nested parsers on MessageProducer level (v3)

AmmarAbouZor:
This PR isn't for to be merged. It's a suggestion in the scope of work on #2073 and #2078 
It's inspired the suggestion in #2071  

* `LogMessage` doesn't implement `Display` anymore. Instead, it has the method `try_resolve()` which can provide an enum of either a text message if the message can be resolved or a template of resolved and unresolved chunks when a full resolve isn't possible + Resolve hints for the session to know which method should be provided to resovle the rest of the chunks.
* The session has a `ParseRestResolver` struct which can try to resolved the rest unresolved chunks in the templates depending on the error hint, by having a SomeIP parser in the case of SomeIP Error hint.
* If the resolver fails to parse the rest of the message, we can add the message to the faulty messages store ( When should will be implemented as part of #2070 ) 

Here are basic benchmarks for the three PRs to solve the DLT with SomeIP case:
```
++++++++++++++  V1 ++++++++++++++ 

Run 1: 4051
Run 2: 4087
Run 3: 3931
Run 4: 4050
Run 5: 3989

Avg: (4051 + 4087 + 3931 + 4050 + 3989) / 5 = 4021.6


++++++++++++++  V2  ++++++++++++++

Run 1: 3852
Run 2: 3832
Run 3: 3805
Run 4: 3731
Run 5: 3788

Avg: (3852 + 3832 + 3805 + 3731 + 3788) / 5 = 3801.6


++++++++++++++  V3  ++++++++++++++

Run 1: 3917
Run 2: 4007
Run 3: 3818
Run 4: 3894
Run 5: 3875

Avg: (3917 + 4007 + 3818 + 3894 + 3875) / 5 = 3902.2
```


AmmarAbouZor:
I still think this kind of solutions should be implemented only if the standard case of processing the log is with various nested parsers. Otherwise, 
it would better to nest SomeIP within DLT parser to not change the general architectures to resolve a special case.

DmitryAstafyev:
@AmmarAbouZor thanks much for the possible solution. Here is a short summary by all current variants

| Criteria                         | V.1                             | V.2                  | V.3                  |
|----------------------------------|----------------------------------|----------------------|----------------------|
| Owner of nested parser           | initial parser                  | Producer             | Producer             |
| Scope of calling nested parser   | `Display()` of initial parser   | Producer Loop        | Producer loop        |
| Call of nested parser            | During rendering                | Before rendering     | Before rendering     |
| Nested parser error              | part of row (log message)       | bound error msg      | bound error msg      |
| Render row                       | By initial parser               | By initial parser    | By Producer with template |
| Changes in Producer's loop       | no                              | yes                  | yes                  |

If I miss something, let me know please

AmmarAbouZor:
> @AmmarAbouZor thanks much for the possible solution. Here is a short summary by all current variants

Hi @DmitryAstafyev, thanks for the useful summery for the three approaches. I think I may have a small remark on scope of V3 because this solution actually replaces `Display()` in the rendering phase with the new `try_resolve()` method, which makes calling the nested parsers rather in the rendering call than in the producer loop itself, or it sits in between.

AmmarAbouZor:

PR #2085, opened 2024-09-05
Build CLI: More Fixes & Improvements

AmmarAbouZor:
## Roadmap:
- [x] Reinstall TS packages in production mode only when the tool is creating a release, deactivating this step for the developer local production runs.
- [x] Never skip any job on release jobs.
- [x] Remove `--production` flag from `reset-checksums` command and reset checksum for both development and production since only one of them can be available at a time.
- [x] Provide integration python test for release command
- [x] Provide an extra safety action for graceful shutdown which will kill the app after 5 seconds from calling the method regardless on the current app state
- [x] Create change log file, providing the changes on each version 

This PR resolves a clippy warning by using `lines()` instead of `split()` with pattern

### DIFF
--- a/cli/src/cli_args.rs
+++ b/cli/src/cli_args.rs
@@ -137,11 +137,7 @@ pub enum Command {
     /// Resets the checksums records what is used to check if there were any code changes for
     /// each target.
     #[clap(visible_alias = "reset")]
-    ResetChecksum {
-        /// Reset release records
-        #[arg(short, long, default_value_t = false)]
-        production: bool,
-    },
+    ResetChecksum,
     /// Generate shell completion for the commands of this tool in the given shell,
     /// printing them to stdout.
     #[clap(visible_alias = "compl")]
--- a/cli/src/jobs_runner/jobs_state.rs
+++ b/cli/src/jobs_runner/jobs_state.rs
@@ -11,6 +11,10 @@ use crate::tracker::get_tracker;
 /// Duration to wait for jobs when cancellation is invoked.
 pub const TIMEOUT_DURATION: Duration = Duration::from_secs(2);
 
+/// Duration to wait after starting the shutdown process, to force the program to exit if it's
+/// still active after the given has passed.
+pub const FORCE_EXIT_DURATION: Duration = Duration::from_secs(2);
+
 /// [`JobsState`] singleton
 static JOBS_STATE: OnceLock<JobsState> = OnceLock::new();
 
@@ -63,7 +67,14 @@ impl JobsState {
         &self.task_tracker
     }
 
-    /// Closes the tasks trackers and waits for spawned jobs to exit gracefully within the [`TIMEOUT_DURATION`]
+    /// Closes the tasks trackers and waits for spawned jobs to exit gracefully within
+    /// the [`TIMEOUT_DURATION`].
+    ///
+    /// # Note:
+    ///
+    /// This function will spawn another thread to close the program immediately if graceful
+    /// shutdown has failed, which will wait for [`FORCE_EXIT_DURATION`] then forces the program
+    /// to exit.
     pub async fn graceful_shutdown(&self) {
         self.task_tracker.close();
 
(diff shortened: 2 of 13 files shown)


cli/src/cli_args.rs
```diff
@@ -137,11 +137,7 @@ pub enum Command {
     /// Resets the checksums records what is used to check if there were any code changes for
     /// each target.
     #[clap(visible_alias = "reset")]
-    ResetChecksum {
-        /// Reset release records
-        #[arg(short, long, default_value_t = false)]
-        production: bool,
-    },
+    ResetChecksum,
```
kruss, line 140: Maybe only naming it "reset", as user would not care about details (which could be in help description, only).

cli/src/jobs_runner/jobs_state.rs
```diff
@@ -11,6 +11,10 @@ use crate::tracker::get_tracker;
 /// Duration to wait for jobs when cancellation is invoked.
 pub const TIMEOUT_DURATION: Duration = Duration::from_secs(2);
 
+/// Duration to wait after starting the shutdown process, to force the program to exit if it's
+/// still active after the given has passed.
+pub const FORCE_EXIT_DURATION: Duration = Duration::from_secs(2);
```
kruss, line 16: I would give it some more time, maybe 5 seconds. On virus scanner days any machine can have some latency of 1-2 sec sporadically..

cli/src/jobs_runner/jobs_state.rs
```diff
@@ -22,31 +26,33 @@ pub struct JobsState {
     cancellation_token: CancellationToken,
     task_tracker: TaskTracker,
     fail_fast: bool,
+    is_app_release: bool,
```
kruss, line 29: Just a naming.. 'is_release_build' would make it more clear.

cli/src/jobs_runner/jobs_state.rs
```diff
@@ -61,16 +67,40 @@ impl JobsState {
...
+    /// shutdown has failed, which will wait for [`FORCE_EXIT_DURATION`] then forces the program
+    /// to exit.
     pub async fn graceful_shutdown(&self) {
         self.task_tracker.close();
 
         if timeout(TIMEOUT_DURATION, self.task_tracker.wait())
             .await
             .is_err()
         {
+            // If task_tracker fails to close here, then it could a dead-lock or another undefined
```
kruss, line 85: Spelling: "it could be a"

cli/src/jobs_runner/jobs_state.rs
```diff
@@ -80,4 +110,9 @@ impl JobsState {
     pub fn fail_fast(&self) -> bool {
         self.fail_fast
     }
+
+    /// Gets if the jobs are running to build and bundles a release of Chipmunk.
```
kruss, line 114: Spelling: "Gets <what?>"

cli/src/jobs_runner/mod.rs
```diff
@@ -167,7 +167,7 @@ fn spawn_jobs(
             continue;
         }
 
-        let skip = if job_def.job_type.is_part_of_build() {
+        let skip = if job_def.job_type.is_part_of_build() && !jobs_state.is_app_release() {
```
kruss, line 170: Could become "let skip = if job_def.job_type.is_skipable()"

kruss (approved):
Some small suggestions, that could be considered or not..

cli/src/cli_args.rs
AmmarAbouZor: I think `reset` only could be a little bit ambiguous and could mean `clean` as well. 
Actually, this command can be used with `reset` only because it has as an alias :grinning: 

```
reset-checksum    Resets the checksums records what is used to check if there were any code changes for each target [aliases: reset]
```

cli/src/jobs_runner/jobs_state.rs
AmmarAbouZor: The whole time here is actually 4 seconds because the app will wait for the first TIMEOUT_DURATION (2 Seconds) then this 2, but I think it would make sense to increase it more because I was trying it on my beefy machine :+1:

AmmarAbouZor:
@kruss Thanks for the great renaming suggestion, it makes the code clearer.
Naming variable and function has never been my strong suit :sweat_smile:

AmmarAbouZor:
I've resolved Clippy warning about pattern with `split()` by using the `lines()` method instead since the split is looking for new line character.

DmitryAstafyev (approved):

PR #2087, opened 2024-09-09
Allow nested parsers on MessageProducer level (v4) 

AmmarAbouZor:
This PR isn't for to be merged. It's a suggestion in the scope of work on #2073, #2078 and #2084
It provides the basic implementation for #2071 on parser level.

* The struct `ParseRestResolver` implemented to provide methods to parse a given chunks of bytes with a hint about how to parse those bytes.
* `LogMessage` doesn't implement Display anymore. Instead, it has the method `try_resolve(&self, resolver: Option<&mut ParseRestResolver>) -> Result<impl Display, ...>` to try to parse the message using the provided resolver if possible and returning an error to express the case when parsing fails.
* The trait `ParseLogMsgError` is defined with methods to parse the messages lossy (replace none-parsed bytes with their byte representation), besides the severity of the error and an error message to be used to collect the errors for proper error handling of parse log messages (as mentions in #2071)
* `LogMessage` has and associated error type which implement ParseLogMsgError, giving a chance for parser which never fails to provide Infallible Error type, to tell the compiler that this types will never error for compile time optimizations. Parser that can fail can use the struct `GeneralParseLogError` which implements the trait `ParseLogMsgError`.

### DIFF
--- a/application/apps/indexer/Cargo.lock
+++ b/application/apps/indexer/Cargo.lock
@@ -650,9 +650,8 @@ dependencies = [
 
 [[package]]
 name = "dlt-core"
-version = "0.14.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6fd69a328826e883613fbbcf468a33a53df1198c3e39bae7ad079c449431bfd"
+version = "0.16.0"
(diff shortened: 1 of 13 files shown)


AmmarAbouZor:

PR #2091, opened 2024-09-12
Replace async-trait crate with Rust native support

AmmarAbouZor:
This PR replaces the macro `async-trait` for `ByteSource` trait with rust native support for async functions in traits, avoiding the extra heap allocation on the return value that was enforce by the `async-trait` macro.

The warning about async in public traits is suppressed since we are using the trait in our own codebase only.

I've made simple benchmarks on a DLT file with the size 409M and got a slightly better performance:
```md
**Results in Mili-seconds**

# Native Support for async functions in traits:

File Read Took: 8182
File Read Took: 8113
File Read Took: 8225
File Read Took: 7996

** Avg: 8129 MS **
-----------------------------------------------------
# With 'async-trait' crate

File Read Took: 8395
File Read Took: 8258
File Read Took: 8252
File Read Took: 8291

** Avg: 8299 MS ** 
```

### DIFF
--- a/application/apps/indexer/Cargo.lock
+++ b/application/apps/indexer/Cargo.lock
@@ -136,17 +136,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-trait"
-version = "0.1.74"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


DmitryAstafyev (approved):
Finally! Very nice! Thanks Ammar!

PR #2097, opened 2024-09-19
Use dynamic caching for SOME/IP DLT network-trace parsing

kruss:
This PR improves the performance of SOME/IP DLT network-trace parsing by using dynamic caches for:

- Fallback Service-Version lookup
- Generated SOME/IP payload types

The resulting performance improvement is about 100%


AmmarAbouZor (commented):
I think using `Mutex` instead of `RwLock` here could be more efficient because the code is acquiring write access only and there is no use-case for read access.
I guess using `Mutex` would provide similar or slightly better performance with less overhead and dead locks potential

DmitryAstafyev (approved):

marcmo (approved):

PR #2102, opened 2024-09-23
Build CLI: Run Commands from within the Platform's Standard Shell 

AmmarAbouZor:
This PR changes the way the process commands (Async/Standard) run.
 
Previously, we tried to resolve each command with the `which` crate and ran each command directly as a process which still caused problems on some systems (Like x86 MacOS). 
This PR changes that to calling the standard shell on each platform (`sh` on unix based, `cmd` on Windows) and pass the process command as argument to the shell, avoiding the need for resolving the paths for development tools manually and allowing custom shell configurations on the users' side to be loaded.

Tested on Linux and Windows


mueller-o (approved):

mueller-o:
pls make sure to test on macos also

DmitryAstafyev:
@AmmarAbouZor looks good on mac (intel)
```
******** Tests for all commands of Chipmunk Build CLI Tool succeeded ********
```
@itsmesamster please check on arm also

DmitryAstafyev (approved):

PR #2106, opened 2024-09-30
Build CLI: Ensure Notarizing Success on MacOS by checking the command output 

### LINKED ISSUE #2100: Release process for macOS needs to fail if the notarisation step does not return the status as Accepted.
We have moved away from electron based code-signing for macOS to a separate, manual code-signing process.

Once the code signing is completed and we submit the signed app to the notary tool, the notary tool runs a number of validations on the signed files within the app, and if everything is fine, the command execution status is returned as success and provides a response with status="Accepted".
If there are issues identified by the notary tool, the command execution status is returned as success but the notary tool provides a response with status="Invalid". In this case, the process needs to fail, so that it becomes evident that the entire code sign process has been unsuccessful.

AmmarAbouZor:
This PR closes #2100 

Notarize command on MacOS return status OK even if the notarize has failed, therefore we must check the output of the command to ensure the success of the notarizing.

Since the command has the word `Accepted` many times (basically on each file) we can check for this word only.
However, the command returns currently the final result in one line `status: Accepted` [link from last release logs](https://github.com/esrlabs/chipmunk/actions/runs/11071421230/job/30763248391#step:12:1767) which we can compare each line against to rather have false negatives that positives.
The line to match against (currently `status: Accepted`) is set in the configuration file to avoid release a new version of the build cli tool each time apple changes the command output.


DmitryAstafyev:
@AmmarAbouZor thanks for these changes. I have only one question. As soon as output of macos cmds is out of our control and we cannot always be sure in default values of parameter `accepted_line` does it make sense to warn the user if in the configuration file parameter `accepted_line` is missed? Also, I would put some additional error output in case of failure, like "Please check field accepted_line to make sure output of notarization command is still relevant to configuration" or something like that.

AmmarAbouZor:
> @AmmarAbouZor thanks for these changes. I have only one question. As soon as output of macos cmds is out of our control and we cannot always be sure in default values of parameter `accepted_line` does it make sense to warn the user if in the configuration file parameter `accepted_line` is missed? Also, I would put some additional error output in case of failure, like "Please check field accepted_line to make sure output of notarization command is still relevant to configuration" or something like that.

Hi @DmitryAstafyev
The filed accepted_line is required while parsing the configuration files. I commented it out for testing and got the following error message before the release has started
```bash
Error: Error while serializing code sign configuration file

Caused by:
    TOML parse error at line 41, column 1
       |
    41 | [notarize_command]
       | ^^^^^^^^^^^^^^^^^^
    missing field `accepted_line`
```

I've updated the error message if the accepted line can't be found to the following:
``` 
Accepted line couldn't be found in notarize command output
Ensure that the content of the `accepted_line` field in the codesign configuration file is included as a single line in the output of the notarize command"
```

PR #2108, opened 2024-10-01
Support pretty format in details view

kruss:
This PR allows SOME/IP payloads to be rendered with pretty format in details view.

### DIFF
--- a/application/apps/indexer/parsers/src/someip.rs
+++ b/application/apps/indexer/parsers/src/someip.rs
@@ -21,6 +21,11 @@ use log::{debug, error};
 use regex::Regex;
 use serde::Serialize;
 
+/// Marker for a column separator in the output string.
+const COLUMN_SEP: &str = "\u{0004}"; // EOT
+/// Marker for a newline in the output string.
+const LINE_SEP: &str = "\u{0006}"; // ACK
+
 /// Wrapper for a fibex-model (new-type pattern).
 pub struct FibexMetadata {
     model: FibexModel,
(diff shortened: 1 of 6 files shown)


application/apps/indexer/parsers/src/someip.rs
```diff
@@ -85,7 +90,12 @@ impl FibexMetadata {
             if let Some(som_type) = lock.get_som_type(fibex_type) {
                 match som_type.parse(&mut SOMParser::new(payload)) {
                     Ok(_) => {
-                        return Some(som_type.to_string().replace([' ', '\n'], ""));
+                        return Some(
+                            som_type
+                                .to_string()
+                                .replace('\n', LINE_SEP)
+                                .replace(&" ".repeat(4), "\t"),
```
DmitryAstafyev, line 97: Thanks @kruss for this changes. Just one moment. If Some/IP parser already have tabs as 4 spaces in it's output, no needs to replace `\s\s\s\s` with `\t`, because on front-end (bellow in your changes) we can do just
```
this.row !== undefined
                    ? this.sanitizer.bypassSecurityTrustHtml(
                          this.row.content
                              .replace(/\u0006/gi, '<br>')
                              .replace(/\t/gi, '&nbsp;'.repeat(4))
                              .replace(/\s/gi, '&nbsp;'), // replace spaces with &nbsp; to make it visible in outout
                      )
                    : undefined;
```
I mean in current solution you can safely remove `.replace(&" ".repeat(4), "\t"),` and output in details view should stay same

DmitryAstafyev (changes requested):

DmitryAstafyev (approved):

PR #2115, opened 2024-10-09
Build CLI: Check for newer version locally & Commands Info in Logs

### LINKED ISSUE #2112: Check for updates on Build CLI tool
It would be useful to check on the latest available version of the build CLI tool and warn the user if they are not running the latest available version.

For that we can check the implementation on the updater which implement the same functionality

AmmarAbouZor:
This PR closes #2112 

* It adds a check to compare the version of the installed tool to the latest version of the build CLI tool on the user's local repository. This is pragmatic solution to avoid checking for the updates from master branch on GitHub on each run of the tool or develop more complex solution by checking for updates once daily and cache the value somewhere. 
Since the tool will be running inside Chipmunk repo anyway then it would make sense to check the available version on the local repo.
If this check fails, then a warning will be printed to the users, but the tool will continue executing.

* Logs are extended with each running command with the current directory where it's running

* This PR adds the version to the current output of the help command as well.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 0.2.6
+
+## Features:
+
+* Check the latest version of this tool on the local repository and compare it to the installed one.
+
+
 # 0.2.5
 
 ## Features:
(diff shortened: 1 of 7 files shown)


DmitryAstafyev (approved):

DmitryAstafyev:
Agreed, remote check can be added as another feature.

PR #2116, opened 2024-10-10
Parse DLT network-trace prefix for SOME/IP messages

kruss:
This PR will add information from the DLT network-trace prefix to the resulting SOME/IP information being displayed.

In DLT network-traces for SOME/IP the DLT payload contains of two arguments, where the first argument is a prefix with additional information on the actual SOME/IP message in the the second argument.

For the implementation the prefix will be expected to consist of:
- 4 bytes as IPv4 address
- 2 bytes as udp/tcp port
- 1 byte as protocol type (0 = local, 1 = tcp, 2 = udp)
- 1 byte as message direction (0 = incoming, 1 = outgoing)
- 1, 2 or 4 bytes as SOME/IP instance-id

### DIFF
--- a/application/apps/indexer/addons/someip-tools/src/lib.rs
+++ b/application/apps/indexer/addons/someip-tools/src/lib.rs
@@ -1,0 +1,153 @@
+use std::{
+    fmt::{self, Display},
+    net::Ipv4Addr,
+};
+
+use nom::{
+    combinator::map,
+    number::streaming::{be_u16, be_u32, be_u8},
+    sequence::tuple,
+    IResult,
+    Finish,
+};
+
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("Parse error: {0}")]
+    Parse(String),
+    #[error("Incomplete, not enough data for a message")]
+    Incomplete,
+}
+
+impl nom::error::ParseError<&[u8]> for Error {
+    fn from_error_kind(input: &[u8], kind: nom::error::ErrorKind) -> Self {
+        Error::Parse(format!(
+            "Nom error: {:?} ({} bytes left)",
+            kind,
+            input.len()
+        ))
+    }
+
+    fn append(_: &[u8], _: nom::error::ErrorKind, other: Self) -> Self {
+        other
+    }
+}
+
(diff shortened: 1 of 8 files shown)


DmitryAstafyev:
Requires discussion with @marcmo . In one hand this is not related to Some/IP parser directly and looks like cannot be a part of Some/IP parser; at the same time it's related to DLT parser, but not related to DLT protocol directly; but adding deep parsing operations into chipmunk directly is something, what we didn't do before. Let's discuss it.

DmitryAstafyev:
@kruss just as a mind. Could it be optional `feature` of dlt-core? Does it make sense?

kruss:
> @kruss just as a mind. Could it be an optional `feature` of some of the parsers? Does it make sense?

As it's not related to the protocol specifications of either DLT or SOME/IP it does imo not make sense. Probably this code could be moved to `application/apps/indexer/addons` in chipmunk?!

kruss:
Moved the code to parse someip network-trace prefixes to `application/apps/indexer/addons/someip-tools`

DmitryAstafyev (approved):

application/apps/indexer/addons/someip-tools/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+
+/// Parses a DLT Network-Trace prefix for a SOME/IP message from the given input.
+///
+/// A valid prefix will consist of:
+/// - 4 bytes as IPv4 address
+/// - 2 bytes as udp/tcp port
+/// - 1 byte as protocol type (0 = local, 1 = tcp, 2 = udp)
+/// - 1 byte as message direction (0 = incoming, 1 = outgoing)
+/// - 1, 2 or 4 bytes as instance-id
+pub fn parse_prefix(input: &[u8]) -> IResult<&[u8], std::string::String, Error> {
```
marcmo, line 45: best not to leak library dependent types in public APIs.
easily replaced by a general `Result`
```rust
pub type IResult<I, O, E = Error<I>> = Result<(I, O), Err<E>>;
```

kruss: done.

marcmo (approved):

PR #2129, opened 2024-10-22
Build CLI: Configurations on User Level & CI Improvements

### LINKED ISSUE #2119: Build CLI: Configuration file on User Level
Users of build CLI tool should be able to configure some aspects of the Build CLI Tool via configuration file.

Key Points:
* Configuration file location should be in `.chipmunk` directory in HOME
* Users should be able to set their shell their
* Users should be able to set their default -u mode
* CLI tool should provide commands to print the current configuration and to print the path of the configuration file on the system

### LINKED ISSUE #2140: Run CI check manually with UI option & Improvements
Following improvements and fixes are needed for the current checks for out CI checks for PRs and Master:
* It would be helpful to start the PR checks manually and choose the preferred UI option for the logs in `cargo chipmunk`. Available options are `print` and `immediate`
* `cargo chipmunk` commands for linting and testing shouldn't specify any target, then the tool itself will pick all the needed targets. Currently only some of the targets are specified which resulting in some if the targets being ignored.
* `wasm-pack` is installed currently via a script. But we can install it from cargo directly and get the benefits form caching it since we are using `cargo-cache` action in out pipelines

### LINKED ISSUE #2141: Build CLI: Timeouts for all commands
We need to add a timeout for process commands in the build CLI tool to avoid running the commands for a long time in GitHub Action if any command got into an infinite loop.
 
The timeout value should be configurable once the configuration file PR #2129  is merged and should have a reasonable default value (5-10) minutes.

AmmarAbouZor:
This PR closes #2119
This PR closes #2140
This PR closes #2141

It provides a way for the users to configure their environment and preferences in a file in Chipmunk home directory to be loaded and used by the build CLI tool

It Provides fixes improvements in fixes to run CI checks which are described in the linked issues.

The following configuration are provided currently:
* The Shell to run the processes on (Bash, Zsh, Fish...)
* Default UI preference to use with the tool (bars, report, print, immediate) 
 
## Roadmap:
- [x] Implement user configurations as singleton with their load method
- [x] Provide an enum to represent the shell running by users with methods to provide command to run process in the selected shell.
- [x] Replace the implementation global commands function with the one from the user configurations.
- [x] Add CLI commands to print the path of the configuration file and dump the default configurations to `stdout` and to the configuration file.
- [x] Validate given shell before running any further commands.
- [x] Consider preferred UI mode form the configuration.
- [x] Use default values for not defined configurations
- [x] Extend integration tests.
- [x] Extend README.
- [x] Increase tool version.
- [x] Timeout for all spawned tasks to prevent stuck task from running in GitHub actions forever. 
- [x] Remove specified targets on test commands in CI checks
- [x] Use cargo to install `wasm-pack`
- [x] Print error logs to `stderr` in `immediate` mode.
- [x] Print info line for skipped jobs in `immediate` mode.
- [x] Test on Windows.
- [x] Test on MacOS.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 0.2.8
+
+## Features:
+
+* User can configure the tool on user levels to set their preferred shell and UI mode.
+* Default shell will be retrieved from environment variables in unix-based environments.
+
 # 0.2.7
 
 ## Features:
--- a/.github/workflows/lint_master.yml
+++ b/.github/workflows/lint_master.yml
@@ -44,7 +44,7 @@ jobs:
       - name: install wasm-pack
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
       - name: JS/TS linting
-        run: cargo chipmunk lint -u print
+        run: cargo chipmunk lint -u immediate
       - name: TypeScript Check - Client Application
         working-directory: application/client
         run: yarn run check
@@ -92,4 +92,4 @@ jobs:
       - name: install wasm-pack
         run: cargo install wasm-pack
       - name: Execute tests
-        run: cargo chipmunk test -u print
+        run: cargo chipmunk test -u immediate
--- a/.github/workflows/lint_master.yml
+++ b/.github/workflows/lint_master.yml
@@ -92,4 +92,4 @@ jobs:
       - name: install wasm-pack
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
       - name: Execute tests
-        run: cargo chipmunk test core wrapper wasm -u print
+        run: cargo chipmunk test -u print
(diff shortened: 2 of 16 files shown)


AmmarAbouZor:
@DmitryAstafyev The default shell will be retrieved from SHELL environment variable.
I'll test this PR on Linux and Windows, but could you please test it on MacOS

cli/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.2.8
+
+## Features:
+
+* User can configure the tool on user levels to set their preferred shell and UI mode.
+* Default shell will be retrieved from environment variables in unis-based environments.
```
marcmo, line 6: unix

AmmarAbouZor:
Test passes on Windows

DmitryAstafyev (approved):

PR #2139, opened 2024-10-29
Core: Snapshot Testing for Output of Observe Session

AmmarAbouZor:
This PR provides new tests for the process of reading from binary files (DLT, Pcapng, Pcap legacy), then parsing with (DLT, SomeIP, DLT and SomeIP) parsers, then writing the output to Chipmunk temporary directory. By reading those temp files and having a snapshot for them as a reference to compare to of future changes.

* The crate [insta](https://crates.io/crates/insta) is used to handle snapshot testing. It's preferred to be used with its CLI tool [cargo-insta](https://crates.io/crates/cargo-insta).
* Output files will be read and represented as serializable struct which is written to the snapshot files via yaml format (The suggested format by `insta` crate) because it's clear to when diffing the snapshots
* Integration of the tool are built into Build CLI Tool with the option to accept all the changes at once. However, using `cargo insta` provides more benefits with colored diffs and reviewing each item snapshot at once.
* CI checks adjusted accordingly by adding CI environment flag as needed for `insta` crate.
* Documentation is added to the test file directly to make it easier to find for developers when tests are failing.


application/apps/indexer/session/src/lib.rs
```diff
@@ -1,4 +1,6 @@
-#![deny(unused_crate_dependencies)]
+// TODO AAZ: Better docs
+// Rust can't tell if a dependency is used in development only
+// #![deny(unused_crate_dependencies)]
```
AmmarAbouZor, line 3: Fix todo

marcmo (approved):
LGTM

AmmarAbouZor:
This PR provides missing updates in `Cargo.lock` files in master

DmitryAstafyev (approved):

PR #2146, opened 2024-11-11
Resolve rust version update issues with self hosted runner

itsmesamster:
This resolves the issue where self hosted runner did not get update the rust version and was locked on an older version of nj-cli(due to local dependencies on the self hosted runner).

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -35,25 +35,25 @@ jobs:
         with:
           node-version: "lts/*"
 
-      - name: Install Ruby and related tools
-        run: ./scripts/benchmarks/setup_dependencies.sh
+      - name: Install Ruby and other dev tools
+        run: ./scripts/benchmarks/setup_dev_tools.sh
 
       - name: Install cargo tools
         run: |
-          cargo install nj-cli wasm-pack --locked
+          cargo install nj-cli wasm-pack
           npm install -g tslib
 
       - name: Run Jasmine performance tests
         run: |
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-          export PATH="/root/.cargo/bin:$PATH"
-          source ~/.bashrc
-          if [[ "${{ github.event_name }}" == 'workflow_dispatch' && "${{ github.event.inputs.pr_id }}" -eq 0 ]]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -36,24 +36,24 @@ jobs:
...
 
       - name: Run Jasmine performance tests
         run: |
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-          export PATH="/root/.cargo/bin:$PATH"
-          source ~/.bashrc
-          if [[ "${{ github.event_name }}" == 'workflow_dispatch' && "${{ github.event.inputs.pr_id }}" -eq 0 ]]; then
-            echo "Running manually"
-            ruby scripts/benchmarks/process.rb 1
+          rustup -V
```
AmmarAbouZor, line 48: This is the command to get the version of rust toolchain manager and not rust version itself. Although they are printing rust version now but there is no guarantee this will still the case. To get Rust version we can use `rustc --version`

scripts/benchmarks/setup_ruby.sh
```diff
@@ -15,4 +15,9 @@ ruby --version
 sudo chown -R $(whoami) /usr/local
 rvm use 3.1.2 --default
 export PATH="/usr/share/rvm:$PATH"
-gem install dotenv json octokit tmpdir fileutils
\ No newline at end of file
+gem install dotenv json octokit tmpdir fileutils
+curl https://sh.rustup.rs -sSf | RUSTUP_INIT_SKIP_PATH_CHECK=yes sh -s -- -y
+rustup default stable
+rustup update
```
AmmarAbouZor, line 21: you can specify the channel you want to update here with `rustup update stable`

AmmarAbouZor (changes requested):
I've found small improvements on rust command to get and update the version.

~~Other than that, I've noticed that there is no code here to ensure that `nj-cli` is running in the latest version, but I don't know if we supposed to put this kind of check or the install command here or if another part on the server is responsible for that.~~

Edit: Also the name of the file `setup_ruby.sh` doesn't match its content anymore. Can we consider renaming it to `setup_env` or something similar

Edit2: I've just seen the command to install `nj-cli` tool

AmmarAbouZor (approved):

PR #2163, opened 2024-12-08
Include Build CLI integration tests in GitHub Action

itsmesamster:


.github/workflows/pullrequest_check.yml
```diff
@@ -59,15 +61,14 @@ jobs:
         run: yarn run check
 
   integration_and_unit_tests:
-    if: github.event.pull_request.draft == false
```
AmmarAbouZor, line 62: I don't think this line should be removed here. 
With this line removed the normal tests will run none cli related draft PRs.

.github/workflows/pullrequest_check.yml
```diff
@@ -59,15 +61,14 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2
       - name: libudev-dev
         run: |
           sudo apt-get update
-          sudo apt-get install -y libudev-dev
+          sudo apt-get install -y libudev-dev python3
```
AmmarAbouZor, line 71: I think we can move installing python to within running the cli tests job, because few PRs only will have the build CLI label. And the other PRs shouldn't have to install python each run (Although it may be installed already on ubuntu)

AmmarAbouZor (changes requested):
Very thanks for this PR.

I've found some small issues and points for improvements. 
Besides that, I think this change should be document somewhere (Maybe in Contribution or in Readme)

.github/workflows/pullrequest_check.yml
itsmesamster: That was so silly. I removed it by mistake. I have put that back in.

.github/workflows/pullrequest_check.yml
itsmesamster: Agreed. I have made the changes.

AmmarAbouZor (approved):

PR #2179, opened 2025-01-27
Remove buf redux

kruss:
This PR removes the chipmunk dependencies to buf_redux.

### DIFF
--- a/application/apps/indexer/addons/bufread/README.md
+++ b/application/apps/indexer/addons/bufread/README.md
@@ -1,0 +1,2 @@
+# bufread
+A buffered reader implementation in Rust with special behavior.
--- a/application/apps/indexer/addons/bufread/benches/bench_config.rs
+++ b/application/apps/indexer/addons/bufread/benches/bench_config.rs
@@ -1,0 +1,11 @@
+use criterion::Criterion;
+use std::time::Duration;
+
+pub fn bench_config(sample_size: usize) -> Criterion {
+    Criterion::default()
+        .warm_up_time(Duration::from_secs(5))
+        .measurement_time(Duration::from_secs(10))
+        .sample_size(sample_size)
+        .significance_level(0.01)
+        .noise_threshold(0.05)
+}
--- a/application/apps/indexer/addons/bufread/benches/reader_benches.rs
+++ b/application/apps/indexer/addons/bufread/benches/reader_benches.rs
@@ -1,0 +1,34 @@
+#![allow(dead_code)]
+
+use criterion::{criterion_group, criterion_main, Criterion};
+use std::hint::black_box;
+
+#[path = "./bench_config.rs"]
+mod bench_config;
+use bench_config::bench_config;
+
+#[path = "../tests/proto.rs"]
+mod proto;
+
+use bufread::BufReader;
+use proto::{Parser, Source, MAX_PACKET_LEN};
+
+fn proto_benchmark(c: &mut Criterion) {
+    let source_min_size = 100 * MAX_PACKET_LEN;
+    let buffer_max_size = 3 * MAX_PACKET_LEN;
+    let buffer_min_size = MAX_PACKET_LEN;
+
+    let source = Source::fixed(source_min_size);
+    let reader = BufReader::new(buffer_max_size, buffer_min_size, source.data());
+    let mut parser = Parser::new(reader);
+
+    c.bench_function("proto", |b| b.iter(|| Parser::run(black_box(&mut parser))));
+}
+
+criterion_group! {
+    name = benches;
+    config = bench_config(1000);
+    targets = proto_benchmark
+}
+
+criterion_main!(benches);
--- a/application/apps/indexer/addons/bufread/src/lib.rs
+++ b/application/apps/indexer/addons/bufread/src/lib.rs
@@ -1,0 +1,403 @@
+use std::{
+    cmp::min,
+    io::{BufRead, Read, Result, Seek, SeekFrom},
+    ptr::copy,
+};
+
+/// A buffered reader that could refill while still some remaining data is stored.
+pub struct BufReader<R> {
+    reader: R,
+    buffer: DeqBuffer,
+    min: usize,
+}
+
+impl<R> BufReader<R> {
+    /// Creates a new reader with the given maximum space and minimum bytes to be buffered ahead.
+    pub fn new(max: usize, min: usize, reader: R) -> Self {
+        BufReader {
+            reader,
+            buffer: DeqBuffer::new(max),
+            min,
+        }
+    }
+
+    /// Returns the total capacity of the inner buffer.
+    pub fn capacity(&self) -> usize {
+        self.buffer.capacity()
+    }
+
+    /// Returns the number of currently available bytes of the inner buffer.
+    pub fn len(&self) -> usize {
+        self.buffer.read_available()
+    }
+
+    /// Answers if the inner buffer is currently empty.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Returns the current slice to read from of the inner buffer.
+    pub fn buffer(&self) -> &[u8] {
+        self.buffer.read_slice()
+    }
+}
+
+impl<R: Read> Read for BufReader<R> {
+    fn read(&mut self, buffer: &mut [u8]) -> Result<usize> {
+        if self.buffer.write_available() < self.min {
+            self.buffer.flush();
+        }
+
+        let size = self.fill_buf()?.read(buffer)?;
+        self.consume(size);
+        Ok(size)
+    }
+}
+
+impl<R: Read> BufRead for BufReader<R> {
+    fn fill_buf(&mut self) -> Result<&[u8]> {
+        if self.buffer.read_available() < self.min {
+            self.buffer.flush();
+        }
+
+        if let Ok(size) = self.reader.read(self.buffer.write_slice()) {
+            self.buffer.write_done(size);
+        }
+
+        Ok(self.buffer.read_slice())
+    }
+
+    fn consume(&mut self, size: usize) {
+        self.buffer.read_done(size);
+    }
+}
+
+impl<R: Seek> Seek for BufReader<R> {
+    fn seek(&mut self, pos: SeekFrom) -> Result<u64> {
+        let result: u64;
+
+        if let SeekFrom::Current(pos) = pos {
+            result = self
+                .reader
+                .seek(SeekFrom::Current(pos - self.len() as i64))?;
+        } else {
+            result = self.reader.seek(pos)?;
+        }
+
+        self.buffer.clear();
+        Ok(result)
+    }
+}
+
+/// A fixed buffer that could be used as a byte dequeue.
+pub struct DeqBuffer {
+    slice: Box<[u8]>,
+    start: usize,
+    end: usize,
+}
+
+impl DeqBuffer {
+    /// Creates a new buffer with the given maximum space.
+    ///
+    /// # Safety
+    ///
+    /// The buffer needs to be written to before read from.
+    pub fn new(size: usize) -> Self {
+        let mut vec = Vec::with_capacity(size);
+        let cap = vec.capacity();
+
+        // SAFETY: Force the length of the vector to its capacity
+        // instead of using 'resize' for performance reasons.
+        unsafe {
+            vec.set_len(cap);
+        }
+
+        DeqBuffer {
+            slice: vec.into(),
+            start: 0,
+            end: 0,
+        }
+    }
+
+    /// Returns the total capacity of the buffer.
+    pub fn capacity(&self) -> usize {
+        self.slice.len()
+    }
+
+    /// Clears the buffer by consuming all available bytes
+    /// and returns the number of bytes being cleared.
+    pub fn clear(&mut self) -> usize {
+        let size = self.read_available();
+        self.read_done(size)
+    }
+
+    /// Reads from this buffer into the given output
+    /// and returns the number of bytes being read.
+    pub fn read_to(&mut self, buffer: &mut [u8]) -> usize {
+        let size = min(self.read_available(), buffer.len());
+        buffer[..size].copy_from_slice(&self.read_slice()[..size]);
+        self.read_done(size)
+    }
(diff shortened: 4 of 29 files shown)


application/apps/indexer/addons/bufread/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+        &mut self.slice[self.end..]
+    }
+
+    /// Sets the amount of newly written bytes.
+    pub fn write_done(&mut self, size: usize) -> usize {
+        let before = self.end;
+
+        self.end = min(self.end + size, self.slice.len());
+        self.end - before
+    }
```
AmmarAbouZor, line 195: Method names could be a little bit confusing because they do different kind of jobs but have same names:
* `write_available()` and `write_slice()` are both getters
* `write_done()` is a setter
* `write-from()` actually writes the data into internal buffer.

I think having the function names starts with get, do, set or mark would be helpful to understand their purpose.
The same is for read function above as well

application/apps/indexer/addons/bufread/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+        let before = self.start;
+
+        if self.start != 0 {
+            unsafe {
+                copy(
+                    self.read_slice().as_ptr(),
+                    self.slice.as_mut_ptr(),
+                    self.read_available(),
+                );
+            }
```
AmmarAbouZor, line 208: Also safety comment :wink:

application/apps/indexer/addons/bufread/tests/proto.rs
```diff
@@ -0,0 +1,153 @@
...
+        let mut packet: Vec<u8> = Vec::with_capacity(packet_len);
+        unsafe {
+            packet.set_len(packet_len);
+        }
+
+        let header = (payload_len as u16).to_be_bytes().to_vec();
+        packet[0] = header[0];
+        packet[1] = header[1];
+
+        rand::thread_rng().fill_bytes(&mut packet[HEADER_LEN..]);
```
AmmarAbouZor, line 85: The bytes in the slice `packet[2..HEADER_LEN]` are uninitialized memory which may lead to undefined behavior. Is this intended? In that case we may need some comments in code here to explain the intention behind that.

**Note**: I would consider initialize the packet here with zeros instead the unsafe code since performance here isn't that relevant

application/apps/indexer/addons/text_grep/src/buffer.rs
```diff
@@ -1,39 +1,62 @@
-use buf_redux::{
-    do_read,
-    policy::{DoRead, ReaderPolicy},
-    Buffer,
-};
+// TODO this duplicates: application/apps/indexer/processor/src/search/buffer.rs
```
AmmarAbouZor, line 1: Just to make sure this todo is intended to be done later and hadn't got forgotten :smile:

application/apps/indexer/addons/text_grep/src/buffer.rs
```diff
@@ -1,39 +1,62 @@
...
+        
+        self.buffer.read(buffer)
+    }
+}
 
-        let cap = buffer.capacity();
+impl<R: Read> BufRead for CancellableBufReader<R> {
+    fn fill_buf(&mut self) -> Result<&[u8]> {
+        if self.cancel.is_cancelled() {
+            return Ok(&[][..]);
```
AmmarAbouZor, line 39: This can be 
```rust
            return Ok(&[]);
```

application/apps/indexer/addons/text_grep/src/buffer.rs
```diff
@@ -1,39 +1,62 @@
-use buf_redux::{
```
AmmarAbouZor, line 1: Formatting seems to be broken on some files. Can you please run `cargo fmt` in `indexer` directory.
Running that on my machine introduced changes to the files:
* addons/text_grep/src/buffer.rs       
* processor/src/search/buffer.rs       
* processor/src/search/searchers/mod.rs

application/apps/indexer/processor/src/search/buffer.rs
```diff
@@ -1,43 +1,62 @@
...
+        
+        self.buffer.read(buffer)
+    }
+}
 
-        let cap = buffer.capacity();
+impl<R: Read> BufRead for CancellableBufReader<R> {
+    fn fill_buf(&mut self) -> Result<&[u8]> {
+        if self.cancel.is_cancelled() {
+            return Ok(&[][..]);
```
AmmarAbouZor, line 39: Like the other file the formatting is off and this line can be 
```rust
            return Ok(&[]);
```

AmmarAbouZor (changes requested):
Great work :+1: 
I've found some minor issues with naming, formatting and missing comments alongside with a suggestion to avoid undefined behavior inside one of the proptests

AmmarAbouZor:
It may be worth it to look into [miri](https://github.com/rust-lang/miri) to see if we can use it since we are introducing unsafe code in this PR

DmitryAstafyev (approved):
@kruss does `CancellableBufReader<R>` introduced twice in solution, but in different members? Does it make sense to move it out to have only one single implementation?

application/apps/indexer/addons/bufread/benches/reader_benches.rs
```diff
@@ -0,0 +1,46 @@
...
+#[path = "../tests/proto.rs"]
+mod proto;
+
+use bufread::BufReader;
+use proto::{Parser, Source, MAX_PACKET_LEN};
+
+fn proto_benchmark(c: &mut Criterion) {
+    let source_min_size = 100 * MAX_PACKET_LEN;
+    let buffer_max_size = 3 * MAX_PACKET_LEN;
+    let buffer_min_size = MAX_PACKET_LEN;
```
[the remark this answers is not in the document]
kruss: I don't think it's necessary to wrap the nested items of the parser, as it is the only input to the function under test and is then black-boxed by `c.bench_function("proto", |b| b.iter(|| Parser::run(black_box(&mut parser))));`. See [doc](https://doc.rust-lang.org/beta/std/hint/fn.black_box.html): "When benchmarking a function, you usually want to wrap its inputs in black_box so the compiler cannot make optimizations".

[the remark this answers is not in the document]
AmmarAbouZor: I think I'm still can't remove the best practices of the legacy `black_box()` function from criterion of my mind. 
That's right wrapping the function will black_box its output but won't prevent optimizations inside the function itself if that also matters benchmarks :+1:

application/apps/indexer/addons/bufread/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+}
+
+impl DeqBuffer {
+    /// Creates a new buffer with the given maximum space.
+    pub fn new(size: usize) -> Self {
+        let mut vec = Vec::with_capacity(size);
+        let cap = vec.capacity();
+        unsafe {
+            vec.set_len(cap);
+        }
```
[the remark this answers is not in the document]
kruss: done!

application/apps/indexer/addons/bufread/tests/proto.rs
kruss: Used `vec::resize` instead of `vec::set_len` here.

kruss:
> miri

Maybe it would make sense to add it to the build cli.

kruss:
> @kruss does `CancellableBufReader<R>` introduced twice in solution, but in different members? Does it make sense to move it out to have only one single implementation?

This was the case before this PR and should be considered in another PR, I created issue for it: https://github.com/esrlabs/chipmunk/issues/2191

application/apps/indexer/addons/text_grep/src/buffer.rs
kruss: I created an issue for it: https://github.com/esrlabs/chipmunk/issues/2191

application/apps/indexer/addons/bufread/README.md
```diff
@@ -0,0 +1,2 @@
+# bufread
+A buffered reader implementation in Rust with special behavior.
```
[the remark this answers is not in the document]
kruss: Not for now, as this is currently only an internal chipmunk crate. However a more detailed readme would make sense if the code becomes a standalone crate in future.

application/apps/indexer/addons/bufread/src/lib.rs
kruss: The API of the `DeqBuffer` is to be used for reading/writing from/to its internal buffer and signaling back the number of bytes read/written a posteriori, thus naming is not intended to reflect plain getter/setter semantics.

AmmarAbouZor (approved):

application/apps/indexer/addons/bufread/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    /// Moves any remaining bytes within the buffer to its front
+    /// and returns the number of bytes being moved.
+    pub fn flush(&mut self) -> usize {
+        let before = self.start;
+
+        if self.start != 0 {
+            // SAFETY: Copying the bytes is safe as pointers will not overlap.
+            unsafe {
+                copy(
```
marcmo, line 215: if pointers are guaranteed not to overlap, why not use `copy_nonoverlapping`?

application/apps/indexer/addons/bufread/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+            vec.set_len(cap);
+        }
+
+        DeqBuffer {
+            slice: vec.into(),
+            start: 0,
+            end: 0,
+        }
+    }
+
```
marcmo, line 133: couldn't it make sense for us not to forcefully shorten the vector in case more space was reserved than we need?
of course it would alter the semantics of the function, but it's internal API

application/apps/indexer/addons/bufread/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+            self.end -= self.start;
+            self.start = 0;
+        }
+
+        before
+    }
+}
+
+#[cfg(test)]
```
marcmo, line 230: since we have a `tests` module, why not put all the tests in it?

application/apps/indexer/addons/bufread/tests/proto.rs
```diff
@@ -0,0 +1,151 @@
...
+
+use bufread::BufReader;
+use rand::prelude::*;
+use std::{
+    io::{BufRead, Result},
+    mem::size_of,
+};
+
+/// The fixed header length of the protocol.
+pub const HEADER_LEN: usize = size_of::<u16>() as usize;
```
marcmo, line 28: `as` cast not needed

application/apps/indexer/addons/bufread/tests/proto.rs
```diff
@@ -0,0 +1,151 @@
...
+        Source { data, num_packets }
+    }
+
+    /// Creates a new packet with the given payload length.
+    fn create_packet(payload_len: usize) -> Vec<u8> {
+        assert!(payload_len <= MAX_PAYLOAD_LEN);
+
+        let packet_len = HEADER_LEN + payload_len;
+        let mut packet: Vec<u8> = Vec::with_capacity(packet_len);
+        packet.resize(packet_len, 0);
```
marcmo, line 77: better: ` let mut packet: Vec<u8> = vec![0; packet_len];`

application/apps/indexer/addons/bufread/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+mod tests {
+    use super::*;
+    use std::io::Cursor;
+
+    #[test]
+    fn test_buffer() {
```
marcmo, line 236: this would make the perfect prop test:
* randomly creating chunks of data and writing them
* interleaved with reading those chunks and clearing
has to work repeatedly

marcmo (commented):
really nice job Kevin! I left some comments.
After addressing those, we can merge it.

application/apps/indexer/addons/bufread/src/lib.rs
kruss: This are local unit-tests for the code in this source file and I would keep them here. In the tests module there are the prob-tests and some more comprehensive ("integration") tests using the crate for a pseudo-protocol parser, however.

application/apps/indexer/addons/bufread/src/lib.rs
kruss: Actually it's more a forcefully inflate of the vector, because after Vec::with_capacity "the returned vector has the minimum capacity specified, the vector will have a zero length". Instead of 'Vec::set_len' same can be achieved by 'Vec::resize', but with a significant performance drawback.

application/apps/indexer/addons/bufread/src/lib.rs
kruss: This is a mistake in my safety-comment. The regions do overlap, so "copy_nonoverlapping" cannot be use. What was meant by the safety comment was that the pointers will keep aliasing restrictions (eg do not "overlap") on copy the regions, which is sufficient for the "copy" call. I adjusted the safety comment accordingly.

application/apps/indexer/addons/bufread/benches/bench_config.rs
```diff
@@ -0,0 +1,23 @@
...
+//
+// NOTICE:  All information contained herein is, and remains
+// the property of E.S.R.Labs and its suppliers, if any.
+// The intellectual and technical concepts contained herein are
+// proprietary to E.S.R.Labs and its suppliers and may be covered
+// by German and Foreign Patents, patents in process, and are protected
+// by trade secret or copyright law.
+// Dissemination of this information or reproduction of this material
+// is strictly forbidden unless prior written permission is obtained
+// from E.S.R.Labs.
```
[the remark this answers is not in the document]
kruss: I removed the c-right statements in the newly added files.

application/apps/indexer/addons/bufread/src/lib.rs
kruss: Yes, but I would it in a separate PR, however.

PR #2186, opened 2025-01-28
Improvements to Generating typescript types

### LINKED ISSUE #2181: Improvement for generating TS types
The current workflow for the generating typescript types from rust types needed to optimize to resolve the following issues:
- [ ] The types will be generated into `bindings` directory inside `stypes` crates and must be copied manually to `platform`. Keeping two copies of those files is confusing and having to manually copy them each time can be avoided.
- [ ] The generated typescript files aren't formatted with `prettier` rules that are defined in the repo, leading to have the typescript files in various formats

AmmarAbouZor:
This PR closes #2181 

It provides the following:
* Sets the path for the generated files directly to `platform/types/bindings` to avoid the manual copying and having two sets for generated files at the same time.
* Add note on all types that are manually adjusted after generation to rule out any accidentally changes in the future.
* Apply formatting on all generated files to ensure that all of them have the same formatting enforcing further changes to apply the formatting as well. (This can be automated in separate PR)


DmitryAstafyev:
@AmmarAbouZor  using environment variables is not the most "transparent" approach. We want developers to understand what’s happening and how the protocol build process works, rather than relying on hidden "magic." Storing configuration in hidden files like `./.cargo/config.toml` makes the solution even less obvious for external developers.  

Improving the build system is a great goal, but I suggest sticking to a more conventional approach. If we want to make copying "automatic," let’s simply add a `build.sh` script that handles copying to the correct location and cleaning up temporary files. A straightforward script will make it easier for developers to understand what’s happening and how the process works.

AmmarAbouZor:
Actually, I've followed the recommended way to set the export path of the generated types from `ts-rs` [Documentation](https://docs.rs/ts-rs/10.1.0/ts_rs/trait.TS.html#:~:text=Generates%20a%20test,OVERRIDE_DIR%3E%22%2C%20relative%20%3D%20true%20%7D).
Besides, that I've written comments in multiple places in code to explaining how it's configured.

Personally, I don't find it more convenient to add extra step by manually copying the files after generating them that lives outside of the rust-code base in an external bash script. I think that having the solution inside rust code with the documentation as recommended by the crate developers is more convenient.

DmitryAstafyev:
I understand that you followed the recommended approach from the `ts-rs` documentation. However, "recommended" doesn’t always mean "best for every project."

The issue here is not whether the solution is technically possible, but how easy it is to discover and understand. Scattering comments in multiple places and relying on a hidden `.cargo/config.toml` file makes the process less transparent for new developers. Someone cloning the repo won’t immediately know that they need to tweak a hidden config file before running the build. And looking for explanation in docs of `ts-rs`. Not everyone is deeply familiar with Cargo internals, and a hidden `.cargo` folder can be confusing.  

A separate `build.sh` script would provide a single, clear entry point. No need to dig through documentation or track down scattered comments—just run one script, and everything works. This keeps things explicit and maintainable for the team in the long run. Any necessary comments, by the way, can be included directly in the bash script.  

Would you be open to discussing an approach that balances automation with clarity?

AmmarAbouZor:
Sure, the approach from `ts-rs` team isn't optimal by any mean. It's just as none-optimal as having a script to manually move items after generation. 
I understood that `.cargo/config` is and for project specific configuration and it's not a per-user [link](https://doc.rust-lang.org/cargo/reference/config.html#:~:text=Cargo%20allows%20local%20configuration%20for%20a%20particular%20package%20as%20well%20as%20global%20configuration.).
I think it's similar to how `.github` directory is a dot directory that lives with project repository.

Anyway, the two approaches aren't optimal in my opinion, and I can change it to manually copying if you find it better

DmitryAstafyev:
I would say like it: if I would have my own private project, I would select your suggested way (using .cargo/config.toml); if I'm choosing for public repo - I would follow transparency for other users, even I could not like something ... that's life ;)

AmmarAbouZor:
It would be actually interesting to get more opinions about this, because I still can't imagine a scenario where copying files manually in a separate script is the more convenient option.

@marcmo @kruss Could you please give us some insights here.

**tldr:***
The suggested way from `ts-rs` crate to control the path of the generated typescript files is to set an environment variable inside cargo configuration file located in `stypes/.cargo/config.toml`, which is implemented here with comments in the code.
@DmitryAstafyev thinks that this is confusing for developer since the configuration file is somehow hidden for developers and it would be less confusing to keep the files generated in the default path and extend the bash script for generating the files with other commands to move those generated files to the wanted destination after generating them.

@DmitryAstafyev Please feel free to extend this short description if I have missed something here

DmitryAstafyev:
I’m very grateful for this initiative. The desire to make the solution cleaner and better is always welcome.

However, we have already discussed this topic in detail, a decision has been made, and our priority is transparency, clarity for the majority of users, and minimizing dependencies on third-party APIs. Besides, we already have test.sh to run tests, and we are already using shell scripts to manage this sub-project.

Also, as mentioned last Tuesday, I specifically asked not to spend time on this but to focus on high-priority tasks.

I don’t believe it’s productive to continue this discussion at this point, nor to take up our colleagues’ time on a secondary matter.

Therefore, I’m closing this PR. We can revisit it later, once high-priority tasks are completed - and only if we encounter any actual issues that need to be addressed.

At the moment, I don’t see any such issues.

Please, let’s focus on the priority work.

AmmarAbouZor:
I think we can close the issue here since it's getting out of sync.
It's still linked to the original issue so won't get forgotten

PR #2204, opened 2025-02-13
Add *.md support for plugin's details

DmitryAstafyev:
# Added
- Plugins Manager makes an attempt to read `README.md` file in plugin's folder. 
- If  `README.md` has been found, it would be rendered in details frame

## Change
- Introduce `PluginRunData` entity to store run, load and checks data


AmmarAbouZor (approved):
This looks great.

We will still need to deliver the path of the md file from the backend, to keep all loading logic in one place and having the front end knowing less as possible.
I'll create another PR for that once I'm done with changes on the plugins manager

application/apps/indexer/stypes/src/plugins/extending.rs
```diff
@@ -84,3 +148,58 @@ impl Display for PluginType {
...
+    }
+    pub fn info<S: AsRef<str>>(&mut self, msg: S) {
+        self.logs.push(PluginLogMessage::info(msg));
+    }
+    pub fn err<S: AsRef<str>>(&mut self, msg: S) {
+        self.logs.push(PluginLogMessage::err(msg));
+    }
+    pub fn warn<S: AsRef<str>>(&mut self, msg: S) {
+        self.logs.push(PluginLogMessage::warn(msg));
+    }
```
AmmarAbouZor, line 164: * [optional] With Rust [String deref coercion](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/deref-coercions.html#:~:text=Using%20an%20ampersand%20in%20front,%26str%20%2C%20which%20foo()%20takes.) you can remove the generic bounds and use `&str`
* Very short documentation may be needed here since the methods are public

application/apps/indexer/stypes/src/plugins/extending.rs
```diff
@@ -84,3 +148,58 @@ impl Display for PluginType {
...
+}
+
+impl PluginLogMessage {
+    fn tm() -> u64 {
```
AmmarAbouZor, line 168: Can we rename it to `timestamp()`

application/apps/indexer/stypes/src/plugins/extending.rs
```diff
@@ -84,3 +148,58 @@ impl Display for PluginType {
...
+            tm: Self::tm(),
+        }
+    }
+    pub fn warn<S: AsRef<str>>(msg: S) -> Self {
+        Self {
+            msg: msg.as_ref().to_string(),
+            level: PluginLogLevel::Warn,
+            tm: Self::tm(),
+        }
+    }
```
AmmarAbouZor, line 204: Also the optional removing generic bound with deref coercion + Short docs

application/apps/indexer/stypes/src/plugins/mod.rs
```diff
@@ -157,8 +160,57 @@ pub struct PluginEntity {
...
+#[cfg_attr(
+    all(test, feature = "test_and_gen"),
+    derive(TS),
+    ts(export, export_to = "plugins.ts")
+)]
+pub struct PluginLogMessage {
+    /// The severity level of the log message.
+    pub level: PluginLogLevel,
+    /// The timestamp of when the log message was generated, represented as a Unix timestamp.
+    pub tm: u64,
```
AmmarAbouZor, line 196: Also renaming this to `timestamp` would be great :pray:

application/apps/indexer/stypes/src/plugins/proptest.rs
```diff
@@ -244,15 +240,10 @@ impl Arbitrary for InvalidPluginEntity {
...
-        )
-            .prop_map(|(dir_path, plugin_type, error_msgs)| Self {
+        (any::<PathBuf>(), any::<PluginType>())
+            .prop_map(|(dir_path, plugin_type)| Self {
                 dir_path,
                 plugin_type,
-                error_msgs,
             })
             .boxed()
     }
```
AmmarAbouZor, line 249: I think we need a proptest for `PluginRunData` here

application/apps/protocol/src/lib.rs
```diff
@@ -137,6 +137,7 @@ gen_encode_decode_fns!(PluginConfigValue);
 gen_encode_decode_fns!(PluginConfigSchemaType);
 gen_encode_decode_fns!(PluginConfigSchemaItem);
 gen_encode_decode_fns!(PluginEntity);
+gen_encode_decode_fns!(PluginRunData);
```
AmmarAbouZor, line 140: I think we need to extend the tests in `ts-bindings/spec/session.protocol.spec.ts`

AmmarAbouZor (changes requested):
Great work nicely done @DmitryAstafyev 

I still found some points mainly regarding the naming of some fields like `ts` or `rd`. Also we may need to extend the proptests and protocol test

application/apps/indexer/plugins_host/src/plugins_manager/mod.rs
```diff
@@ -30,41 +33,87 @@ impl PluginsManager {
...
     }
 
     /// Provide all invalid plugins.
-    pub fn invalid_plugins(&self) -> &[InvalidPluginEntity] {
-        &self.invalid_plugins
+    pub fn invalid_plugins(&self) -> Vec<&InvalidPluginEntity> {
+        self.invalid_plugins
+            .iter()
+            .map(|en| en.into())
+            .collect::<Vec<&InvalidPluginEntity>>()
```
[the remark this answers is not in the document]
DmitryAstafyev: I have changed the function signature from `pub fn invalid_plugins(&self) -> Vec<&InvalidPluginEntity>` to `impl Iterator<Item = &InvalidPluginEntity>` to align with the original implementation of `PluginsManager`, which contains methods returning iterators.

However, @AmmarAbouZor , I want to point out that you are making the API more complex and less transparent. Returning a vector provides a clear understanding of the API’s purpose, making it more readable and intuitive. Returning an iterator, on the other hand, is a less obvious design choice that does not provide any significant advantages.

I don’t see any places in the code where an iterator is actually needed—we always work with vectors. If an iterator is required for a specific purpose, a dedicated method like installed_plugins_iter would be a much more canonical approach.

I kindly ask you to consider this and strive to keep the API (both public and internal) simple and clear. The pursuit of "perfect" code should not come at the cost of clarity and intuitiveness.

application/apps/indexer/plugins_host/src/plugins_manager/mod.rs
```diff
@@ -30,41 +33,87 @@ impl PluginsManager {
     }
 
     /// Provide full infos of all loaded and valid plugins.
-    pub fn installed_plugins(&self) -> &[PluginEntity] {
-        &self.installed_plugins
+    pub fn installed_plugins(&self) -> Vec<&PluginEntity> {
+        self.installed_plugins
+            .iter()
+            .map(|en| en.into())
+            .collect::<Vec<&PluginEntity>>()
```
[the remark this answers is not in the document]
DmitryAstafyev: Left comment about it bellow

application/apps/indexer/stypes/src/plugins/extending.rs
DmitryAstafyev: Using `&str` instead of the more flexible `<S: AsRef<str>>` in logging methods might seem reasonable at first, as it does not introduce any significant overhead. However, using `AsRef<str>` allows for much broader API usability. The user does not need to worry about the exact type of string they pass—whether it’s `&str`, `String`, `Cow<str>`, or even `Path`. The API seamlessly handles these cases, providing maximum flexibility.

I see no valid reason to restrict this API to `&str` only. Moreover, I want to emphasize that this is not a matter of any meaningful resource consumption. The overhead is negligible, while the benefits of maintaining a more versatile API are clear.

application/apps/indexer/stypes/src/plugins/extending.rs
DmitryAstafyev: Docs added as for rest, left comment above

application/apps/indexer/plugins_host/src/plugins_manager/mod.rs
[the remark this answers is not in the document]
AmmarAbouZor: Thanks Dmitry, thanks for pointing out the point of clarity of the function. 
Actually, having the function returning a vector can be more readable if we are not used functions returning iterations, but they would save us one not needed memory allocation if users don't want to allocate the provided values directly, but they want to do some mapping on them before collecting them. Which is actually the case in `application/apps/indexer/session/src/unbound/commands/plugins.rs`
Before:
```rust
// First allocation of memory
fn installed_plugins(...) -> Vec<..> { self...collect() };

// Then in use we allocated memory another time.
let plugs = manager.installed_plugins().into_iter().cloned().collect();
```
Now
```rust
// No allocations
fn installed_plugins(...) -> impl Iterator<...> { ... }
// One allocation on use only
let installed_plugins = manager.installed_plugins().cloned().collect();
```

The performance different here isn't noticeable. It's just that when you right too much Rust you kinda get really sensitive to each memory allocation :crying_cat_face:

AmmarAbouZor (approved):
This looks great, very thanks for the changes and the clarifications on some of the tricky points :pray:

PR #2226, opened 2025-02-14
Fixes on TCP source and producer loop

AmmarAbouZor:
This PR is work in progress.

## TCP Source:
Tcp source should return the count of copied bytes into buffer, because that's the count what get copied into buffer, not the bytes count we get from the socket.


AmmarAbouZor:
Close in favor for #2230

PR #2195, opened 2025-02-07
Remove legacy text grep

kruss:
This PR fixes #2191

PR #2229, opened 2025-02-17
Remove legacy addon text_grep

kruss:
Remove legacy addon text grep #2195


marcmo (commented):
sorry...this is still needed in chipmunk in the future, can't remove

marcmo:
can you change the usage of buf_redux?

PR #2230, opened 2025-02-17
Socket sources: Improvements for internal buffers

AmmarAbouZor:
This PR includes:
* Increase the capacity of the internal buffer in socket byte sources to 1MB instead of the default value 8KB.
* Add Checks for socket byte source to avoid filling the internal buffer if the server is sending the data with high frequency because the producer loop will call load on each iteration while some parsers will parse one item only on each iteration.
* Call flush on consume when there is not enough writable memory.
* Another check if the buffer is filled and return an error if we can't add all the incoming bytes into the buffer.

### DIFF
--- a/application/apps/indexer/sources/src/socket/tcp.rs
+++ b/application/apps/indexer/sources/src/socket/tcp.rs
@@ -4,7 +4,7 @@ use crate::{
 use bufread::DeqBuffer;
 use tokio::{net::TcpStream, task::yield_now};
 
-use super::{ReconnectInfo, ReconnectResult, ReconnectToServer};
+use super::{ReconnectInfo, ReconnectResult, ReconnectToServer, MAX_BUFF_SIZE, MAX_DATAGRAM_SIZE};
 
 pub struct TcpSource {
     buffer: DeqBuffer,
@@ -14,8 +14,6 @@ pub struct TcpSource {
     reconnect_info: Option<ReconnectInfo>,
 }
 
-const MAX_DATAGRAM_SIZE: usize = 65_507;
-
 impl TcpSource {
     pub async fn new<A: Into<String>>(
         addr: A,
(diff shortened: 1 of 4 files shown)


AmmarAbouZor:
I've created two tests to demonstrate two issues:
* **TCP test** where the `DeqBuffer` is panicking
* **UDP test** prints how many bytes won't be parsed with parsers not consuming all the available bytes

AmmarAbouZor:
* TCP test is fixed to not reproduce the panic. However, this shows us that the safe functions of `DeqBuffer` could produce an undefined behavior if miss used.
* UDP test is working now, but it still printing how many bytes won't be parsed
* The ultimate fix for the cumulated bytes is to change parsers implementation to parse as possible from the provided slice of bytes and to not parse the first item only. I think This change is out of scope of this PR because it would introduce bigger changes + performance lose with the current implementation of the producer loop

DmitryAstafyev (approved):

application/apps/indexer/sources/src/socket/tcp.rs
```diff
@@ -92,6 +90,14 @@ impl ByteSource for TcpSource {
         &mut self,
         _filter: Option<&SourceFilter>,
     ) -> Result<Option<ReloadInfo>, SourceError> {
+        // If buffer is almost full then skip loading and return the available bytes.
+        // This can happen because some parsers will parse the first item of the provided slice
+        // while the producer will call load on each iteration making data accumulate.
+        if self.buffer.read_available() > MAX_BUFF_SIZE - MAX_DATAGRAM_SIZE {
```
marcmo, line 96: better: 
```rust
if self.buffer.write_available() < MAX_DATAGRAM_SIZE
```

AmmarAbouZor: Done

marcmo (approved):
LGTM
just one comment

PR #2231, opened 2025-02-17
Development CLI: Remove targets record file on Clean.

### LINKED ISSUE #2227: Development CLI: Remove checksum file on all clean commands
The clean command currently tries to update the checksum file to remove the checksum of the involved targets only. 
This would reduce the reliability of this command in case the checksum file in the repo has an out-dated structure. 
Therefore we need to remove the checksum file upfront and avoid updating it.

AmmarAbouZor:
This PR closes #2227 

It provides Removing record file entirely on the start of any clean job instead of trying to update the record file, making clean command more reliable because updating the record file may fail more often than removing it.

### DIFF
--- a/cli/development-cli/CHANGELOG.md
+++ b/cli/development-cli/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.4.1
+
+## Changes:
+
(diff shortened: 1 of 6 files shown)


marcmo (approved):
LGTM

PR #2233, opened 2025-02-18
Parsers returns multiple items & Reuse Items buffer instead of streams 

AmmarAbouZor:
This PR introduces the following changes:

### Producer:
* Unused field `index` is removed.
* Cancel safety if `load_next_segment()` is described in the comments in code and tested manually.
* Unit tests has been adjusted for the new producer API, and they all pass without any change in their core functionality. 

### Parsers:

**Note**: The checks on TCP & UDP byte sources introduced in #2230 aren't needed anymore for the built-in parser. However, I kept them to support plugin parser that can parse one item at a time only.

### Fixes:
* Mock parser for benchmarks avoids allocating memory by using `iter::repeate_n()` instead.
* Chipmunk CLI: Add warning for external dependencies.

----

## Performance Changes:
```md

### Current PR:
File Read Took: 12058 ms
File Read Took: 11972 ms
File Read Took: 12103 ms

### Master: 
File Read Took: 12648 ms
File Read Took: 12639 ms
File Read Took: 12680 ms

### Current PR:
File Read Took: 3509 ms
File Read Took: 3461 ms
File Read Took: 3538 ms

### Master: 
File Read Took: 3518 ms
File Read Took: 3485 ms
File Read Took: 3507 ms

### Current PR:

### Master: 
```

### DIFF
--- a/application/apps/indexer/Cargo.lock
+++ b/application/apps/indexer/Cargo.lock
@@ -930,13 +930,11 @@ name = "dlt-tools"
 version = "0.1.0"
 dependencies = [
  "dlt-core",
- "futures",
  "indexer_base",
  "log",
  "parsers",
  "sources",
  "tokio",
- "tokio-stream",
  "tokio-util",
 ]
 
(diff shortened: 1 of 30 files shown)


AmmarAbouZor:
We still need to have more considerations for cancel safety without using streams

AmmarAbouZor:
Closed in favor of #2244

PR #2238, opened 2025-02-21
Plugin protobuf

DmitryAstafyev:
## Adds
- Adds Example of Protobuf parser, which is suitable only for TCP/UPD streams and pcap/pcapng files


plugin_examples/protobuf/src/protobuf/parser.rs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        // Read descriptor file (it may include multiple protobuf definitions)
+        let pool = DescriptorPool::decode(fs::read(desc_path)?.as_slice())?;
+
+        // Collect all available message descriptors
+        let mut dmsgs = Vec::new();
+        for file in pool.files() {
+            dmsgs.append(&mut get_message(&file));
+        }
```
AmmarAbouZor, line 51: If you want you can replace this with the one linerd;
```rust
        let dmsgs: Vec<_> = pool.files().flat_map(|f| get_message(&f)).collect();
```

plugin_examples/protobuf/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+        let msg = self.one_of_any(data);
+        if msg.is_none() {
+            log::warn!("Fail to decode bytes {:x?}", data);
+        }
+        let msg = self
+            .one_of_any(data)
+            .map(|(name, msg)| item(name, msg.to_text_format(), len as u64))
+            .or_else(|| Some(item("_", &format!("{:X?}", data), len as u64)));
+        Ok(iter::successors(msg, |_| None))
+    }
```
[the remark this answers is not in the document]
AmmarAbouZor: I think we don't need to do that for now since we still don't have a way to determine if the bytes are part of an incomplete message

plugin_examples/protobuf/src/protobuf/parser.rs
AmmarAbouZor: This is unnecessary for now as well

AmmarAbouZor (approved):

AmmarAbouZor (approved):

PR #2240, opened 2025-02-25
TCP/UDP Sockets: Move flush to load method.

AmmarAbouZor:
Flushing the internal buffer on load is more reliable to avoid potential bugs when having an incomplete parser result on last packet before having the internal buffer almost filled.


AmmarAbouZor:
Closed in favor of #2244

PR #2241, opened 2025-02-26
Update observe related view

DmitryAstafyev:
## Changes
- Update observe view to make it for friendly for users. Actually, it removes a nested list of plugins and includes it into the main list of parsers.


AmmarAbouZor (approved):
Looks great, Thanks @DmitryAstafyev

PR #2244, opened 2025-02-26
Producer: Replace Stream with Caching & Parsers: Return Multiple values & TCP Socket Reconnect

AmmarAbouZor:
~This PR extends #2233 and replaces #2240.~
~It marked is draft until #2233 is merged~

This PR replaces ##2240 and #2233. 

It provides the following changes:

### Parsers General:

### Producer Loop:

* Current unit tests have been adjusted with the new producer API, and they all pass without any change in their core functionality. 
* Optimize handling parse error by avoiding dropping all loaded bytes. Instead, it will skip them one by one and reapply parse to avoid data loss.
* Abort the session on continues initial parse errors for the first megabyte of input bytes to avoid calling parse repeatedly on each byte drop of the source in case the input isn't suitable for the selected parser.
* Extend unit tests for producer with mocks to cover parse error cases.
* Extend unit tests for cancelation safety of producer loop covering all parser error cases.

### TCP/UDP Sources:
* Moves reconnect to separat task making `load()` method cancel safe. `load()` task will be communicating with reconnecting task via channels and both of them are sharing the same TCP socket, which will be replaced on successful reconnect attempts.
* Unit tests for Reconnect and Cancel Safety.
* Flushing the internal buffer is moved to `load()` method to tie freeing bytes in internal buffer and loading it with more data, making it more reliable to avoid potential bugs when having an incomplete parser result on last packet before having the internal buffer almost filled.

### Chipmunk CLI:
* Provide CLI arguments for configuring auto-reconnect in case of physical reset to the TCP server.
* Update README & Increase app version & changelog 

### SomeIP parser:
Fix converting errors from SomeIP crate into parser, ensuring `Incomplete` error type is delivered correctly in producer loop since this error will be handled different than parse errors.

## Performance Changes:
```md

### Current PR:
File Read Took: 12058 ms
File Read Took: 11972 ms
File Read Took: 12103 ms

### Master: 
File Read Took: 12648 ms
File Read Took: 12639 ms
File Read Took: 12680 ms

### Current PR:
File Read Took: 3509 ms
File Read Took: 3461 ms
File Read Took: 3538 ms

### Master: 
File Read Took: 3518 ms
File Read Took: 3485 ms
File Read Took: 3507 ms

### Current PR:

### Master: 
```


AmmarAbouZor:
Needs to be split into two PRs

AmmarAbouZor:
Done in #2256

PR #2251, opened 2025-03-03
Improve TCP Reconnect & Handling Parse errors & TCP internal buffer

AmmarAbouZor:
#2244 Provides better solution than this PR, which should be removed once the other one is merged.

This PR is a temporary solution only providing the following changes:

### TCP Source:
* Add reconnect timeout to ensure reconnect function will be called when server is physically reset.
* Move flushing internal buffer check from consume method to load method making it more reliable since this is where more bytes will be added to the buffer.

### Parse Errors:
* Avoid data loss on parse errors by dropping the bytes one by one instead of dropping all the loaded bytes.
* Terminate the session when first parse call returns parsing error assuming that users have chosen a file with unsupported format for the selected parser.

### Use `dlt-core` from Master:
 * Reference `dlt-core` from its master branch on GitHub since it contains needed fixes that aren't published on crates.io yet


AmmarAbouZor:
Done in #2256

PR #2256, opened 2025-03-13
Parser: Multiple Return values & Producer Loop Improvements

AmmarAbouZor:
This PR is a subset of #2244 without TCP reconnecting part

It provides the following changes:

### Parsers General:

### Producer Loop:

* Current unit tests have been adjusted with the new producer API, and they all pass without any change in their core functionality. 
* Optimize handling parse error by avoiding dropping all loaded bytes. Instead, it will skip them one by one and reapply parse to avoid data loss.
* Abort the session on continues initial parse errors for the first megabyte of input bytes to avoid calling parse repeatedly on each byte drop of the source in case the input isn't suitable for the selected parser.
* Handling incomplete parse error is improved by Skipping one byte if load function didn't deliver any more bytes instead of terminating the whole session.
* Extend unit tests for producer with mocks to cover parse error cases.
* Extend unit tests for cancelation safety of producer loop covering all parser error cases.

### TCP/UDP Sources:
* Flushing the internal buffer is moved to `load()` method to tie freeing bytes in internal buffer and loading it with more data, making it more reliable to avoid potential bugs when having an incomplete parser result on last packet before having the internal buffer almost filled.

### SomeIP parser:
Fix converting errors from SomeIP crate into parser, ensuring `Incomplete` error type is delivered correctly in producer loop since this error will be handled different than parse errors.

## Performance Changes:
```md

### Current PR:
File Read Took: 12058 ms
File Read Took: 11972 ms
File Read Took: 12103 ms

### Master: 
File Read Took: 12648 ms
File Read Took: 12639 ms
File Read Took: 12680 ms

### Current PR:
File Read Took: 3509 ms
File Read Took: 3461 ms
File Read Took: 3538 ms

### Master: 
File Read Took: 3518 ms
File Read Took: 3485 ms
File Read Took: 3507 ms

### Current PR:

### Master: 
```


DmitryAstafyev (approved):

application/apps/indexer/parsers/src/dlt/mod.rs
```diff
@@ -173,48 +198,56 @@ impl<'m> Parser<FormattableMessage<'m>> for DltParser<'m> {
...
-                    end: self.offset + c as usize,
-                },
-            }
-        });
-        let total_consumed = input.len() - rest.len();
-        let item = (total_consumed, msg.map(|m| m.into()));
-
-        Ok(iter::once(item))
+        parse_all(input, timestamp, MIN_MSG_LEN, |input, timestamp| {
+            self.parse_item(input, timestamp)
```
marcmo, line 238: since we always want to parse_all in the parser, should we not make `parse_item` the function of our Parser trait?

application/apps/indexer/sources/src/socket/mod.rs
```diff
@@ -71,3 +72,34 @@ pub enum ReconnectStateMsg {
...
+enum BuffCapacityState {
+    /// Buffer can be loaded with more data.
+    CanLoad,
+    /// Buffer is almost full. Loading it may cause data loss.
+    AlmostFull,
+}
+
+/// Checks if the buffer can be loaded with more data, calling flush on it when necessary,
+/// and return the buffer capacity state.
+fn hanlde_buff_capacity(buffer: &mut DeqBuffer) -> BuffCapacityState {
```
marcmo, line 87: spelling: handle_buff_capacity

do we have test coverage for this code?

marcmo (approved):
LGTM...nice work @AmmarAbouZor 
pls take a look at my comments

application/apps/indexer/parsers/src/dlt/mod.rs
AmmarAbouZor: With forcing future parsers to implement `parse_item()`, we will be taking the opportunity of parsing items in parallel for them. I think we may need to have more thoughts about this. 
I'll create an issue to rethink the parser trait and to see if we can change the implementation of the current parser to introducing parsing in parallel there as well.

application/apps/indexer/sources/src/socket/mod.rs
AmmarAbouZor: Yes, it's tested in `test_source_buffer_overflow()` in both TCP and UDP

AmmarAbouZor:
Handling incomplete parse error is improved by Skipping one byte if load function didn't deliver any more bytes instead of terminating the whole session.

application/apps/indexer/parsers/src/dlt/mod.rs
AmmarAbouZor: Issue #2260 is created for this change

PR #2262, opened 2025-03-18
TCP: Reconnect with Cancel-Safety & Use `keepalive` + Chipmunk CLI

AmmarAbouZor:
This PR provide the following:

* Reimplement the reconnect function with cancel-safety in mind. The new implementation will spawn reconnecting in a new task keeping track on each progress.
* Possibility to configure `keepalive` on TCP source + CLI arguments for its configuration on Chipmunk CLI, keeping the implementation on Chipmunk GUI for later.
* Unit tests for reconnect and cancel-safe load and reconnect calls.
* Change duration units on Chipmunk CLI from milliseconds to units since seconds accuracy is more suitable for this domain.
* Increase Chipmunk CLI version + Update readme and changelog

**Note:**
Manual timeout checks aren't implemented here, because they would increase the code complexity and introducing `Arc<RwLock<>` wrappers around the tcp sockets even though that this option may not ever be needed. However, I still have a prototype in a different branch in case we find out that is needed

### DIFF
--- a/application/apps/indexer/sources/src/socket/tcp/reconnect.rs
+++ b/application/apps/indexer/sources/src/socket/tcp/reconnect.rs
@@ -88,13 +88,13 @@ impl TcpReconnecter {
             self.task_handle.is_none(),
             "There must be no spawned reconnect task when spawn reconnect is called"
         );
-        let hanlde = tokio::spawn(reconnect(
+        let handle = tokio::spawn(reconnect(
             self.reconnect_info.clone(),
             self.binding_address,
             self.keep_alive.clone(),
         ));
 
-        self.task_handle = Some(hanlde);
+        self.task_handle = Some(handle);
     }
 }
 
@@ -139,7 +139,7 @@ async fn reconnect(
                         // Make sure the message has been sent before returning.
                         yield_now().await;
                     }
-                    log::warn!("Reconnecting to TCP server failed after {attempts} attemps.");
+                    log::warn!("Reconnecting to TCP server failed after {attempts} attempts.");
 
                     return ReconnectResult::Error(err);
                 }
--- a/application/apps/indexer/sources/src/socket/tcp/reconnect.rs
+++ b/application/apps/indexer/sources/src/socket/tcp/reconnect.rs
@@ -1,0 +1,151 @@
+use std::{net::SocketAddr, time::Duration};
+use tokio::{
+    net::TcpStream,
+    sync::watch,
+    task::{yield_now, JoinHandle},
+};
+
+use super::KeepAliveConfig;
+
+#[derive(Debug, Clone)]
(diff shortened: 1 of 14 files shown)


application/apps/indexer/sources/src/socket/tcp/reconnect.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+async fn reconnect(
+    reconnect_info: ReconnectInfo,
+    binding_address: SocketAddr,
+    keep_alive: Option<KeepAliveConfig>,
+) -> ReconnectResult {
+    if let Some(sender) = &reconnect_info.state_sender {
+        sender.send_replace(ReconnectStateMsg::Reconnecting { attempts: 0 });
```
marcmo, line 115: maybe not necessary to send the initial message

application/apps/indexer/sources/src/socket/tcp/reconnect.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    let mut attempts = 0;
+    loop {
+        attempts += 1;
+        if let Some(sender) = &reconnect_info.state_sender {
+            sender.send_replace(ReconnectStateMsg::Reconnecting { attempts });
+        }
+        log::info!("Reconnecting to TCP server. Attempt: {attempts}");
+        tokio::time::sleep(reconnect_info.interval).await;
```
marcmo, line 128: maybe move after attempt

DmitryAstafyev (approved):
Approved on my side because it's needed ASAP for CLI tool. 

But I just want to leave my opinion, which I've already shared during the discussion with you and @marcmo: I still believe - reconnection should be implemented on the upper level. Yes, the source should track the connection state, but losing of connection is actually an error and the source should just drop this kind of error outside. The decision of reconnection is a decision of the upper layers (client) and the client should trigger it.  

With the current solution, we are introducing "hidden" states of source - the client doesn't see the process of reconnection, and cannot count attempts; at the end - the user doesn't have normal control.

application/apps/indexer/sources/src/socket/tcp/reconnect.rs
```diff
@@ -0,0 +1,151 @@
...
+    ///
+    /// # Panics:
+    ///
+    /// This function panics if there is an already spawned task which hadn't been consumed yet.
+    pub fn spawn_reconnect(&mut self) {
+        assert!(
+            self.task_handle.is_none(),
+            "There must be no spawned reconnect task when spawn reconnect is called"
+        );
+        let hanlde = tokio::spawn(reconnect(
```
marcmo, line 91: spelling

AmmarAbouZor:
Adjustments after review are done :+1:

PR #2267, opened 2025-03-19
Draft1: Cancel Safety for SDE Communications

### LINKED ISSUE #2257: Producer loop: SDE Cancel safety.
The current implementation of `SDE` messages in producer loop won't be cancel-safe anymore since we are removing wrapping the function inside a stream.

Changes can be eighter directly in producer loop or by moving the implementation to the corresponding byte-sources which supports this functionality.

AmmarAbouZor:
This PR closes #2257

This PR tries to ensure cancel safety for SDE communications while keeping it inside the producer loop.
To achieve that:
* SDE communication is moved to the load method, 
* SDE channel will be borrowed instead of the current take() and reassign logic.
* We will keep state of the received msg in case income() method got cancelled.
* Writing to serial ports will block the current thread unit it's done to prevent interruption while writing.

This implementation force incomplete function to be cancel safe, but keep the SDE communications inside the producer loop. However, it would be better to move SDE communications to the session to avoid forcing `income()` method on byte source to be cancel safe.

### DIFF
--- a/application/apps/indexer/sources/src/command/process.rs
+++ b/application/apps/indexer/sources/src/command/process.rs
@@ -215,7 +215,7 @@ impl ByteSource for ProcessSource {
 
     async fn income(
         &mut self,
-        request: stypes::SdeRequest,
+        request: &stypes::SdeRequest,
     ) -> Result<stypes::SdeResponse, SourceError> {
         let bytes = match request {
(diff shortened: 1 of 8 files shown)


AmmarAbouZor:
Closed in favor of #2269

PR #2269, opened 2025-03-20
Move SDE communication to observe session.

### LINKED ISSUE #2257: Producer loop: SDE Cancel safety.

AmmarAbouZor:
This PR close #2257 

SDE communications are special case that are used in observe sessions only and can be moved to there since producer isn't locked with stream interface anymore. 
This change won't force cancel safety on income method, and ensure that writing to serial ports byte by byte won't be interrupted.

### DIFF
--- a/application/apps/indexer/addons/dlt-tools/src/lib.rs
+++ b/application/apps/indexer/addons/dlt-tools/src/lib.rs
@@ -42,7 +42,7 @@ pub async fn scan_dlt_ft(
                 None,
                 with_storage_header,
             );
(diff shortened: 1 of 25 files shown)


DmitryAstafyev (approved):

marcmo (approved):
nice refactoring!

PR #2275, opened 2025-03-25
Plugins Manager: Implementing Caching to Avoid Loading Plugins on each Start 

### LINKED ISSUE #2206: Cache support for Plugins Manager
Currently, plugins manager scans plugins directory on app start loading each plugin to get its metadata and configurations schema and other needed data. This could affect startup time in case of many plugins, even if users won't be using those plugins.
caching the plugins info in a file while keeping track on the changes in plugins will avoid loading the plugins on each start up.

**Note**: This has lower priority than the other tasks in Step 1

AmmarAbouZor:
This PR closes #2206 

It provides:
* Implement caching plugins states instead of loading the plugins on each app start.
* Cache will keep track on binary files changes by comparing hashes of the binaries.
* Add dependency on `dir_checksum` crate from development-CLI.
* Refactoring for loading plugins functions improving code quality and moving functions around to share them among different parts in the codebase.
* Plugins manager won't stop on errors while loading cached infos. In this case the error will be logged (in development printed to stderr) since it will be overwritten with the infos after loading the plugins


application/apps/indexer/plugins_host/src/plugins_manager/paths.rs
```diff
@@ -0,0 +1,82 @@
+use std::path::{Path, PathBuf};
+
+/// The name of Chipmunk directory inside home directory.
+const CHIPMUNK_HOME: &str = ".chipmunk";
+
```
DmitryAstafyev, line 5: I think we should have it somewhere in the scope of  indexer/session. If not, it would be nice to introduce as a global `const`, but not related to plugins only. It's  some kind of `TODOs` for the future.

AmmarAbouZor: The same function and const is defined in `sessions` crate as well, which I can't be referenced form the plugin library because of circular references.
We may need to introduce a new library for shared const and functions between the other library without any dependency on them so it can be among all other libs in Chipmunk.

PR #2276, opened 2025-03-26
Removing observe options from all levels upper indexer

DmitryAstafyev:
## Problematic

Currently, in `platform` we have a detailed description of the `observe` object. It's more than just a description of the interface of settings. Actually, the major issue is:
- The client knows which kind of sources/parser are available, but it isn't confirmed by core (theoretically some source/parser can be unavailable) 
- The client knows the options (settings) of each parser/source, but isn't confirmed by core (theoretically, some settings might be changed)
- binding of TS types with core types is strong, but doesn't give 100% of match. For example, in the scope of validation, a client has own logic, core - own one. 

In the context of a desktop application (without plugins), it's not critical. But it become important if:
- chipmunk as a remote service. Actually, in this situation client has no idea about the  configuration of the backend service (is DLT still included, for example, or pcapng source, which is useless for remote service)
- plugins. In the context of plugins, a client has to "ask" core list of plugins and its settings/options.

## Possible solution
In the same way as with plugins, a client should "ask" core about available sources/parsers. As soon as a client has a list, a client can request a scheme of options/settings. In short it might look like:
1. Client requests list of parsers/sources
2. User selects an `origin` (origin of source: file(s), folder(s), stream).
3. Client request options scheme based on `origin` 
4. The client validates options by requesting to core (no client logic at all)
5. The client checks the compatibility of pair `parser` and `source` with core, but not based on his own logic.


AmmarAbouZor (commented):
Very nice work @DmitryAstafyev 
I have one notice about using tasks as data in rust in scopes that expected to live for long time. 
Although modelling the solution with this approach is attractive in other programming languages for its simplicity, I don't think this will be work well in Rust because it has really strict rules regarding lifetimes and concurrency. We already can sense that in the prototype here by enforcing encapsulating the future inside Pin<Box<>> and adding `Send` bound and I think it would make more problem once real lazy loading is involved.

I would suggest a different approach here using enums and match statements. For example for parser we can define the enum 
```rust
pub enum ParserType {
    Dlt,
    SomeIp,
    Plugins(String)
}
```
Then the component will have a list for parsers with their UUID and parser type. Then we we get a request we can match on the enum types and call the corresponding mathod on the needed parser. Something like:
```rust
impl Components {
    async fn lazy(ident: Ident) {
        let parser_type: ParserType = self.parsers.get(iden);
        match parser_type {
            ParserType::Dlt => parser::Dlt::get_lazy_field().await,
            ParserType::SomeIp => parser::Dlt::get_lazy_field().await,
            ParserType::Plugins(name) => plugin_reg.get(name).get_lazy_field().await,
        }
    }
}
```
I think this approach is more suited since it doesn't bother with the lifetimes of the functions and what they capture and it doesn't require the function to be send or have any other constrains we may encounter in the future.

Issue #2223, opened 2025-02-13
Move Plugin manager from unbound session


AmmarAbouZor:
Plugins-manager is currently a part of unbound session, but with it introducing a growing amount of APIs functions which are related to plugins only, it would better to have the plugin manger as a stand-alone service.

AmmarAbouZor:
Unbound session is not used anymore and plugins manager is initialized correctly in Chipmunk 4

PR #2277, opened 2025-03-28
Plugins: Add and Remove Functions

### LINKED ISSUE #2208: Plugins Manager: Api for adding and removing plugins
On backend we need to provide APIs to add and remove plugins with the following points:
* **Add**: Add function with the path for the directory of the plugin and the plugin type. This function should validate the directory then copy it to the matching plugins directory in chipmunk home-directory. If the plugin isn't valid it should return an error directly.
* **Remove**: Function with the path for the plugin directory. This function will return an error if the plugin doesn't exist. Otherwise, it will remove the plugin directory completely.

AmmarAbouZor:
This PR closes #2208 

It provides:
* Implement adding and removing plugins in plugins manager.
* Basic implementation for add and remove in plugin manager UI.
* Provide the option to infer plugin type from the binary file.
* Change field in plugin metadata from `name` to `title` emphasizing that this field is used as display name for the users and doesn't match the plugin name for the application. 
* Refactoring & Renaming & Typos.

After this PR is merged we will need some adjustments on the UI, which will be provided as a separate issue

### DIFF
--- a/application/apps/indexer/plugins_host/src/plugins_manager/mod.rs
+++ b/application/apps/indexer/plugins_host/src/plugins_manager/mod.rs
@@ -260,6 +260,7 @@ impl PluginsManager {
                             PluginLogLevel::Debug | PluginLogLevel::Info => false,
                         })
                         .for_each(|msg| {
+                            // Writing to a string never fails.
                             _ = writeln!(&mut err_msg, "{}", msg.msg);
                         });
 
--- a/application/apps/indexer/plugins_host/src/plugins_manager/mod.rs
+++ b/application/apps/indexer/plugins_host/src/plugins_manager/mod.rs
@@ -10,13 +10,16 @@ mod tests;
 
 use std::path::{Path, PathBuf};
 
+use crate::plugins_shared::load::{load_and_inspect, WasmComponentInfo};
 use cache::CacheManager;
+use load::{load_plugin, validate_plugin_files, PluginEntityState, PluginFilesStatus};
+use paths::extract_plugin_file_paths;
 use stypes::{
     ExtendedInvalidPluginEntity, ExtendedPluginEntity, InvalidPluginEntity, PluginEntity,
-    PluginRunData,
+    PluginLogLevel, PluginRunData, PluginType,
 };
 
-pub use errors::{InitError, PluginsCacheError};
+pub use errors::{PluginsCacheError, PluginsManagerError};
 
 /// Plugins manager responsible of loading the plugins, providing their states, info and metadata.
 #[derive(Debug)]
@@ -28,7 +31,7 @@ pub struct PluginsManager {
 
 impl PluginsManager {
     /// Load plugins from their directory.
-    pub async fn load() -> Result<Self, InitError> {
+    pub async fn load() -> Result<Self, PluginsManagerError> {
         let mut cache_manager = match CacheManager::load() {
             Ok(manager) => manager,
             Err(err) => {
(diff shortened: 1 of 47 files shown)


application/apps/indexer/plugins_host/src/plugins_manager/mod.rs
```diff
@@ -156,4 +159,166 @@ impl PluginsManager {
...
+    /// Adds a plugin with the given directory path and the optional plugin type.
+    ///
+    /// * `plugin_dir_src`: Path of the plugin directory to be copied into chipmunk plugins directory.
+    /// * `plugin_type`: Type of the plugin, when not provided plugin type will be entered from plugin
+    ///   `WIT` signature in its binary file.
+    pub async fn add_plugin(
+        &mut self,
+        plugin_dir_src: PathBuf,
+        plugin_type: Option<PluginType>,
+    ) -> Result<(), PluginsManagerError> {
```
DmitryAstafyev, line 172: I am not sure, we need install/uninstall functions to be async. Even if we need to leave it as async (because of some API inside), I'm not sure about `tokio::fs`. I just want to say: installation of plugin isn't something what we want to make in really async way. Even more, if at some time we would allow users, for example install multiple plugins, we would not do it in parallel but sequentially (because it is always safer). 

Well up to you, but I just don't see any needs to use async `tokio::fs` in that context.

application/apps/indexer/plugins_host/src/plugins_manager/mod.rs
```diff
@@ -156,4 +159,166 @@ impl PluginsManager {
...
+                    invalid_entity
+                        .run_data
+                        .logs
+                        .into_iter()
+                        .filter(|msg| match msg.level {
+                            PluginLogLevel::Err | PluginLogLevel::Warn => true,
+                            PluginLogLevel::Debug | PluginLogLevel::Info => false,
+                        })
+                        .for_each(|msg| {
+                            _ = writeln!(&mut err_msg, "{}", msg.msg);
```
DmitryAstafyev, line 263: is here an error suppression? yes, probably no chance to fall down into error, but maybe just use some other way? or collect Result as well, but unwrap it after ?

application/apps/indexer/session/src/unbound/api.rs
```diff
@@ -307,4 +307,35 @@ impl UnboundSessionAPI {
...
+        plugin_path: String,
+    ) -> Result<stypes::CommandOutcome<()>, stypes::ComputationError> {
+        let (tx_results, rx_results) = oneshot::channel();
+        self.process_command(
+            id,
+            rx_results,
+            Command::RemovePlugin(plugin_path, tx_results),
+        )
+        .await
+    }
```
DmitryAstafyev, line 340: we add more and more into unbound session. On step 3, I think, it makes sense to exclude from unbound session everything related plugins to own plugins controller.

DmitryAstafyev (approved):

application/apps/indexer/session/src/unbound/api.rs
AmmarAbouZor: We have already an issue for that #2223
I don't now when it would be a good time to implement this. (If before merging the plugin into master or after)

application/apps/indexer/plugins_host/src/plugins_manager/mod.rs
AmmarAbouZor: Writing to string should actually never fail, but the signature of `writeln!()` returns a result since we can write to a file or a stream. 
But you are right I need to provide a comment here for clarity :+1:

application/apps/indexer/plugins_host/src/plugins_manager/mod.rs
DmitryAstafyev: yes, that makes sense. I've missed that moment

Issue #2218, opened 2025-02-13
Plugins: Provide Tests & Benchmarks for the plugins


AmmarAbouZor:
We need to have function to run tests on plugins without having to run it inside Chipmunk application. For testing we need tests that accepts the path for the plugin path, its serialized configurations and the path for its input source.
The same will be needed for benchmarks as well.

These tools are essentials once we introduce a distribution system to ensure the plugins are working before distributing them and would be a great help for plugin developers

PR #2280, opened 2025-03-28
Plugins: Deliver Plugin Path and Configuration via toml File

AmmarAbouZor:
This PR solves one task from #2218 

It includes:
* Plugins path and configurations can be provided to plugins benchmarks via toml config files.
* Provide benchmarks config files for all example plugins and template.
* Extend plugins readme file with infos about benchmarking.

### DIFF
--- a/application/apps/indexer/plugins_host/src/plugins_manager/cache/mod.rs
+++ b/application/apps/indexer/plugins_host/src/plugins_manager/cache/mod.rs
@@ -102,7 +102,6 @@ impl CacheManager {
         &mut self,
         plug_dir: &Path,
         state: CachedPluginState,
-        persist: bool,
     ) -> Result<(), PluginsCacheError> {
         let calculated_hash = self.calc_plugin_hash(plug_dir)?;
 
@@ -121,10 +120,6 @@ impl CacheManager {
             }
         };
 
-        if persist {
-            self.persist()?;
-        }
-
         Ok(())
     }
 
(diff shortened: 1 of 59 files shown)


application/apps/indexer/plugins_host/src/plugins_manager/cache/mod.rs
```diff
@@ -145,6 +140,29 @@ impl CacheManager {
...
+    /// Remove plugin from cache registry of exists with the option to persist the current cache.
+    pub fn remove_plugin(
+        &mut self,
+        plugin_dir: &Path,
+        persist: bool,
+    ) -> Result<(), PluginsCacheError> {
+        let Some(plugin_idx) = self
+            .plugins_info
+            .iter()
+            .position(|plug| plug.plugin_dir == plugin_dir)
```
DmitryAstafyev, line 152: BTW, maybe some kind of method will make this check more self-explanatory 
```
position(|plug| plug.is_path(plugin_dir))
// or
position(|plug| plug.is_located(plugin_dir))
```

DmitryAstafyev (approved):

application/apps/indexer/plugins_host/src/plugins_manager/cache/mod.rs
AmmarAbouZor: We are looking here currently for a plugin with the exact directory path as the provide one, so the method name could be something like `is_same_path(plugin_dir)` which almost the same as `plug.plugin_dir == plugin_dir`.
Maybe if we add more conditions in the futures then it would make sense to have them in a self-explanatory method instead of multiple conditions :+1:

PR #2287, opened 2025-04-03
Update Plugins Manager (close #2278)

### LINKED ISSUE #2278: Plugins Manager: Improve UI for adding and Removing Plugins
In #2277, we've provided add and remove plugins functionality in core, the communication API and with basic implementation in the UI.
However, the UI is still primitive and needs further improvements:

* UI doesn't get updated after adding or removing plugins.
* Errors while adding and removing plugins should be shown to the users instead of logging them into console only (current state)
* It would be nice to have an indicator while a plugin is being added or removed.
* Plugins logs doesn't update when current valid plugin is changed. It works however for invalid plugins.

@DmitryAstafyev Sure you can provide more nice UI touches and adjust the suggestion if needed 😉

DmitryAstafyev:


AmmarAbouZor (approved):
@DmitryAstafyev Thanks a lot for the fixes.
It seems that one of the CI checks are failing

PR #2288, opened 2025-04-03
Add Experimental Plugin Support with WASM & WASI

### LINKED ISSUE #1984: Plugin mechanism for compiled rust code
we need to be able to link precompiled static libraries.

- [ ] libs need to be pre-built for different architectures
- [ ] should allow to pull in closed source functionality

First example will be the MDF library

AmmarAbouZor:
This PR closes #1984 

This PR provides a complete system for supporting plugins using [WASM](https://webassembly.org/) and [WASI](https://wasi.dev/) and the [component model](https://component-model.bytecodealliance.org/).

This PR provides full support to **Parser Plugins** and initial implementation for **ByteSource Plugins**

Very thanks for @DmitryAstafyev For the great support and the UI implementation on this PR :heart: 

This PR includes the following:
## Plugins API:
* API Types and definitions are documented via [WIT (Wasm Interface Type](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md).
* The version of the API is provided within the `WIT` files themselves.
* Plugins must provide the configuration schema next to multiple infos as method defined in the `WIT` file itself.
* Plugins can provide their metadata (Title and description) via `toml` file.
* Plugins can provide a markdown `README` file that will be rendered in Chipmunk GUI. 
* We've provided an extra crate `plugins_api` for plugins developers in Rust including the generated types from `WIT` files besides multiple useful functions + The export macro to generate Wasm types.
* Multiple examples for parser plugins are provided.
* Template for developing parser plugin is provided.
* Documentation for plugins development is provided as separate README file besides the documentation of the crate `plugins_api`

----
## Plugins Host:
* Chipmunk core will be shipped with [wasmtime](https://github.com/bytecodealliance/wasmtime) runtime to compile and run the provided plugins.
* Plugins API infos (Interface name and Version) will be read from WASM files directly and validated upfront.
* Versioning system is built-in to ensure full control of supporting multiple API Versions.
* Plugins host will cache infos provided by the plugins wasm file like their config schemas and rendering options to avoid loading the plugins on each application start.
* Plugins will have access to stdout, stderr, read access to file and directories which are provided from users via the config schema.
* Plugins has Write access to a temporary directory that will be destroyed by the end of each session.
* Plugins can be added manually to chipmunk home directory or via the Plugins manager in the UI

----
## Plugins UI:
* Plugins can be managed via plugins manager in the UI.
* Parser plugins can be selected with the same dropdown as the native plugins.
* Parsers configuration schemas will be rendered in the UI enabling users to provide the value of each configuration.
* Validation of the configuration is currently static (no live validation)

----
## Benchmarks:
Performance difference between plugins and native parsers are about 10%.
Here is a table comparing native DLT parser and a plugin DLT parser with time needed for parsing session with files of sizes 500MB and 400 MB

| Parser  | File 500MB (ms) | File 400MB (ms) |
|---------|---------------|---------------|
| Native  | 12,181       | 10,990       |
| Plugin  | 14,747       | 11,842       |
| Diff      |  2.5 secs     | 0.9 secs     |


DmitryAstafyev (approved):

application/apps/indexer/plugins_host/src/bytesource_shared/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+        let (component, version) = Self::load(&plugin_path).await?;
+
+        match version {
+            SemanticVersion {
+                major: 0,
+                minor: 1,
+                patch: 0,
+            } => {
+                let source = v0_1_0::bytesource::PluginByteSource::initialize(
+                    component,
```
marcmo, line 100: versions are defined in multiple places...if possible just use 1 central place and reuse

application/apps/indexer/plugins_host/src/bytesource_shared/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+impl WasmPlugin for PluginsByteSource {
+    fn get_type() -> PluginType {
+        PluginType::ByteSource
+    }
+
+    async fn plugin_version(&mut self) -> Result<SemanticVersion, PluginError> {
+        match &mut self.source {
+            PlugVerByteSource::Ver010(source) => source.plugin_version().await,
```
marcmo, line 149: same thing...maybe we can use a newtype somewhere for that...this line should not change for a new version

application/apps/indexer/plugins_host/src/plugins_manager/cache/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub fn load() -> Result<Self, PluginsCacheError> {
+        let cache_path = Self::cache_file_path()?;
+
+        if !cache_path.exists() {
+            return Ok(Self::default());
+        }
+
+        let cache_file = File::open(cache_path)?;
+
+        serde_json::from_reader(cache_file).map_err(|err| err.into())
```
marcmo, line 49: why do we use json here and not binary?

AmmarAbouZor: I think cache file won't be too large to make a difference in performance. With that I chose JSON for the following reasons:
* We still can read the cache infos by opening the file, in case something went wrong and we want to debug.
* We already have `serde_json` dependencies in the project and we don't need to include new one for binary serialization.
* Serde_json is more forgiving for missing infos and it's more battle tested since it's used much more that binary serializers

application/apps/indexer/plugins_host/src/bytesource_shared/mod.rs
AmmarAbouZor: I defined a constant `SematicVersion::V0_1_0` instead of the hard-coded ones. This would make the code a little bit cleaner here, but I don't think we can avoid the match statements because we are using enums for versions management.

Another approach would be using generics by defining a trait for all parser plugins of each version, then we can make the general parser plugin generic over the trait. I can try to implement this approach, but I don't think it would make the code cleaner + We need to force plugins in all versions to implement a trait that we need to define.

However, I can make a prototype of this change and do some measurements in a couple of hours if needed.

application/apps/indexer/plugins_host/src/bytesource_shared/mod.rs
AmmarAbouZor: We could change that with generics as explained in the comment above

application/apps/indexer/plugins_host/src/v0_1_0/parser/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+        &mut self,
+        input: &[u8],
+        timestamp: Option<u64>,
+    ) -> Result<impl Iterator<Item = (usize, Option<p::ParseYield<PluginParseMessage>>)>, p::Error>
+    {
+        let call_res = block_on(self.plugin_bindings.chipmunk_parser_parser().call_parse(
+            &mut self.store,
+            input,
+            timestamp,
+        ));
```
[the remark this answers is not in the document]
AmmarAbouZor: Interesting [link](https://users.rust-lang.org/t/simplest-possible-block-on/48364/33)

PR #2293, opened 2025-04-11
Remove `humantime` crate dependency

### LINKED ISSUE #2255: Replace `Humantime` crate
The crate [humantime](https://github.com/tailhook/humantime) seems to be unmaintained, and it would be good to replace it with our own implementation since we are using one function from it only `humantime::format_rfc3339()`

AmmarAbouZor:
This PR closes #2255 

It replaces the dependency to the unmaintained crate `humantime` with the built-in function in `chrono` crate.

It includes:
* Use built-in time formatter from chrono library instead of the unmaintained `humantime` crate.
* Using the new format method discovered a bug in formatting the time when there are no fractional seconds making the width of the output not always the same.
* Snapshot test results adjusted after resolving the bug.

I've compared the output of parsing two DLT files between current master and this PR without having changes in them


DmitryAstafyev (approved):

marcmo (approved):
LGTM

PR #2304, opened 2025-05-05
Development CLI: Check for tools versions & Enable `corepack` & Option to Skip env checks

### LINKED ISSUE #2303: Development CLI: Include tools minimum versions in environment checks
The current environment check performed by the development CLI tool only verifies that required tools are installed. It does not ensure that the installed versions meet the minimum requirements necessary for developing Chipmunk. 

To solve this, we need to extend the environment checks to maintain and verify against a list of the minimum required versions for each tool.

This version checking approach will work for most prerequisites. However, it requires special handling for `yarn`. Instead of checking the globally installed `yarn` version, the check should ensure that `yarn` is correctly set up to use the specific version defined in the project's configuration (e.g., managed via `Corepack`).

AmmarAbouZor:
This PR closes #2303 

It provides:
* Check for minimum needed versions for external tools used in Chipmunk development. The versions are set in the file `miv_versions.toml` within the `development-cli` directory in Chipmunk repo.
* ~Enable `corepack` in environment tools checks, since it's used to set yarn version.~
* Add a note to contribution.md page mentioning that `corepack` must be enabled to set the correct `yarn` version.
* Option to skip environment checks via CLI argument or User configurations since it may be not needed for developers working day-to-day on Chipmunk

@DmitryAstafyev Could you please have a look into the minimum versions in the file `min_versions.toml`


AmmarAbouZor:
Thanks @DmitryAstafyev for the reviews and the ideas.
I think I've misled you by setting too recent versions of the needed tools so you thought that those are the suggested versions and not the minimum required versions.
What I mean with those versions are actually that those are the minimum versions in order for Chipmunk to compile like if you set the min version for `nj-cli` to `0.4.4` for example, this will mean that the compilation will not work with the version `0.4.3` at all.\
The purposes for those checks are:
* Provide a meaning full error message to the users
* Do the checks upfront so the build won't even start when we are sure that it will error down the way.

Regarding maintaining this, I think the Maintenace would be simple, since we will update it only when we get into an issue where a version is out-dated and we must update the current version. In that case it would easier to update the config file here instead of updating the contribution readme file and broadcast the news to all developers that we know.

I'll set Node version to 20.08 (LTS) since it's the first node version that will be shipped with `corepack`, and I will run another tool to determine which rust version is the minimum one on our current codebase. However we still need to know the minimum `npm` version (I suppose I can set it to 10.0.0)

Determining node and other Javascript tools dynamically will be great if we are able to do that, but we can't do that for other tools like `wasm-tools` and `nj-cli` since they will just output really unhelpful messages when their version isn't supported anymore.

DmitryAstafyev (approved):

AmmarAbouZor:
Updates:
- I've removed `corepack enable` command because it may need administrator permission to run then added a note to the `contribution.md` file mentioning that we needed to be enabled to determine the current `yarn` version.
- I set the min Rust version to `1.81.0` since it's the latest version.
- I set `npm` to `10.0.0` and node to `20.9.0`

PR #2332, opened 2025-06-04
Core: Rust Edition 2024 & Remove `Cargo.lock` files

### LINKED ISSUE #2311: Update to Rust Edition 2024
It's a good practice to stay with the latest rust edition since it has been released for a couple of months now and multiple crates has moved to it, making any updates for our dependencies more complicated.

This update will include updating rust edition in multiple apps:
- [ ] **Rust Core in Chipmunk:** This will be the most complicated one because of the large codebase and that we will need a suitable time to avoid having a lot of merge conflicts on the different branches. This could be done after merging the plugins PR
- [ ] **Development CLI:** This would be the easiest one to start with because it doesn't have direct dependencies to rust core.
- [ ] **Chipmunk CLI:** It has smaller codebase but also it has a dependency to rust core
- [ ] **Plugins API & Examples:** They are also stand-alone codebases that can be upgraded easily

AmmarAbouZor:
This PR closes #2311 

It provides the following:
* Upgrade all crates in rust core to edition 2024, providing all the formatting and clippy fixes.
* Remove `Cargo.lock` Files from crates of type libraries since those aren't useful besides producing developer pain while resolving merge conflicts!
* Update dependencies for multiple crates to make them compatible with rust edition and up to date.
* Remove resolver 2 from cargo since rust edition 2024 introduces resolver 3 [link](https://doc.rust-lang.org/edition-guide/rust-2024/cargo-resolver.html)

I've ran all the test and proptests, tested chipmunk on my machine locally without any problem, but merging this PR with some time before the next release would give us more time to test it.

@DmitryAstafyev The applied changes on the updater target needs extra tests since it was not updated since rust edition 2018


DmitryAstafyev:
> @DmitryAstafyev The applied changes on the updater target needs extra tests since it was not updated since rust edition 2018

Hello @AmmarAbouZor well in this case PR should be in draft state until tests are done. That's super trivial point:

- the latest release now is `3.18.1`
- set your local release to any less, ex. `3.18.0`
- build release version (!)
- run release version (!), not like `yarn run ...`, but just executre from `release` folder
- wait for the prompt to do an update
- update
- chipmunk will restart 
- check in About a version
- if you see `3.18.1` - all good. All rest cases - fail.

AmmarAbouZor:
Updater is tested on Windows and it's working fine. I'll move this issue to ready for review so it can be merged after the updater test on MacOS passes

PR #2333, opened 2025-06-05
Plugins: Configurations for Wasmtime Runtime Engine

### LINKED ISSUE #2324: Plugins: Explore Wasmtime configuration for Safety & Performance
Wasmtime Runtime API offers many ways to configure the runtime very precisely. 
Up until now, we are depending on the default configurations since wasmtime choose really sensible defaults. However, it would be useful to investigate in the configurations and set them explicitly to our needs while documenting these decisions in the code.

Our main targets:
- Ensure the memory consumptions of the plugins is limited to prevent malicious plugins from crashing the system.
- Check if it would be possible to gain more control over memory to improve performance and gain more control over sharing memory between the host and the plugins

AmmarAbouZor:
This PR closed #2324:

It includes:
* Set some configurations for wasmtime runtime as the default values to keep those values present in the source code with comment explaining why those configs are kept.
* Enable `shared everything threads` feature because for slightly performance gain, because it will provide optimizations for the component model.
* Keep comments in code about memory management and limitations.

### DIFF
--- a/application/apps/indexer/plugins_host/src/wasm_host/mod.rs
+++ b/application/apps/indexer/plugins_host/src/wasm_host/mod.rs
@@ -1,6 +1,6 @@
 use std::sync::{Arc, OnceLock};
 
-use wasmtime::{Config, Engine};
+use wasmtime::{Config, Engine, OptLevel, RegallocAlgorithm, Strategy};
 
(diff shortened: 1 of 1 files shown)


marcmo (approved):
LGTM

PR #2336, opened 2025-06-10
Transition to the Component-Based Model 

DmitryAstafyev:
These changes involve a fundamental shift in the application's architecture regarding embedded parsers and data sources.

## Problem Statement

Currently, every parser or source is represented as a variant in a common enum. As a result, all related types (such as settings, options, etc.) become part of the core application and propagate to higher layers. This significantly reduces flexibility. For example, adding or removing a parser/source requires modifying dozens of modules across the application.

### Solution (Basic Description)

- Introduce the concept of a “component,” which can represent either a parser or a data source (with potential for future extension).
- All fixed enums for parsers and sources are removed.
- Sessions are now created using a set of components.

### Outcome

- Parser and source implementations become isolated.
- Changes to any individual parser or source no longer affect the core or higher-level layers of the application.


application/apps/indexer/components/src/lib.rs
```diff
@@ -0,0 +1,386 @@
...
+/// # Notes
+///
+/// `Components` itself does not implement any teardown or destruction logic, as this is unnecessary:
+///
+/// - It only stores `ComponentDescriptor` instances, which are lightweight and do not own the actual
+///   component instances.
+/// - Component instances are created on demand and immediately passed upwards for further use; they
+///   are not retained by `Components`.
+///
+pub struct Components<S, P> {
```
marcmo, line 58: then please also name it `Registry`, not `Components`

PR #2354, opened 2025-06-24
Ensure sessions are using generics completely

AmmarAbouZor:
This PR includes:

* Move matching against parsers and sources to before starting sessions to make sure that producer sessions and making use of generics completely
* Rename parsers and sources enum temporally to avoid conflicts with stypes
* Include SDE in session
* TODOs in code

It also includes skipping CI checks for PRs which aren't targeting master branch

### DIFF
--- a/.github/workflows/pullrequest_check.yml
+++ b/.github/workflows/pullrequest_check.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   ts_and_rust_lint:
-    if: github.event.pull_request.draft == false
+    if: github.event.pull_request.draft == false && github.event.pull_request.base.ref == 'master'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -63,7 +63,7 @@ jobs:
         run: yarn run check
 
   integration_and_unit_tests:
-    if: github.event.pull_request.draft == false
+    if: github.event.pull_request.draft == false && github.event.pull_request.base.ref == 'master'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/application/apps/indexer/parsers/src/dlt/descriptor.rs
+++ b/application/apps/indexer/parsers/src/dlt/descriptor.rs
@@ -48,12 +48,12 @@ impl fmt::Display for StatFields {
     }
 }
 
-impl ComponentFactory<crate::Parser> for Descriptor {
+impl ComponentFactory<crate::AllParserTypes> for Descriptor {
     fn create(
         &self,
         origin: &SessionAction,
         options: &[Field],
-    ) -> Result<Option<crate::Parser>, NativeError> {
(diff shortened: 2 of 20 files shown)


AmmarAbouZor:
Thanks @DmitryAstafyev  for the review and the ideas I also find the target to make `session` crate abstracted as possible to make adding and removing components as simple as possible.

However, I find the restriction `We should be able to remove parsers or sources without changing a line of code in sources what soever` is leading us some anti-patterns as the one that I'm trying to solve here and other ones (more of that below). 
My suggestion is to relax this restriction to allow us matching against the enums defined elsewhere and pass the items when needed (Like with generics). This suggestion wouldn't change the complexity of adding and removing components because it will tiny amount of code with no logic and the compiler will tell us about any human mistakes because of rust exhaustive matches.

Also, I see `sessions` crate rather as "GUI App Controller" which is responsible of keeping track of the state of the GUI APP only and to do the communication between electron front end and the rust core (Parsers, sources, producer, plugins...). Hence, that `Chipmunk CLI` doesn't have any dependency to `session` crate. 

I'll list the anti-patterns that I've found so far, which can't be resolved without allowing us to match against enums in `session` crate:

### 1. Wrapper Enums instead of real generics in Producer (Manual Static Dispatch Anti-pattern)
The producer struct is designed to accept with generic `parser` and `byte-source`, and the current solution is giving them enums of all available parsers and byte-sources and it will match on each call of the traits `parser` and `byte-source`. This behavior is because of the restriction that we can't match against parsers and byte-source in `session` crate, and it's bad because of the following:
1. Losing Generics and optimization at compile time because the compiler won't generate separated code for each type, since we are giving the producer all the types only. This will prevent all kind of compiler optimization based on the types provided
2. Match branching on every function call instead of having one match branching at the start only. For example, the current solution will match against all parsers on each call on `parse()` method and will match on all byte-source on each `load()` and `consume()` methods, which will be called a lot during the sessions. My proposal here will call each match statement once.
3. With my proposal we will end up with less match statements (One per session instead one per trait function) which will make maintainability actually simpler than the current restriction

### 2. `Pin<Box<dyn Future<Output = ???> + Send>>` instead of enums for `LazyFieldsTask`
With the current restriction we are forced to use pin boxed dynamic futures for all components even if they are known at compile time. We could instead introduce enums for component types and match against them then calling the corresponded function on each component.
This would be better to avoid the dynamic dispatch where static dispatch can be used, allowing async function to use none-static items if possible (like plugins_manager).

Those are the anti-patterns that I've found for now and we would have potentially more problems in in further implementation in this refactoring or later on. That's way I wanted to raise a concern with this PR with a real example so we will be aware about the consequences of keeping the strict restriction as it is

AmmarAbouZor:
Closed in favor of #2357

PR #2357, opened 2025-06-26
[ShowCase] Performance improvements for Component based on Generics

AmmarAbouZor:
This PR is more a show case on some of the improvements that wouldn't be possible without having match statements inside `session` library.

The improvements include:
* Using static dispatch instead of dynamic dispatch for both parsers and bytesoruces in `Producer`
* Rollback `Parser` trait definitions and implementation, ensuring it will return a generic Iterator for various parsers

This PR includes other improvements & refactoring regarding `Writer` implementation and definitions


AmmarAbouZor:
Thanks @DmitryAstafyev  for your remarks. The changes in `Writer` are the results of having the parser returning a generic iterator which is the result of having generics instead of enums.
The changes in iterator enabled having closed closure in producer loop which lead to the changes in `Writer` to have one method to appended the messages to its internal buffer (Which should be asynchronous in any scenario) and another function to flush the intermediate buffer to the destination which can be async.

In any case, I agree that we don't need to discuss this further while other important parts (Like plugins) aren't implemented yet. I'll keep this issue as draft for reference while development and closed the other issue #2354

AmmarAbouZor:
This can be closed as we don't intend to continue working on the components anymore

PR #2358, opened 2025-06-27
Components: Parser definitions + Plugins + Producer and Writer Optimizations

AmmarAbouZor:
This PR provides the following:

### Parsers:
* Provide new parser definitions taken from the current master solution but with dynamic dispatch.

### MessagesWriter Changs & Performance Optimizations:
* Change method on Writer to `append()` which intended to append message to the internal memory of the writer without writing it to the underline target. This made this method sync and infallible 
* Optimize writer performance by reusing memory
* Remove redundant LogMesages enums: `Cow` already covers `String` and `&str` cases
* Refactoring by removing nested function.
* Performance improved by about 8-10% percent

### Others
* Improve error handling in producer loop.
* Remove internal buffer from producer.
* Skip CI checks for PRs not targeting master branch

### DIFF
--- a/application/apps/indexer/definitions/src/record.rs
+++ b/application/apps/indexer/definitions/src/record.rs
@@ -89,7 +89,7 @@ pub enum LogRecordOutput<'a> {
 
     /// Structured columnar data. Typically used when the parser can extract
     /// meaningful fields and present them as collection of texts.
-    Columns(Vec<Cow<'a, String>>),
+    Columns(Vec<Cow<'a, str>>),
 
     /// An attachment object, such as a binary blob or associated metadata.
     /// These are handled separately from textual data and require special treatment
@@ -101,23 +101,20 @@ pub enum LogRecordOutput<'a> {
     Multiple(Vec<LogRecordOutput<'a>>),
 }
 
-/// Interface for implementing a writer that handles log records.
+/// Defines an interface for buffering log records before they are persisted.
 ///
-/// An instance of `LogRecordWriter` is passed to a `MessageProducer`.
-/// Each time a new log record is available, the `MessageProducer` invokes the `write` method.
-/// The specific behavior for writing records to the actual destination is left to the implementation.
+/// An instance of a type implementing `LogRecordsBuffer` is passed to a `MessageProducer`.
+/// For each new log record, the `MessageProducer` calls the `append` method. The `flush`
+/// method is used to write the buffered records to their final destination.
 ///
-/// This allows implementors to define their own strategies for data caching, batching, and flushing.
-/// When the `MessageProducer` has no more data to process, it calls `finalize`. At that point,
-/// the writer should persist any buffered data, if applicable.
-///
-/// **Important:** (pending TODO) The `write` method must not be cancel-safe. This is required to keep
-/// the `MessageProducer` cancel-safe. This problems isn't resolved yet.
+/// This design allows for custom strategies regarding caching, batching, and flushing.
+/// When processing is complete, `flush` should be called to ensure all buffered data
+/// is persisted.
 ///
 /// # Examples
-/// An implementation might buffer records in memory and write them in bulk
-/// every few seconds, or directly write to a file, database, or network sink.
-pub trait LogRecordWriter {
+/// An implementation might buffer records in memory and flush them in bulk
+/// to a file, database, or network sink.
+pub trait LogRecordsBuffer {
     /// Appends the provided `record` to the writer intermediate memory.
     ///
     /// # Arguments
@@ -134,5 +131,5 @@ pub trait LogRecordWriter {
     ///
     /// # Returns
     /// A `u16` representing the unique source ID.
-    fn get_id(&self) -> u16;
+    fn get_source_id(&self) -> u16;
 }
--- a/application/apps/indexer/definitions/src/record.rs
+++ b/application/apps/indexer/definitions/src/record.rs
@@ -79,6 +79,7 @@ impl ParseOperationResult {
 ///
 /// If different output is required (e.g., exporting DLT data as raw bytes), a different
 /// parser implementation (such as `DltRaw`) should be used.
+#[derive(Debug)]
 pub enum LogRecordOutput<'a> {
     /// A raw binary message.
     Raw(&'a [u8]),
@@ -94,7 +95,7 @@ pub enum LogRecordOutput<'a> {
 
     /// Structured columnar data. Typically used when the parser can extract
     /// meaningful fields and present them as an array of strings.
-    Columns(&'a [&'a str]),
+    Columns(Vec<Cow<'a, String>>),
 
     /// An attachment object, such as a binary blob or associated metadata.
     /// These are handled separately from textual data and require special treatment
@@ -131,7 +132,7 @@ pub trait LogRecordWriter {
     /// # Returns
     /// * `Ok(())` on success.
     /// * `Err(NativeError)` if the write operation fails.
-    async fn write(&mut self, record: LogRecordOutput<'_>) -> Result<(), stypes::NativeError>;
+    fn write(&mut self, record: LogRecordOutput<'_>) -> Result<(), stypes::NativeError>;
 
     /// Called once when the `MessageProducer` has no more data to provide.
     ///
--- a/application/apps/indexer/definitions/src/record.rs
+++ b/application/apps/indexer/definitions/src/record.rs
@@ -84,17 +84,11 @@ pub enum LogRecordOutput<'a> {
     /// A raw binary message.
     Raw(&'a [u8]),
 
-    /// A borrowed UTF-8 string slice.
-    Str(&'a str),
-
-    /// A string wrapped in a [`Cow`] (Copy-on-Write), allowing either borrowed or owned data.
-    Cow(Cow<'a, str>),
-
-    /// An owned UTF-8 string.
-    String(String),
+    /// Log message line
+    Message(Cow<'a, str>),
 
     /// Structured columnar data. Typically used when the parser can extract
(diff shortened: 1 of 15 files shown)


application/apps/indexer/session/src/handlers/observe/session.rs
```diff
@@ -51,86 +44,61 @@ impl Writer {
...
+
+    async fn flush(&mut self) -> Result<(), stypes::NativeError> {
         if !self.buffer.is_empty() {
-            self.state
-                .write_session_file(self.get_id(), std::mem::take(&mut self.buffer))
-                .await?;
+            // Capacity of cloned item is equal to its length, making this clone
+            // produces one mem_copy command for the needed bytes only on the cloned
+            // string while preserving the capacity of the intermediate buffer.
+            let msgs = self.buffer.clone();
```
DmitryAstafyev, line 97: Thanks. Very nice finding! 
Would it be less expensive, like this?

```
let msgs = std::mem::replace(&mut self.buffer, Vec::with_capacity(self.buffer.capacity()));
```

application/apps/indexer/definitions/src/record.rs
```diff
@@ -79,22 +79,17 @@ impl ParseOperationResult {
...
-    /// A borrowed UTF-8 string slice.
-    Str(&'a str),
-
-    /// A string wrapped in a [`Cow`] (Copy-on-Write), allowing either borrowed or owned data.
-    Cow(Cow<'a, str>),
-
-    /// An owned UTF-8 string.
-    String(String),
+    /// Log message line
+    Message(Cow<'a, str>),
```
DmitryAstafyev, line 88: Ammar, I don’t see any need at this stage to “trim down” the internal API, especially considering that `LogRecordOutput::String` is actively used.

We’ve explicitly designed it so that parsers can choose the most appropriate and efficient representation for their output. For example, the DLT parser uses `LogRecordOutput::String`, while the text parser prefers `LogRecordOutput::Cow`.

I agree that `LogRecordOutput::Str` isn’t currently in use - that’s fine, we can comment it out to signal that it’s part of a potentially extensible API. But replacing all string-related variants with Cow doesn’t seem justified. It simply adds an extra layer of conversion without any tangible benefit, while at the same time sacrificing the clarity and semantic precision of `LogRecordOutput`.

Please let's put back `LogRecordOutput::String` and let DLT and SomeIp parsers use it.

application/apps/indexer/definitions/src/record.rs
```diff
@@ -123,28 +118,14 @@ pub enum LogRecordOutput<'a> {
...
+    /// Appends the provided `record` to the writer intermediate memory.
     ///
     /// # Arguments
     /// * `record` – A reference to a [`LogRecordOutput`] containing the data to write.
-    ///
-    /// # Returns
-    /// * `Ok(())` on success.
-    /// * `Err(NativeError)` if the write operation fails.
-    async fn write(&mut self, record: LogRecordOutput<'_>) -> Result<(), stypes::NativeError>;
+    fn append(&mut self, record: LogRecordOutput<'_>);
```
DmitryAstafyev, line 125: Ammar, I’d prefer to keep the method name `write()` rather than renaming it to `append()`.
The name `append()` implies adding something to an existing dataset, which isn’t the case conceptually.
Also, `write()` aligns with conventional naming in Rust and IO contexts, making it clearer and more consistent.

application/apps/indexer/definitions/src/record.rs
```diff
@@ -123,28 +118,14 @@ pub enum LogRecordOutput<'a> {
...
-    /// to be called only once.
-    ///
-    /// This is the opportunity to flush any buffered data and perform cleanup if necessary.
-    ///
-    /// # Returns
-    /// * `Ok(())` on success.
-    /// * `Err(NativeError)` if the finalization process fails.
-    async fn finalize(&mut self) -> Result<(), stypes::NativeError>;
+    /// Flushes the data from the internal buffer to the underline target
+    async fn flush(&mut self) -> Result<(), stypes::NativeError>;
```
DmitryAstafyev, line 128: Remaining `finalize` to `flush` - is fine.

DmitryAstafyev: Maybe it makes sense also to rename `get_id()` into `get_source_id()` just to make it more semantically clean.

application/apps/indexer/session/src/handlers/observe/session.rs
```diff
@@ -51,86 +44,61 @@ impl Writer {
...
+            self.buffer.clear();
+            self.state.write_session_file(self.get_id(), msgs).await?;
         }
-        for attachment in std::mem::take(&mut self.attachments) {
+        for attachment in self.attachments.drain(..) {
```
DmitryAstafyev, line 101:
```
// TODO: send with 1 call
```
That's important to add, it could boost performance, because it will prevent multiple useless calls to the client with each single Attachment. Maybe make sense to create an issue for it, because it will affect multiple layers of application

application/apps/indexer/definitions/src/record.rs
```diff
@@ -79,22 +79,17 @@ impl ParseOperationResult {
...
-    /// An owned UTF-8 string.
-    String(String),
+    /// Log message line
+    Message(Cow<'a, str>),
 
     /// Structured columnar data. Typically used when the parser can extract
-    /// meaningful fields and present them as an array of strings.
-    Columns(&'a [&'a str]),
+    /// meaningful fields and present them as collection of texts.
+    Columns(Vec<Cow<'a, String>>),
```
DmitryAstafyev, line 92: What is the reason? Plugins?
Btw, we also should remove "columns" logic from DLT & SomeIp parsers. Yes, now in prototype they are suing `String`, but they should return `Columns` instead, and it might be `Columns(&'a [&'a str])` or `Columns(&'a [Cow<'a>])` - we don't know yet. In any way, this is a part of internal API and it my opinion it makes sense to keep it not limited, but quite "open". It costs zero, but gives a good semantic overview of the supported parser's outputs.

application/apps/indexer/definitions/src/record.rs
AmmarAbouZor: Done

application/apps/indexer/definitions/src/record.rs
AmmarAbouZor: Resolved after talking about in separate call

application/apps/indexer/definitions/src/record.rs
AmmarAbouZor: `LogRecordWriter` Changed to `LogRecordsBuffer`

application/apps/indexer/session/src/handlers/observe/session.rs
AmmarAbouZor: I'll create an issue for it to not forget about it

DmitryAstafyev (commented):
### Architectural Comment

The architecture I introduced in the components-model branch simplifies and clarifies the parser API by removing the iterator from `parse(...)` and replacing it with a simple, direct interface (like it was originaly):

```
fn parse(&mut self, input: &[u8], ...) -> Result<(usize, Option<Msg>), ParserError>
```
To support plugin-specific batch processing (required by the plugins-sys), the only additional step needed is to **add an internal buffer** in `PluginParser`:

```
pub struct PluginParser {
    store: Store<ParserPluginState>,
    buffer: Vec<bindings::ParseReturn>, // ← internal buffer for batch output
    plugin_bindings: Parse,
}
```
Including an internal buffer in `PluginParser` is not only appropriate - it’s the correct architectural choice.

Due to the nature of WASM plugins, we cannot receive any data by reference. All data must be passed as owned, static values - e.g., `String` instead of `&str`, `Vec<u8>` instead of `&[u8]`, and so on.

Because of this, there is no technical or semantic obstacle to buffering the plugin’s output inside `PluginParser`.

I fully acknowledge that if `MessageProducer` mistakenly calls `parse()` on `PluginParser` with a new input slice while the buffer is not yet exhausted, we would - de facto - ignore the new input and continue returning buffered items.

However, this is exactly why `impl Iterator for Parser` was introduced - to define consistent behavior at the `MessageProducer` level and control the interaction pattern.

More importantly, this remains internal logic, not something that pollutes the public API.

If we’re still concerned about accidentally calling `parse()` outside of an iteration context, we can simply extend the parser signature with `offset` and `len`, allowing the parser to verify that it's processing a fresh slice - eliminating any ambiguity.

Okay, as I hightlighted, for iterator-style access to parser results within a single byte slice, we provide an implementation:

```
impl Iterator for Parser { ... }
```

This iterator is used internally by `MessageProducer`, allowing stream-style consumption without complicating the core API and preventing problems described above.

**This approach ensures**:

- A clean and minimal parser API
- No leakage of plugin-specific or implementation details into public traits
- Strict separation of concerns between `Parser` and `MessageProducer`

This PR reverts that architectural simplification and reintroduces the `parse(...) -> impl Iterator` model.

As a result:

- In order to support the plugins-sys, the output is now wrapped in a `Box<dyn Iterator>`, causing a heap allocation on every `parse()` call
- Each parser implementation is expected to construct its own iterator, **increasing the likelihood of inconsistent behavior and potential errors**
- The responsibility for stream handling now **leaks into the parser**, blurring the boundary between `Parser` and `MessageProducer`
- Introduces the addition API `SingleParser`, which actually just confuses 

### Summary

Returning an iterator from parse() brings no clear advantage, but introduces:
- API pollution,
- unnecessary allocation overhead,
- and a loss of clear modularity between layers.

My position remains **unchanged**: returning an iterator from `parse()` works stably and predictably, but at the same time, it is an obvious design mistake.

However, due to current project circumstances and time constraints, I will merge this change into, with this comment recorded for future reference.

PR #2361, opened 2025-07-03
Components: Plugins Integration

AmmarAbouZor:
This PR provides basic integration for plugins into components refactoring

It provides:
* It moves plugins manager from unbound session into component session. The reason of this change is that we need access to the plugins manager for registering its plugins as components and for calling the various methods (validate, config, lazy_config...)
* Integrate plugins into components using plugins manager, adding the needed functionality to add parser plugins with placeholder for new functionality and types. 
* The final integration should provide changes on the API level of the plugins which will be done once the changes in this refactoring are finalized
* This PR doesn't included the needed changes in front-end to wire up plugins completely. (Plugins are shown up in selection but no configs are shown in the UI, and plugin sessions fail on start)


DmitryAstafyev (approved):
I think it can go just like it is... I will try to make it work in the scope of my PR

PR #2363, opened 2025-07-08
Components: Use Full Dynamic Dispatch + Refactoring

AmmarAbouZor:
This PR includes:

* Use dynamic dispatch fully inside session and components removing not needed generics and enums.
* Implement Parser/ByteSource for Box<dyn ...> manually because Rust doesn't support the blanket implementation: `impl<T> T for Box<dyn T>`
* Use async_trait to make source trait object safe without generics since async_trait produced `Pin<Box<dyn Future ...>>` instead of the generic `impl Futures<...>`
* Add Debug trait bound on descriptor instead of temporary implementation.


AmmarAbouZor:
@DmitryAstafyev Thanks for your feedback.
In my opinion we can pick either one of:
* **Static Dispatch:** We have enums and a generic struct `Producer` so we match against the enum before constructing the producer. If we care about performance that much then we should go with this approach.
* **Dynamic Dispatch:** We replace the enums with `Box<dyn ...>` making the virtual-table does its job with the dynamic dispatch approach.

Using and enum as a trait object isn't a recommended pattern in Rust, and for keeping it we must have a good reason, which I don't see.

I don't like the dynamic dispatched solution as the static dispatch in the previous PR #2357 . But I went with it because you rejected the previous one.

DmitryAstafyev:
Thanks for your reply, @AmmarAbouZor .

Just to clarify, **the rejection of PR #2357 had nothing to do with the use of static dispatch.** The issue was architectural: that PR broke the core design principle of the `components_model` branch - namely, keeping parsers and sources **isolated** from the core.

It exposed internal parser implementation details and enums at the top-level API (level of `session`), defeating the goal of creating a clean and minimal interface layer. This has been clearly stated and documented in the architectural comment I left at the time.

So please - let’s not misrepresent history or rewrite motivations.

Your current PR introduces a completely different kind of shift: replacing enum-based type clarity with `Box<dyn T>` based dynamic dispatch, without any real-world need for that shift.

Furthermore, we are not working on the master branch, but on the components_model branch - a potential global update of the API. At this stage, the priority is to make it fully functional: integrate plugins, restore functionality, and bring tests back to life. This is not the time to rewrite the architecture. Once we have a final working version, I’ll be happy to discuss what could be improved or revised.

AmmarAbouZor:
> Just to clarify, the rejection of PR https://github.com/esrlabs/chipmunk/pull/2357 had nothing to do with the use of static dispatch. The issue was architectural: that PR broke the core design principle of the components_model branch - namely, keeping parsers and sources isolated from the core.

The only way (without macros) to use static dispatch in this code architecture is to match against the enums in `sessions`, so by rejecting that, we are rejecting the static dispatch advantages in favor of dynamic dispatch for the architecture vision.

> Furthermore, we are not working on the master branch, but on the components_model branch - a potential global update of the API. At this stage, the priority is to make it fully functional: integrate plugins, restore functionality, and bring tests back to life. This is not the time to rewrite the architecture. Once we have a final working version, I’ll be happy to discuss what could be improved or revised.

The `components_model` branch has already changed the whole app architecture, and it would make sense to have an agreed-opun architecture before moving further to avoid having another refactoring right after finishing this one.

I'll stop here until we get together and see if we can have a unified vision about this refactoring. Otherwise, I would be happy to support the great work and vision in this branch by wiring up the plugins system to it when needed :rose:

DmitryAstafyev (changes requested):
Please, take a look to comment above

AmmarAbouZor:
Changes here are outdated with the recent big refactoring on the components main branch.
We'll keep this PR for reference until re-implementing it with the current branch

AmmarAbouZor:
This will be closed as we don't intend to continue working on the components anymore

PR #2400, opened 2025-09-01
Stop File tailing on truncate & Small improvement.

### LINKED ISSUE #2394: Tail stops working after file got truncated
**Prerequisites**
create a text file with this content

```
line 1
line 2
````

affected version:
3.18.2

make sure to check this before filing an issue:

- [x] I am running the latest version
- [x] I checked to make sure that this issue has not already been filed

**Describe the issue**
After a file got truncated, the tail stops working and an empty results pane is shown
This should not happen. Instead the app shall behave like a normal tail

**Steps to Reproduce**

- open the file being created in prerequisites
- open a terminal and append something to this file: echo "line 3" >> file.log (this will work of course)
- open a terminal and append something to this file: echo "new line 1" > file.log (this will cause the issue since the file gets truncated)

The issue can be fixed by monitoring the file size. If the file size decreases, the tail could start over at the beginning of the file

**Environment**

Windows, Version 3.18.2

AmmarAbouZor:
This PR closes #2394 

It provides the following:
* Stop file tailing if file get truncated returning and IO error because session should be terminated at this point.
* Avoid allocating string for logging channels errors since the channel will return the sent item on errors.

### DIFF
--- a/application/apps/indexer/session/src/operations.rs
+++ b/application/apps/indexer/session/src/operations.rs
@@ -202,8 +202,8 @@ impl OperationAPI {
     }
 
     pub fn emit(&self, event: stypes::CallbackEvent) {
-        let event_log = format!("{event:?}");
         if let Err(err) = self.tx_callback_events.send(event) {
+            let event_log = format!("{:?}", err.0);
             error!("Fail to send event {event_log}; error: {err}")
         }
     }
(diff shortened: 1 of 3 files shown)


marcmo (approved):
LGTM

PR #2416, opened 2025-09-22
Refactor Producer: Collect records and write them in bulks & Fix tailing for binary files

### LINKED ISSUE #2412: Producer: Send parsed message in bulks + Inject cache
We've learned from the work in the `components` branch that we can improve the performance by writing the parsed messages in bulks instead of writing them one by one. 
Our current implementation for the parser is parsing all available messages in the provided bytes, which makes this change simple and with no breaking changes.
 
Additionally, we can improve the performance by injecting the writer buffer into the producer instead of the internal cache in the producer to avoid caching the messages twice

AmmarAbouZor:
This PR closes #2412

It takes the inspiration from some of the work that is done in the components branch, to write the log messages into the internal buffer and send them to session in bulks instead of sending them one by one.
Those changes led to significant performance gain in parsing binary files (Up to 2.35X faster). 

It also includes a refactoring to the producer loop, making it deliver useful infos about its current producing state instead of swallowing the errors and return an option only.

While refactoring, I've fixed tailing functionality for binary files which haven't worked before.

### This PR includes:

* Moving producer, its tests and benchmarks into processor crate instead of sources.
* Introduce the trait `LogRecordsCollector` to be provided to the producer to append all the parsed logs into it. Then provide couple of implementations for each use-case:
  - General implementation which collect the items into internal buffer and gives access to that buffer after that.
  - Special implementation for observing sessions with multiple internal buffers to improve the performance.
* Producer main method returns now a result with state of the current produce call or and error in case of errors.
* Fix tailing for binary files.
* Adjustments in implementation for Chipmunk-CLI tool as well, since it uses the producer directly.
* Rewrite the unit tests for the producer and the cancel safety tests.
* Adjustments for benchmarks to match the new interface of the producer + Changes in Development CLI configurations after moving them into processor crate.

### Performance Overview:

Performance can be seen on parsing big binary files, especially on files which have short messages (since we are sending a lot of them in bulks).
The performance gain in my observation is between (60.50% - 36.17%) or (2.53x - 1.57x) faster.

| File Description   | Master (Avg. Time) | Collector (Avg. Time) | Improvement | Times Faster |
|:-------------------|:-------------------|:----------------------|:------------|:-------------|
| DLT File (540 MB)  | 12066.33 ms        | 4766.67 ms            | 60.50%      | 2.53x        |
| DLT File (420 MB)  | 10862.33 ms        | 6933.67 ms            | 36.17%      | 1.57x        |


marcmo (approved):
very nice! super good performance improvements...and nice code!

PR #2422, opened 2025-10-02
Add import and export documentation

### LINKED ISSUE #2417: Include Export/Import of logs in Documentation
The current documentation doesn't include the functionality of exporting and importing logs in a clear way.

It would be helpful for the users to find such functionality quick in the docs since it can be one of the most needed features with basic cooperation use cases

sudeeptarlekar:
Currently Chipmunk docs are missing information around importing and exporting functionality. This commit adds the page dedicated around import/export functionality in Chipmunk

Fixes #2417


AmmarAbouZor (changes requested):
Thanks for improving our documentations with assets :pray: 
I've got multiple points to fix:
* Your current assets use a DLT file which isn't open source. Those assets need to be removed because we don't even own the content of the used DLT file. Instead I would suggest using the logs of chipmunk app itself (Can be found in `$HOME/.chipmunk/..`) or the output of some long command (On linux I use `journalctl` command).
* We already have a part for [exporting and importing filters in our current docs](https://esrlabs.github.io/chipmunk/search/#importing-exporting). You can check that and improve the docs there if needed, but this part should be intended to export and import the logs only. (The title of the new file should include that to avoid any confusion)
* The new file should be included in `mkdocs.yml` file so it's part of the documentation. Also, you need to test the new docs on your machine using the command `mkdocs serve --open`

AmmarAbouZor:
> Do you think it does makes sense to document export the log, as user can only export and import is not possible.

Correct, sorry used to write export and import together. I would suggest naming the whole page `Export Logs`

> So Imo adding import search/filters documentation is main thing...

I think the current place for importing search and filters is suitable because all filters its related stuff are gathered in on document, making the users able to learn about the exporting and importing filters directly after learning about them.

AmmarAbouZor (changes requested):
The recordings are great thank :100: 

I think you misunderstood what I've meant in the last comments. Here are more details:
* We need a stand-alone document with the name `Export Logs`. This should contain info about exporting logs + the assets
* We don't need to change anything in `search.md` document. `Export Logs` isn't part of it (I was talking about `Export/Import Presets` in the last comment and though that you wanted to move them to the `Exporting` document)
* We don't need the `Exporting` stand-alone document

AmmarAbouZor (changes requested):
I've started with the review but I've found out that we still have multiple places where filters are mentioned in the export logs document and where importing filters is removed form filters document, so I don't think this PR is ready for review in this state.

Also, We don't need to break the documentation lines when they reach a certain length, since the web view will wrap them automatically as they fit. Having the breaks in the raw text is a little bit confusing.

AmmarAbouZor (changes requested):
I've added some comments

Also, we don't need to break the lines manually once they've reached a certain length in markdown files.

AmmarAbouZor (changes requested):
I've found a small change that we don't need. Maybe it got forgotten

PR #2428, opened 2025-10-12
Core Refactoring & Associated type for parsers instead of Generics

AmmarAbouZor:
This PR introduces the needed refactoring for core module before starting with native UI development.

It also includes replacing using direct generics to define the output type for each parser with associated type to avoid carrying this generic further in type and function definitions which are generic over parsers.

Also, it includes introducing named types in all places, where too complex tuples were previously used and where `Clippy` Lints were suppressed instead of resolving the issues.

It provided basic documentations for undocumented module and main function in rust core.

Affected modules: `processor`, `session` and `parser`


marcmo (approved):
LGTM

Issue #2430, opened 2025-10-13
Bug: Presets' names start with counter and clicking on it increases the counter


AmmarAbouZor:
**Prerequisites**

affected version: 3.19.1

make sure to check this before filing an issue:

- [x] I am running the latest version
- [x] I checked to make sure that this issue has not already been filed

**Describe the issue**

When we add a filter, then a counter is added to its name automatically `(1)`, then each time we click on the preset this counter will increase.
We need to consider the purpose of this counter and remove if it doesn't provide any value

**Environment**

Ubuntu & Windows

PR #2431, opened 2025-10-14
Fix used counter for Filter and presets

sudeeptarlekar:
When user imports the presets. Chipmunk shows number 1 next to the filename which suggests that filter is already been used. This commit fixes that bug.

Fixes: #2430

### DIFF
--- a/application/client/src/app/service/history/collections.ts
+++ b/application/client/src/app/service/history/collections.ts
@@ -62,7 +62,7 @@ export class Collections implements EntryConvertable, Equal<Collections>, Empty 
             // For older filters saved on the user machine we have the used value
             // set as a 1 by default. To fix the counter we need to reset it to 0
             // because user didn't use it yet.
-            if (def.used >= 1) {
+            if (def.used > 0) {
                 def.used = 0;
             }
             return new Collections(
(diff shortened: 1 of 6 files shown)


AmmarAbouZor (changes requested):
I've tested that on my machine. and I think the current behavior isn't entirely correct, because now we are getting that the number of used filters is set to zero even though the filter is currently in use (See screenshot). Then after I click on the preset the counter will get into 1 and then stop increasing.

This is easily reproduceable by just applying a filter or opening a session with saved filter in it 

image

application/client/src/app/service/history/collections.ts
```diff
@@ -59,9 +59,15 @@ export class Collections implements EntryConvertable, Equal<Collections>, Empty
             );
         } else {
             const def = Collections.fromMinifiedStr(JSON.parse(smth.content));
+            // For older filters saved on the user machine we have the used value
+            // set as a 1 by default. To fix the counter we need to reset it to 0
+            // because user didn't use it yet.
+            if (def.used > 0) {
+                def.used = 0;
+            }
```
AmmarAbouZor: I think we can just set the value to zero without having an if branch.

AmmarAbouZor (changes requested):
Thanks for the fixes. I've added some comments regarding the code changes and here are some points form testing the changes:
* We need to think about highlighting the used one because the current color doesn't affect the readability of the texts. Can you consider leaving the background color the same but make the title of used filters bold instead?
* I ran into situation when I imported filters and set list filter to `All` where some presets have `(0)` and others don't have it
image

sudeeptarlekar:
Closing in as duplicate of #2432

PR #2432, opened 2025-10-27
Bug/2430/used counter

sudeeptarlekar:
Fixes: #2430

### DIFF
--- a/application/platform/env/obj.ts
+++ b/application/platform/env/obj.ts
@@ -418,12 +418,17 @@ export function objToStringMap(obj: {
 }
 
 export function getAsSetOfStringsOrEmpty(src: any, key: string): Set<string> {
-    if (src[key] === undefined) {
-        return new Set();
-    } else if (!(src[key] instanceof Set)) {
-        throw new Error(`Parameter "${key}" should be valid Set`);
+    const target = src[key];
+
+    if (!(target instanceof Set)) {
+         return new Set();
     }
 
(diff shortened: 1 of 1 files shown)


application/platform/env/obj.ts
```diff
@@ -416,3 +416,17 @@ export function objToStringMap(obj: {
...
+export function getAsSetOfStringsOrEmpty(src: any, key: string): Set<string> {
+    if (src[key] === undefined) {
+        return new Set();
+    } else if (!(src[key] instanceof Set)) {
+        throw new Error(`Parameter "${key}" should be valid Set`);
+    }
+
+    Array.from(src[key]).every(
+        id => typeof id === 'string' || (() => { throw new Error(`Parameter "${key}" should be Set of strings`); })()
+    );
```
AmmarAbouZor: We tend to avoid javascript code tricks in favor of readability in Chipmunk.

How about having the function like this? This is the exact code taken from the function `getAsStringsArrayOrUndefined()` at the same page

```typescript
export function getAsSetOfStringsOrEmpty(src: any, key: string): Set<string> {
    const target = src[key];

    if (!(target instanceof Set)) {
         return new Set();
    }

    target.forEach((el: any) => {
        if (typeof el !== 'string') {
            throw new Error(`Parameter "${key}" must be a Set of strings. Found a non-string value.`);
        }
    });

    return target as Set<string>;
}
```

PR #2437, opened 2025-11-13
Spawn Shells for commands & Replace `envvars` with straight-forward shell handling 

### LINKED ISSUE #2434: Environment Variables and Shells implementation behaves like Malware
The current implementation for running shell commands from within Chipmunk, runs the commands as system-processes without running the shell at first. This led to problem with loading environment variables which is solved currently with the crate [envvars](https://github.com/esrlabs/envvars) which execute all available shells at runtime to extract their environment variables so it can apply it to the spawned process (Which on its own a none expected behavior from a log analyzer to execute all available shells when it starts).

However, the implementation of `envvars`, creates an external executable binary with a name generated from UUID inside system temp directory then execute it separately, which is a really suspicious behavior that can get caught be Antiviruses marking Chipmunk app as Malware.

We can solve the original problem without having to execute the shells by simply run the commands by spawning a shell process and giving it the command as argument. 
With that the shell will load its env variables and it will parse the command itself instead of the current primitive command parsing that is implemented in Chipmunk which fails on many edge-cases (like setting env variable before running the command and command piping, etc)

AmmarAbouZor:
This PR closes #2434 

It changes the way Chipmunk handle process command entirely with the following points:
* Chipmunk will spawn a shell now and pass the user commands to it instead of the previous implementation which spawn the command as a process with a built-in basic shell parsing which fails on many cases (For example setting env variables at the start of command `MY_VAR=3 echo $MY_VAR` and reading environment variables + piping commands and many more).
* Remove `envvars` crates with its malware like behavior (Placing executable binary with UUID name in temp directory and running it as external process). This crate was used to execute all the shells installed on the system by the start of chipmunk so we can have their environment variables to pass to the process. This problem is simply solved by spawning the shells so it can load its variables on its own.
* The new implementation will check if a group of well known shells are installed on the system using `which` crate and present them to users. If one of them is selected then it will be spawned instead of the default shell (`sh` on unix and `cmd` on windows). We need to know the shell type up front because each shell has its own flag for passing commands (`-c` on the sane shells, `/C` on cmd and `-Command` on PowerShell).
* Previous behavior was saving the environment variables in the session recent file as plain text without any encryption. This behavior is removed entirely since it's a huge security threat.
* This PR also include parts of source code that isn't used.

### DIFF
--- a/application/apps/protocol/src/lib.rs
+++ b/application/apps/protocol/src/lib.rs
@@ -126,7 +126,6 @@ gen_encode_decode_fns!(FileFormat);
 gen_encode_decode_fns!(ObserveOrigin);
 gen_encode_decode_fns!(FoldersScanningResult);
 gen_encode_decode_fns!(DltStatisticInfo);
-//TODO AAZ: Adjust TS-Bindings tests with the type changes then run proptests
 gen_encode_decode_fns!(ShellType);
 gen_encode_decode_fns!(ShellProfile);
 gen_encode_decode_fns!(ProfileList);
--- a/application/apps/rustcore/ts-bindings/spec/_session.benchmark.spec.ts
+++ b/application/apps/rustcore/ts-bindings/spec/_session.benchmark.spec.ts
@@ -105,7 +105,7 @@ describe('Benchmark Tests', function () {
                                     .process({
                                         command: `less ${tmpobj1.name}`,
                                         cwd: process.cwd(),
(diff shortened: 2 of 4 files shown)


AmmarAbouZor:
Solution tested on Linux Ubuntu 24.04 LTS and Windows

AmmarAbouZor:
I'll merge the changes and create a release since it's urgent by Linux users with antivirus to have the changes as quick as possible

PR #2478, opened 2026-02-05
Resolve chipmunk execution issues in Ubuntu 24

itsmesamster:


AmmarAbouZor (changes requested):
This is working great. 
I'm just wondering for the case if end-users installed chipmunk from the archives and not via `.deb` installer. Will Chipmunk works for them form the command line? 
I think we still need to provide a script that they can run once with `sudo` to solve the problem for them. Then we can extend our README with a command to download the script  (via curt) and run it

PR #2483, opened 2026-02-23
Fix construction of non-verbose args for master

kruss:
This depends on: https://github.com/esrlabs/dlt-core/pull/52


AmmarAbouZor:
I think we can mark this PR as draft until the other one in `dlt-core` is merged

PR #2484, opened 2026-02-23
Fix construction of non-verbose args

kruss:
This depends on: https://github.com/esrlabs/dlt-core/pull/52


kruss:
PR will be done on master only for later final merge with native-ui

PR #2491, opened 2026-03-04
Add native home screen

kruss:
Add a native-ui home screen with support for quick-actions, recent configurations and favorite folders.


application/apps/indexer/gui/application/src/host/ui/home/settings.rs
```diff
@@ -0,0 +1,309 @@
...
+        Ok(())
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct FavoriteFolder {
+    pub path: PathBuf,
+
+    #[serde(skip)]
+    pub files: Vec<(String, String)>,
```
AmmarAbouZor: Can we have a struct here with named fields instead of the tuple `(String, String)`

kruss:
will review in another PR

PR #2496, opened 2026-03-17
TOOL-700: Add native attachments

the-nick-fischer:
[TOOL-700:](https://esrlabs.atlassian.net/browse/TOOL-700) This PR adds a basic implementation of the attachments panel to the native UI:
- General layou with header, list and preview sections
- Functional attachments list
- Selection and saving of attachments
- Display of attachments symbol in logs table

What is missing:
- Attachments preview
- Quick filters for attachments
- Go to / select related row(s)

### DIFF
--- a/application/apps/indexer/gui/application/src/host/service/mod.rs
+++ b/application/apps/indexer/gui/application/src/host/service/mod.rs
@@ -1,6 +1,5 @@
 use std::{
     collections::HashMap,
-    fs,
     ops::Not,
     path::{Path, PathBuf},
     thread,
@@ -8,7 +7,7 @@ use std::{
 
 use anyhow::Result;
 use itertools::Itertools;
-use log::{error, trace};
+use log::trace;
 use tokio::runtime::Handle;
 use uuid::Uuid;
 
@@ -201,57 +200,10 @@ impl HostService {
                 // Cleanup on shutdown goes here.
                 let _ = confirm_tx.send(());
             }
-            HostCommand::CopyFiles { copy_file_infos } => {
-                for copy_file_info in copy_file_infos {
-                    self.copy_file(copy_file_info.source, copy_file_info.destination)
-                        .await?;
-                }
-            }
+            HostCommand::CopyFiles { copy_file_infos } => file::copy_files(copy_file_infos).await?,
         }
 
         Ok(())
-    }
-
-    async fn copy_file(&self, source: PathBuf, destination: PathBuf) -> Result<(), HostError> {
-        let source_clone = source.clone();
-        let destination_clone = destination.clone();
-
-        let copy_result =
-            tokio::task::spawn_blocking(move || fs::copy(&source, &destination)).await;
-
-        match copy_result {
-            Ok(Ok(_)) => Ok(()),
-            Ok(Err(error)) => {
-                let error_message = format!(
-                    "Failed to copy file '{}' from '{}' to '{}': {error}",
-                    source_clone.file_name().unwrap_or_default().display(),
-                    source_clone.display(),
-                    destination_clone.display(),
-                );
-
-                error!("{error_message}");
-                Err(HostError::NativeError(NativeError {
-                    severity: Severity::ERROR,
-                    kind: NativeErrorKind::Io,
-                    message: Some(error_message),
-                }))
-            }
-            Err(join_error) => {
-                let error_message = format!(
-                    "Failed to copy file {} from '{}' to '{}': task failed: {join_error}",
-                    source_clone.file_name().unwrap_or_default().display(),
-                    source_clone.display(),
-                    destination_clone.display(),
-                );
-
-                error!("{error_message}");
-                Err(HostError::NativeError(NativeError {
-                    severity: Severity::ERROR,
-                    kind: NativeErrorKind::Io,
-                    message: Some(error_message),
-                }))
-            }
-        }
     }
 
     async fn open_single_file(&self, file_path: PathBuf) -> Result<(), HostError> {
--- a/application/apps/indexer/gui/application/src/host/ui/actions/mod.rs
+++ b/application/apps/indexer/gui/application/src/host/ui/actions/mod.rs
@@ -8,6 +8,7 @@ mod host_action;
 
 pub use file_dialog::{FileDialogFilter, FileDialogHandle, FileDialogOptions};
 pub use host_action::HostAction;
+
 /// A handle to be passed between UI components to get access to
 /// shared UI functions like notifications.
 #[derive(Debug)]
--- a/application/apps/indexer/gui/application/src/session/ui/logs_table/mod.rs
+++ b/application/apps/indexer/gui/application/src/session/ui/logs_table/mod.rs
@@ -25,6 +25,7 @@ use crate::{
     },
 };
 
+/// This is used for storing some attachment state during rendering of the logs table only.
 pub enum LogAttachmentInfo {
     NoAttachment,
     WithAttachment { color: Option<egui::Color32> },
--- a/application/apps/indexer/gui/application/src/session/ui/mod.rs
+++ b/application/apps/indexer/gui/application/src/session/ui/mod.rs
@@ -1,7 +1,7 @@
 use std::rc::Rc;
 
 use egui::{CentralPanel, Frame, Margin, Panel, Ui};
-use log::error;
+use log::warn;
 use tokio::sync::mpsc::Sender;
 
 use crate::{
@@ -251,7 +251,7 @@ impl Session {
                 SessionMessage::AttachmentsUpdated { attachment, len } => {
                     self.shared.attachments.add(attachment);
                     if self.shared.attachments.attachments().len() as u64 != len {
-                        error!(
+                        warn!(
                             "Unexpected internal error: Attachment count mismatch: expected {} from backend, got {}.",
                             len,
                             self.shared.attachments.attachments().len()
--- a/application/apps/indexer/gui/application/src/session/ui/shared/attachments.rs
+++ b/application/apps/indexer/gui/application/src/session/ui/shared/attachments.rs
@@ -1,4 +1,3 @@
-use log::error;
 use rustc_hash::FxHashMap;
 use stypes::AttachmentInfo;
 use uuid::Uuid;
@@ -7,8 +6,6 @@ use crate::host::common::colors;
 
 #[derive(Debug, Default)]
 pub struct AttachmentsState {
-    /// TODO: Evaluate other options of how we can store the attachment state without Vec and
-    /// multiple indices for lookups.
     /// NOTE: Ordered, index-accessible storage is required for egui::ScrollArea::show_rows()
     attachments: Vec<AttachmentInfo>,
     /// Index for lookups: attachment UUID -> attachments list index
@@ -22,32 +19,29 @@ pub struct AttachmentsState {
 impl AttachmentsState {
     pub fn add(&mut self, attachment: AttachmentInfo) {
         let uuid = attachment.uuid;
+        let index = self
+            .index_by_uuid
+            .get(&attachment.uuid)
+            .copied()
+            .unwrap_or(self.attachments.len());
 
-        // In case an attachment is already registered, avoid duplication.
+        for &position in &attachment.messages {
+            self.index_by_position.insert(position, index);
+        }
+
+        // Assumption: index_by_uuid and color_by_extension cannot change when a attachment is updated.
         if self.index_by_uuid.contains_key(&uuid) {
-            error!(
-                "Attachment with uuid {uuid} has already been registered; skipping attachment: {} {:?}",
-                attachment.name, attachment.filepath,
-            );
             return;
         }
 
-        let attachments_index = self.attachments.len();
-
-        for &position in &attachment.messages {
-            self.index_by_position.insert(position, attachments_index);
-        }
-
-        self.index_by_uuid.insert(uuid, attachments_index);
+        self.index_by_uuid.insert(uuid, index);
 
         if let Some(ext) = attachment.ext.as_deref()
             && !self.color_by_extension.contains_key(ext)
         {
-            let extensions_index = self.color_by_extension.len();
-
             self.color_by_extension.insert(
                 ext.to_string(),
-                colors::search_value_color(extensions_index),
+                colors::search_value_color(self.color_by_extension.len()),
             );
         }
 
--- a/application/apps/indexer/gui/application/src/host/service/file.rs
+++ b/application/apps/indexer/gui/application/src/host/service/file.rs
@@ -4,11 +4,13 @@
 //! scan directories for specific file types, and format file sizes for user display.
 
 use std::{
-    io,
+    fs, io,
     path::{Path, PathBuf},
 };
 
-use stypes::FileFormat;
+use stypes::{FileFormat, NativeError, NativeErrorKind, Severity};
+
+use crate::host::{command::CopyFileInfo, error::HostError};
 
 /// Detects the [`FileFormat`] of a file at the given path.
 pub fn get_file_format(file_path: &Path) -> io::Result<FileFormat> {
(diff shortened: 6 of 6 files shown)


application/apps/indexer/gui/application/src/session/ui/bottom_panel/search/search_table.rs
```diff
@@ -181,22 +181,41 @@ impl<'a> LogsDelegate<'a> {
...
             })
             .unwrap_or_default();
-        let is_bookmarked = self
-            .get_log_item(cell)
-            .is_some_and(|item| self.shared.logs.is_bookmarked(item.element.pos as u64));
 
-        let bookmark_pos = self.get_log_item(cell).map(|item| item.element.pos as u64);
-        let response =
-            common::logs_tables::render_row_header(ui, text, color_idx, is_bookmarked, || {
+        let bookmark_pos: Option<usize> = self.get_log_item(cell).map(|item| item.element.pos);
```
AmmarAbouZor: I think we should rename it `element_pos`

application/apps/indexer/gui/application/src/session/ui/shared/attachments.rs
```diff
@@ -0,0 +1,78 @@
...
+    // TODO: Evaluate other options of how we can store the attachment state without Vec and
+    // multiple indices for lookups.
+    // NOTE: Ordered, index-accessible storage is required for egui::ScrollArea::show_rows()
+    attachments: Vec<AttachmentInfo>,
+    // Index for lookups: attachment UUID -> attachments list index
+    index_by_uuid: FxHashMap<Uuid, usize>,
+    // Index for lookups: log position -> attachment UUID
+    uuid_by_position: FxHashMap<usize, Uuid>,
+    // Index for lookups: attachment extension -> associated color
+    color_by_extension: FxHashMap<String, egui::Color32>,
```
AmmarAbouZor: I would use triple slashes `///` for normal documentation so they are visible for LSP hover actions

application/apps/indexer/gui/application/src/session/ui/shared/attachments.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+impl AttachmentsState {
+    pub fn add(&mut self, attachment: AttachmentInfo) {
```
AmmarAbouZor, line 20: Is it possible to got multiple updates for the same attachment?
In that case we must check if the attachment already exits and avoid duplicating it

application/apps/indexer/gui/application/src/session/ui/shared/attachments.rs
```diff
@@ -0,0 +1,78 @@
...
+#[derive(Debug, Default)]
+pub struct AttachmentsState {
+    // TODO: Evaluate other options of how we can store the attachment state without Vec and
+    // multiple indices for lookups.
+    // NOTE: Ordered, index-accessible storage is required for egui::ScrollArea::show_rows()
+    attachments: Vec<AttachmentInfo>,
+    // Index for lookups: attachment UUID -> attachments list index
+    index_by_uuid: FxHashMap<Uuid, usize>,
+    // Index for lookups: log position -> attachment UUID
+    uuid_by_position: FxHashMap<usize, Uuid>,
```
AmmarAbouZor: Is it possible to make this map `index_by_position // log position -> attachment index` ?
This will save us a lookup in `attachment_by_log_position()`

application/apps/indexer/gui/application/src/session/ui/common/logs_tables.rs
```diff
@@ -94,6 +94,8 @@ pub fn render_row_header(
     color_idx: Option<usize>,
     is_bookmarked: bool,
     on_toggle_bookmark: impl FnOnce(),
+    has_attachment: bool,
+    attachment_color: Option<egui::Color32>,
```
AmmarAbouZor: I would pass the attachment index or id here and figure out the color inside the function. This is needed later for showing the overlay.
Also it will avoid misusing the API with something like `has_attachment: false, attachment_color: Some(Red)`

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
```diff
@@ -1,20 +1,461 @@
...
+            .rect_filled(row_rectangle, 0, row_background_color);
+
+        if is_selected {
+            ui.painter().rect_filled(
+                egui::Rect::from_min_size(
+                    row_rectangle.min,
+                    egui::vec2(4.0, row_rectangle.height()),
+                ),
+                egui::CornerRadius::same(0),
+                SELECTION_BAR_BLUE,
```
AmmarAbouZor: I would use egui app global styles so we can change it from one place `ui.visuals().selection.stroke`

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
```diff
@@ -1,20 +1,461 @@
...
+        });
+    }
+
+    fn render_attachments_header(&self, ui: &mut egui::Ui, attachments_count: usize) {
+        egui::Sides::new().show(
+            ui,
+            |ui| ui.label(format!("Attachments ({})", attachments_count)),
+            |ui| {
+                ui.menu_button(
+                    egui::RichText::new(egui_phosphor::fill::FUNNEL).size(16.0),
```
AmmarAbouZor: If you want the `fill` variant for an icon then you need to set the font_family for the RichText. Otherwise you can use `regular` instead

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
```diff
@@ -1,20 +1,461 @@
...
+
+    fn render_attachments_list(
+        &mut self,
+        attachments_state: &AttachmentsState,
+        ui_actions: &mut UiActions,
+        ui: &mut egui::Ui,
+    ) {
+        let attachments = attachments_state.attachments();
+        ui.spacing_mut().item_spacing.y = 0.0;
+        // TOD: Reevaluate if this can be done better or uniformely across UI
```
AmmarAbouZor: I think this can be either resolve if we are talking about UI in this widget. For other views across the app I would leave it for each view to decide the size as each one of them contains different content

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
```diff
@@ -1,20 +1,461 @@
...
+                    |ui, row_range| {
+                        for current_row in row_range {
+                            let is_selected = self.selected_rows.contains(&current_row);
+                            let attachment = &attachments[current_row];
+
+                            let extension_color = attachment
+                                .ext
+                                .as_deref()
+                                .and_then(|ext| attachments_state.color_by_extension(ext))
+                                .unwrap_or(egui::Color32::GRAY);
```
AmmarAbouZor: I would put the default extensions color in the shared colors module as it can be reached from this view and the tables

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
```diff
@@ -1,20 +1,461 @@
...
+                                is_selected,
+                            );
+
+                            if response.clicked() {
+                                self.clicked_row = Some(current_row);
+                                self.handle_row_click(current_row, ui.input(|i| i.modifiers));
+                            }
+
+                            if response.secondary_clicked() {
+                                self.clicked_row = Some(current_row);
```
AmmarAbouZor, line 191: Is it intentionally that secondary click doesn't call `handle_row_click()`

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
```diff
@@ -1,20 +1,461 @@
...
 
-#[allow(unused)]
 #[derive(Debug)]
 pub struct AttachmentsUi {
+    #[allow(unused)]
     cmd_tx: mpsc::Sender<SessionCommand>,
+    selected_rows: FxHashSet<usize>,
+    clicked_row: Option<usize>,
+    previously_clicked_row: Option<usize>,
+    pending_attachment_save: FxHashSet<Uuid>,
```
AmmarAbouZor, line 44: I would like to see short docs for some fields explaining their purpose without having to figure that out from the code itself

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
```diff
@@ -1,20 +1,461 @@
...
+        }
+    }
+
+    fn save_attachment(
+        &self,
+        ui_actions: &mut UiActions,
+        attachment: &AttachmentInfo,
+        destination: &Path,
+    ) {
+        if let Err(error) = fs::copy(attachment.filepath.as_path(), destination) {
```
AmmarAbouZor: We shouldn't do IO operation in the rendering thread. We need to introduce a command to copy the file and send it to the backend

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
```diff
@@ -1,20 +1,461 @@
...
+        const KB: usize = 1024;
+        const MB: usize = KB * 1024;
+        const GB: usize = MB * 1024;
+
+        match bytes {
+            b if b < KB => format!("{} B", b),
+            b if b < MB => format!("{:.2} KB", b as f64 / KB as f64),
+            b if b < GB => format!("{:.2} Mb", b as f64 / MB as f64),
+            b => format!("{:.2} GB", b as f64 / GB as f64),
+        }
```
AmmarAbouZor: You can use `file_utls::format_file_size()` to instead of the duplication here

application/apps/indexer/gui/application/src/session/ui/bottom_panel/search/search_table.rs
the-nick-fischer: `log_position` would make the most sense to me - I just left the `bookmark_pos` because I thought I do not fully understand it. It's up to you ;)

application/apps/indexer/gui/application/src/session/ui/shared/attachments.rs
the-nick-fischer: I wrongly assumed that this case would be handled in the session. I will replicate what was implemented in attachments.ts: report an error and skip the attachment

application/apps/indexer/gui/application/src/session/ui/shared/attachments.rs
the-nick-fischer: Yes, that's also possible. It felt better to me to look up the UUID of the attachment, not its index in the attachment list. Using the UUID keeps the data logic decoupled from the order of the Vec which in my mental model already is more part of the UI.
Also, this might get tricky once we introduce filtering attachments, no?

application/apps/indexer/gui/application/src/session/ui/common/logs_tables.rs
the-nick-fischer: This makes a lot of sense to me. Then I would need to pass `self.shared.attachments` inside `render_row_header`.
To be honest I struggled a bit with the closure that is being passed to the funtion and ran into ownership / multiple mut borrow issues. Maybe you can give me a hint how that would work?

application/apps/indexer/gui/application/src/session/ui/bottom_panel/search/search_table.rs
AmmarAbouZor: I agree `log_position` is a better name. You can apply it here if you want

application/apps/indexer/gui/application/src/session/ui/shared/attachments.rs
AmmarAbouZor: I don't attachments will be removed from the vector once they are inside therefore, I would use the direct index since to save a lookup on each render loop for each log item.
For filtering we will introduce another Set with the indices of matching attachments but the vector of attachments will stay as our main source of truth here

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
the-nick-fischer: That's how it behaved in the electron UI I believe. From a UX perspective it feels weird to me.


What I would expect instead:
If we secondary click:
- Selection is none: behave like primary click + open context menu
- Selection is some: keep selection + open context menu

Should I leave it as is?

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
the-nick-fischer: Okay, makes sense. Would these operations also go into UiAtions?

application/apps/indexer/gui/application/src/session/ui/common/logs_tables.rs
AmmarAbouZor: I think this will include bigger refactoring so we can skip it for now. However, I would still define an enum for the attachments to avoid passing invalid state. Something similar to
```rust
enum LogAttachmentsInfo { NoAttachment, WithAttachment { color: Option<egui::Color32> }}
```

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
AmmarAbouZor: I think you can define a general `HostCommand` to copy the file with source and destination. Then you may need to pass host command sender to the AttachmentUI from SessionUI then you can send the command directly using `UiActions`

application/apps/indexer/gui/application/src/session/ui/side_panel/attachments.rs
AmmarAbouZor: I've implemented that logic lately for log lines in the table. I think you can leave a todo for me here to apply the same logic as in logs tables after this PR

application/apps/indexer/gui/application/src/host/service/mod.rs
```diff
@@ -200,11 +201,59 @@ impl HostService {
                 // Cleanup on shutdown goes here.
                 let _ = confirm_tx.send(());
             }
+            HostCommand::CopyFiles { copy_file_infos } => {
+                for copy_file_info in copy_file_infos {
+                    self.copy_file(copy_file_info.source, copy_file_info.destination)
+                        .await?;
```
AmmarAbouZor: This will stop on first error and won't copy the rest of the files. I think we need to catch the errors and report them if any at end.

For that I would suggest creating a new module in host service module that is related for file system. You can name it `file_system` to avoid collusion with `fs` from std. There you can have both function for copy files where you can collect all errors and return them as one merged error and the already exists `copy_file()` below

application/apps/indexer/gui/application/src/host/ui/actions/mod.rs
```diff
@@ -8,7 +8,6 @@ mod host_action;
 
 pub use file_dialog::{FileDialogFilter, FileDialogHandle, FileDialogOptions};
 pub use host_action::HostAction;
-
```
AmmarAbouZor, line 11: I think this is a formatting error. We need to revert it

application/apps/indexer/gui/application/src/session/ui/logs_table/mod.rs
```diff
@@ -25,6 +25,11 @@ use crate::{
     },
 };
 
+pub enum LogAttachmentInfo {
+    NoAttachment,
+    WithAttachment { color: Option<egui::Color32> },
+}
```
AmmarAbouZor, line 32: I would add short docs here just to mention that this is related to tables rendering only

application/apps/indexer/gui/application/src/session/ui/shared/attachments.rs
```diff
@@ -0,0 +1,88 @@
...
+use rustc_hash::FxHashMap;
+use stypes::AttachmentInfo;
+use uuid::Uuid;
+
+use crate::host::common::colors;
+
+#[derive(Debug, Default)]
+pub struct AttachmentsState {
+    /// TODO: Evaluate other options of how we can store the attachment state without Vec and
+    /// multiple indices for lookups.
```
AmmarAbouZor: Do we still need this todo? I think we've found out that the current solution is suitable enough

application/apps/indexer/gui/application/src/session/ui/shared/attachments.rs
```diff
@@ -0,0 +1,88 @@
...
+        let uuid = attachment.uuid;
+
+        // In case an attachment is already registered, avoid duplication.
+        if self.index_by_uuid.contains_key(&uuid) {
+            error!(
+                "Attachment with uuid {uuid} has already been registered; skipping attachment: {} {:?}",
+                attachment.name, attachment.filepath,
+            );
+            return;
+        }
```
AmmarAbouZor: Would it be possible to have the attachments sent twice like if a new log has arrived which references the on of the already define attachments? I'm not sure how this handled but it this is not possible then it worth to be documented in a code comment here

application/apps/indexer/gui/application/src/session/ui/mod.rs
```diff
@@ -247,6 +248,16 @@ impl Session {
                     // Potential components which keep track for operations can go here.
                 }
                 SessionMessage::FileReadCompleted => self.shared.observe.set_file_read_completed(),
+                SessionMessage::AttachmentsUpdated { attachment, len } => {
+                    self.shared.attachments.add(attachment);
+                    if self.shared.attachments.attachments().len() as u64 != len {
+                        error!(
```
AmmarAbouZor: Nit: I would log this as warning as it will not break the behavior in the application

application/apps/indexer/gui/application/src/host/service/mod.rs
the-nick-fischer: Implemented fix as discussed

application/apps/indexer/gui/application/src/host/ui/actions/mod.rs
the-nick-fischer: reverted :+1:

application/apps/indexer/gui/application/src/session/ui/shared/attachments.rs
the-nick-fischer: Fixed as discussed

application/apps/indexer/gui/application/src/session/ui/mod.rs
the-nick-fischer: Done :+1:

PR #2516, opened 2026-04-13
Native: Allow files arrangement in multiple files

sudeeptarlekar:
In legacy version of Chipmunk, while loading multiple files user can rearrange the order by dragging them around. This feature was missing from the NativeUI and this commit fixes that missing feature.


sudeeptarlekar:
@AmmarAbouZor can you please review?

Thanks in advance :)

AmmarAbouZor (changes requested):
The changes are solid. However, I think we need to reconsider the decision to remove the color completely from source items and retrieve the color in render loop as items can go out of sync on multiple edge cases.

Like for now the if you disable one of the sources in the middle then the colors in segments in side view will not match the colors of items in the table (see snapshot)

I think saving the color with source items will keep the logic in one place. It would need to reassign the colors of sources on sort changes but it will keep the logic centered in one place instead of spreading the index logic everywhere and having to deal with edge-cases that we didn't count for

Image

sudeeptarlekar:
I was going through this today, in order to move the color to source, are you suggesting that we add extra field in the `GrabbedElement` struct?

AmmarAbouZor:
No we don't want to change the behavior outside of this view. My suggestion are regarding the colors in the view here:

In your changes you've removed the color filed for both the items in main view table and in small graph in side view, then after reviewing that, I've got the impression that may be not the direction to solve the coloring because the colors of files are going out of sync between the table and side view (See previous screenshot).

So the suggestion is to bring the color fields here back, but re-assign them when the order change so they will stay align with the colors when the session starts.

I hope, I expressed the idea more clearly this time

PR #2521, opened 2026-04-17
Resolve macOS code notarisation issue with electron upgrade

itsmesamster:


AmmarAbouZor (approved):
LGTM. I'll create a release after merging this so we can test it

PR #2533, opened 2026-04-27
Add alpha release workflow

itsmesamster:


application/apps/indexer/.gitignore
```diff
@@ -5,4 +5,3 @@
 lineMetadata.json
 vim-markdown-preview.html
 .DS_Store
-Cargo.lock
```
AmmarAbouZor, line 8: Are you sure we need to include `Cargo.lock` again at this point? I know this will be done once we merge the native-ui branch but for now it will just introduce multiple merge conflicts that we can avoid.

However, we can include from now if we have a good reason for that

.github/workflows/chipmunk_alpha_release.yml
```diff
@@ -0,0 +1,126 @@
+name: BuildAndReleaseChipmunkAlpha
+
+on:
+  push:
+    tags:
+      - "chipmunk-alpha-v*"
```
AmmarAbouZor: Will this be the template for release alpha release names and tags? like `chipmunk-alpha-v4.0.0` ?

Current implementation expects to get release names and tags following the template `4.0.0-alpha.1`.

AmmarAbouZor (changes requested):
I've left some comments in the code. But my main concern is about introducing a new rust cli tool:

We've tried replacing code scripts with rust cli tools in the past and it wasn't a great experience since we had to install rust toolchain, compile the CLI tool then thinking about caching it and then using it. 
Therefore, I think we should really have a good reason about why we need to suffer with that instead of writing a small python or `just` script and be done with it?

application/apps/indexer/.gitignore
itsmesamster: Hi Ammar, yes because when we are building the release, then I was facing issues since I was using the --locked flag and then when trying to build the release, github was throwing errors, if this was not included.

.github/workflows/chipmunk_alpha_release.yml
itsmesamster: I can change the tag pattern, as this was just a placeholder for the release until we finalized something. In either case, this doesn't change the naming of the artifacts, etc since those are named on the basis of the package version from application/apps/indexer/Cargo.toml

itsmesamster:
I have removed the release-app tool and instead used a python script now. I had used the cli tool since previously we used to do this via scripts and then moved to cli tool. Nonetheless, updated the same.

.github/workflows/chipmunk_alpha_release.yml
AmmarAbouZor: The name of the artifacts isn't relevant at the time being as users are supposed to download them manually.

For the release and tags name, it would be great if it can follow the pattern `4.0.0-alpha.1` ... `4.0.0-alpha.2` ...

AmmarAbouZor (commented):
I think this can be merged once the tag and release names are resolved 👍

AmmarAbouZor (approved):
This looks great. I think we can merge it then test it tomorrow

PR #2566, opened 2026-05-27
Use alpha release workflow for major releases

itsmesamster:
Use alpha release workflow for major releases
Include installers for macOS and Windows


AmmarAbouZor (changes requested):
I think we need to have our main ideas aligned here. Here is how  I would imaging the workflows:
* We should aim to end up with one release workflow on master, which is the stable release for Chipmunk 4.
* The current alpha release workflow is temporary one to do alpha releases and is expected to be removed once the stable release is there, so there is no need extend it with the release for stable Chipmunk 4
* Legacy Chipmunk doesn't live in this branch, so we don't need to include its release workflow in this branch. I think we don't need to consider the legacy release until we really need to have one.

Other points to consider:
* I would like to have a fixed tag fromat for the releases like `4.x.x`
* Having comments in GitHub Actions and in python code will help developers to understand the domain and decision

development/scripts/release_app.py
```diff
@@ -118,24 +156,171 @@ def package_macos(version, code_sign):
...
+def write_windows_license_rtf(path):
+    """Copy the license text shown by the Windows installer UI."""
+    shutil.copy2(windows_license_rtf_path(), path)
+
+
+def package_linux_update_binary(version):
+    """Publish the raw Linux binary used by the native app update flow."""
+    update_binary = app_release_path() / "chipmunk@{}-{}".format(
+        version, linux_update_platform_name()
+    )
```
AmmarAbouZor: It seems that the uploaded Linux binary doesn't have an extension `chipmunk@*-linux-x86_64` but we still need to add `tgz` extension for the file to be uploaded

AmmarAbouZor (changes requested):
Looks mostly done. But we still need to add the extension for Linux archive

AmmarAbouZor (approved):

PR #2579, opened 2026-06-08
Add debian and rpm packaging

itsmesamster:
Include chipmunk-cli in release workflow
Remove README.md from tar release artifact
Remove the unused .release file
Add debian and rpm packaging

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   create_release:
-    name: Create app release
+    name: Create release
     runs-on: ubuntu-latest
     outputs:
       upload_url: ${{ steps.create_release.outputs.upload_url }}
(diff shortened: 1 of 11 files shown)


AmmarAbouZor (commented):
I've run checks using AI agent and got this result back. However, you are the domain expert so it's up to you to decide which of them needs to be fixed if any:


_**Review Generated by AI:**_

1. **macOS CLI artifact is published unsigned**
   - `development/scripts/release_app.py:72`, `development/scripts/release_app.py:131`
   - The new CLI `.tgz` is appended after macOS app signing/notarization and `package_cli_portable()` only copies the raw CLI binary. Since CI uploads all `*.tgz`, signed macOS releases will now include an unsigned `chipmunk-cli@...-darwin*.tgz`. Users downloading it with quarantine may be blocked by Gatekeeper. Sign/notarize the CLI artifact on macOS or skip publishing the macOS CLI archive.

2. **Linux package directory modes depend on caller umask**
   - `development/packaging/linux/packages.py:325`
   - `make_dir()` uses `mkdir(parents=True)` but only `chmod`s the final directory. Parent directories created during staging inherit the process umask, so local release builds with restrictive umasks can produce DEB/RPM trees with bad directory modes. Walk and chmod each created directory under the staging root.

PR #2593, opened 2026-06-12
Improve file type detection & Notification messages

AmmarAbouZor:
* Improve detection of file types providing new error types for text files with unsupported encoding formats
* Improve messages in notifications for users providing them with more info about the reason why the file isn't supported by Chipmunk


AmmarAbouZor:
Checks ran manually as it seems that GitHub actions are down currently

PR #2599, opened 2026-06-18
Add someip-statistics

kruss:
This PR adds someip-statistics to the ui and let user apply corresponding filters on parsing.


crates/core/parsers/src/dlt/fmt.rs
AmmarAbouZor: This will write the `Option<SomeipLogMessage>` instead of the `SomeipLogMessage`.
I think we can either: 
* Use `unwrap_or_default()`
* Check if None and skip writing message

crates/app/src/host/ui/session_setup/main_config/someip.rs
```diff
@@ -0,0 +1,375 @@
...
+            .resizable(false)
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
```
AmmarAbouZor: 9 Columns are declared here whereas only 7 columns are needed. 
I suggest using 
```rust
columns(Column::reminder(), 7)
```
to increase visibility for columns count

crates/app/src/host/common/someip_stats.rs
```diff
@@ -0,0 +1,758 @@
...
+    pub fn merge(&mut self, other: &MessageDistribution) -> &mut Self {
+        self.sd += other.sd;
+        self.event += other.event;
+        self.request += other.request;
+        self.response += other.response;
+        self.fire_forget += other.fire_forget;
+        self.error += other.error;
+        self
+    }
+}
```
AmmarAbouZor, line 108: Optional:

You can use pattern matching here so the compiler will remind you to update those functions if we added new fields.
I know it seems more verbose but this could provide value avoiding potential bugs on adding fields to the struct

```rust
    /// Returns the type specific numbers of messages.
    pub fn values(&self) -> [usize; 6] {
        let Self {
            sd,
            event,
            request,
            response,
            fire_forget,
            error,
        } = self;

        [*sd, *event, *request, *response, *fire_forget, *error]
    }

    pub fn merge(&mut self, other: &MessageDistribution) -> &mut Self {
        let Self {
            sd,
            event,
            request,
            response,
            fire_forget,
            error,
        } = self;

        *sd += other.sd;
        *event += other.event;
        *request += other.request;
        *response += other.response;
        *fire_forget += other.fire_forget;
        *error += other.error;

        self
    }
```

crates/app/src/host/ui/session_setup/main_config/someip.rs
```diff
@@ -0,0 +1,375 @@
...
+            .sense(Sense::click())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
+            .column(Column::remainder())
```
AmmarAbouZor: Also here it seems that columns count is 9 while the headers are 7

crates/app/src/host/ui/session_setup/state/parsers/someip.rs
```diff
@@ -27,6 +47,129 @@ impl SomeIpParserConfig {
...
+    selected_ids: &FxHashSet<MessageId>,
+    ids_with_messages: &FxHashMap<MessageId, MessageDistribution>,
+) -> MessageDistribution {
+    let mut messages = MessageDistribution::default();
+
+    for selected_id in selected_ids {
+        if let Some((_, l)) = ids_with_messages.iter().find(|(id, _)| *id == selected_id) {
+            messages.merge(l);
+        }
+    }
```
AmmarAbouZor: Optional: 

it's possible to use `filter_map(...)` here if you want but it wouldn't improve code readability that much

crates/app/Cargo.toml
```diff
@@ -52,6 +52,8 @@ blake3.workspace = true
 nucleo-matcher.workspace = true
 anstyle-parse.workspace = true
 unicode-segmentation.workspace = true
+pcap-parser = "0.16"
+someip-messages = { git = "https://github.com/esrlabs/someip" }
```
AmmarAbouZor: We are defining all of our dependencies in `Cargo.toml` for the whole work space (in repos root) so we change them once for all components. So we need to move those definitions there.

Also we need to set `rev = "{commit sha}"` here so we target a specific commit from the git branch to keep the dependency determanistic.

And it would be nice to have a comment here stating why we are targeting a git repo instead of published crate

crates/core/parsers/src/someip.rs
```diff
@@ -187,39 +188,49 @@ impl FibexTypeCache {
...
-    ) -> Result<(usize, SomeipLogMessage), Error> {
+    ) -> Result<(usize, Option<SomeipLogMessage>), Error> {
         let time = timestamp.unwrap_or(0);
         match Message::from_slice(input) {
             Ok(Message::Sd(header, payload)) => {
+                let filtered = match filter_config {
+                    Some(config) => {
+                        !(config
+                            .messages
+                            .contains(&(header.message_id.service_id, header.message_id.method_id)))
```
AmmarAbouZor: I think we don't need the brackets here

crates/core/parsers/src/someip.rs
```diff
@@ -228,14 +239,26 @@ impl SomeipParser {
...
+                    },
                 ))
             }
 
             Ok(Message::Rpc(header, payload)) => {
+                let filtered = match filter_config {
+                    Some(config) => {
+                        !(config
+                            .messages
+                            .contains(&(header.message_id.service_id, header.message_id.method_id)))
```
AmmarAbouZor: Also here we have extra brackets

crates/core/parsers/Cargo.toml
```diff
@@ -24,6 +24,7 @@ thiserror.workspace = true
 rand.workspace = true
 someip-messages.workspace = true
 someip-payload.workspace = true
+stypes = { path = "../../stypes" }
```
AmmarAbouZor: All crates are defined once in workspace to avoid changing them in multiple place.
You can use here `staype.workspace = true` , and also please move it up to the region of internal dependencies

AmmarAbouZor (changes requested):
Thanks a lot, this looks really clean and mostly fine. I've found: 
* Two small correctness issues: Printing an Option instead of the direct value and columns count mismatch for the new tables. (Issues found by AI and confirmed by me, as I'm not that good as a reviewer!)
* Adjustments for dependencies declaration to keep them in the root workspace

I've also added some optional ideas in some place so you can apply any of them if you see that they fit

crates/app/src/host/common/someip_stats.rs
kruss: I see the benefit of exhaustiveness here, but I find the destructuring version significantly less readable. Since this is simple aggregation logic, I’d rather keep the direct field access and cover behavior with unit tests. If we later add a field, the tests for count, values, and merge should catch the missing update.

AmmarAbouZor (approved):
This looks great! Thanks 🙏 
Let's merge it

Issue #2610, opened 2026-07-04
Windows portable release artifact format issue: Incorrectly named internal .tgz file


z-Fng:
**Prerequisites**

affected version: `4.0.0`, `4.0.1`

make sure to check this before filing an issue:

- [x] I am running the latest version
- [x] I checked to make sure that this issue has not already been filed

**Describe the issue**

In the current Windows release artifacts, there is a packaging format mismatch with the nested archive files.

When extracting the main .tgz release package, the inner file is labeled with a .tgz extension, but it is actually just an uncompressed .tar file.

**Steps to Reproduce**

1. Download the Windows release package (.tgz).
2. Extract the outer package.
3. Try to decompress the inner .tgz file using 7zip.
4. Observed Result: The tool reports it is not in gzip format.

<img width="564" height="358" alt="Image" src="https://github.com/user-attachments/assets/5787484c-8ab3-4515-83e8-6e7d78c451e6" />

**Related issue**

https://github.com/ScoopInstaller/Extras/issues/18031#issuecomment-4697563767

AmmarAbouZor:
Issue resolved in #2611

z-Fng:
Thanks for the fix! Looking forward to the new release!

Issue #2606, opened 2026-06-25
Installer Fixes on Windows + Chipmunk Icon


AmmarAbouZor:
We got some issues to fix and improve on Windows as mentioned in [a comment here](https://github.com/esrlabs/chipmunk/issues/2475#issuecomment-4730516511)

I think we can continue working on the first three points from the comment here.

AmmarAbouZor:
Issue resolved in #2611

PR #2611, opened 2026-07-06
Resolve Windows installer issues

itsmesamster:
This resolves the Windows installer issues mentioned in 
[https://github.com/esrlabs/chipmunk/issues/2606](https://github.com/esrlabs/chipmunk/issues/2606)
[https://github.com/esrlabs/chipmunk/issues/2610](https://github.com/esrlabs/chipmunk/issues/2610)

### DIFF
--- a/crates/app/data/win/build.rs
+++ b/crates/app/data/win/build.rs
@@ -1,0 +1,12 @@
+// Cargo build scripts are package-wide. This Windows asset hook delegates to
+// the packaging helper that embeds chipmunk.ico into chipmunk.exe, so Explorer
+// shows the portable executable with the Chipmunk app icon.
+#[path = "../../../../development/packaging/windows_icon_resource.rs"]
+mod windows_icon_resource;
+
+fn main() {
+    println!("cargo:rerun-if-changed=data/win/build.rs");
+    println!("cargo:rerun-if-changed=../../development/packaging/windows_icon_resource.rs");
+
+    windows_icon_resource::embed_app_icon("chipmunk", "data/win/chipmunk.ico");
+}
--- a/crates/app/src/main.rs
+++ b/crates/app/src/main.rs
@@ -1,3 +1,7 @@
+// Release Windows builds are GUI apps. Without this subsystem marker, launching
+// chipmunk.exe from Explorer or the installer keeps an extra console window open.
+#![cfg_attr(all(windows, not(debug_assertions)), windows_subsystem = "windows")]
+
 fn main() -> anyhow::Result<()> {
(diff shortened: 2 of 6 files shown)


crates/app/data/win/build.rs
```diff
@@ -0,0 +1,9 @@
+#[path = "../../../../development/packaging/windows_icon_resource.rs"]
+mod windows_icon_resource;
+
+fn main() {
+    println!("cargo:rerun-if-changed=data/win/build.rs");
+    println!("cargo:rerun-if-changed=../../development/packaging/windows_icon_resource.rs");
+
+    windows_icon_resource::embed_app_icon("chipmunk", "data/win/chipmunk.ico");
```
AmmarAbouZor, line 11: Have you considered using the crate [winresource](https://crates.io/crates/winresource) instead of manually implementing all the details?

crates/app/data/win/build.rs
```diff
@@ -0,0 +1,9 @@
+#[path = "../../../../development/packaging/windows_icon_resource.rs"]
```
AmmarAbouZor, line 4: It would be helpful to write a comment about the purpose of this build script file

crates/app/src/main.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg_attr(all(windows, not(debug_assertions)), windows_subsystem = "windows")]
```
AmmarAbouZor, line 3: Also here a comment is needed to tell what is this attribute for.

AmmarAbouZor (changes requested):
All changes are mostly good. I left comments about:
* Commenting the code in important places
* Considering using a crate for embedding windows icon instead of manually implementing it

crates/app/data/win/build.rs
itsmesamster: Hi Ammar, 
I considered `winresource`, but I’d prefer not to add it here. We only need to embed one existing `.ico` file, and using `winresource` would add a build dependency plus resource-compiler/toolchain behavior for a very small packaging detail. The current helper is std-only, isolated under `development/packaging`, and the app crate only has a small Windows build hook. 
Additionally, the underlying crate winres has been left unmaintained, so it might pose issues going forward.

crates/app/src/main.rs
itsmesamster: Added comment

AmmarAbouZor (approved):
Thanks for the fixes. Let's merge it and we can test the updater in Windows in our general pre-release tests workflow

PR #2618, opened 2026-07-17
Reorder and Move Filters with Drag and Drop + Improve Multiple Files Drag and Drop 

### LINKED ISSUE #2571: Support for Drag and Drop for Filters
Chipmunk 4 still doesn't support drag and drop for filters and charts in side panel for running sessions.
Drag & Drop was used for two purposes:
* Change the oder of filters or charts
* Move an item between filters and charts if compatible

Feature was requested in #2475

AmmarAbouZor:
This PR closes: #2571 

It adds support for drag and drop in filters side panel for running sessions providing the following:
- Reordering for filters and charts via drag and drop
- Move compatible items between filters and charts via drag and drop

This PR also includes:
- Change drag and drop behavior in multiple files table view to match the behavior on filter adding highlighting on both top and bottom of items and following the relative position of mouse course when hovering over the item while dragging.
- Check and remove stale edit state for filters when filters are removed from other UI components (like Library view)
- Refactor `filters.rs` file splitting it into multiple sub-modules and removing low-signal unit tests


itsmesamster (approved):
Looks good to me. Kindly proceed.

PR #2620, opened 2026-07-31
Add Scoped Search Inside Search Results

### LINKED ISSUE #2604: Scoped Search for log results view
Chipmunk 3 provides the feature to search the logs inside the result view. Function is triggered with `Ctrl+Shift+f`

We need to port this function into Chipmunk 4

AmmarAbouZor:
This PR closes #2604

It adds nested scoped search for finding matches within filtered logs. It also includes several fixes and improvements to the core implementation.

It includes:

- Search within filtered logs with regex, case, and whole-word options.
- Navigate between matches while synchronizing both log tables.
- Highlight nested matches in search results and main logs.
- Match visible text in ANSI-formatted logs.
- Add menu and keyboard access to nested search. Search will be activated with `Ctrl+Shift+f`.
- Improve focus, progress, empty-query, and no-match behavior.
- Fix indexed-row synchronization across search results and bookmarks.
- Extend test coverage and for search in core.
- Remove `Breadcrumb` code as this feature won't be implemented in Chipmunk 4 anymore.  
- Add documentation with a demo video.


crates/core/processor/src/search/searchers/linear.rs
```diff
@@ -44,6 +48,41 @@ impl LineSearcher {
     /// * `true` - If the line matches the regular expression.
     /// * `false` - Otherwise.
     pub fn is_match(&self, ln: &str) -> bool {
-        self.re.is_match(ln)
+        // Check raw text first to avoid ANSI strip's parsing and allocation
```
itsmesamster: Image
The ANSI visible-text matching still checks raw text first. That can match hidden escape sequences like 31m, causing nested search to jump to rows where no visible/highlighted match exists, as shown in the screenshot.

itsmesamster (changes requested):

crates/core/processor/src/search/searchers/linear.rs
AmmarAbouZor: Great Catch!
I've fixed it applying scanning for escape sequences unconditionally.
Also, regression test is provided

itsmesamster (approved):

